PR #3, opened 2017-01-26
build: Fix warning about linking with a non-POSIX ar

bertwesarg:
```
ltlibrary.am: warning: 'libunwind-sh.la': linking libtool libraries using a non-POSIX
ltlibrary.am: archiver requires 'AM_PROG_AR' in 'configure.ac'
src/Makefile.am:622:   while processing Libtool library 'libunwind-sh.la'
```


djwatson:
thanks!

PR #4, opened 2017-02-01
Thread safe find eh_frame

bertwesarg:
Instead of just removing the 'static' from sec_hdrs, I decided to make code from elfxx.h re-usable to find any EFL section.

### DIFF
--- a/src/dwarf/Gfind_proc_info-lsb.c
+++ b/src/dwarf/Gfind_proc_info-lsb.c
@@ -525,19 +525,14 @@ dwarf_find_debug_frame (int found, unw_dyn_info_t *di_debug, unw_word_t ip,
 
 #ifndef UNW_REMOTE_ONLY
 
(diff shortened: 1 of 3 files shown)


src/dwarf/Gfind_proc_info-lsb.c
```diff
@@ -549,57 +544,22 @@ dwarf_find_eh_frame_section(struct dl_phdr_info *info)
   Debug (1, "looking for .eh_frame section in %s\n",
          file);
 
-  fd = open(file, O_RDONLY);
-
-  if (fd == -1)
+  rc = elf_map_image (&ei, file);
```
djwatson: It looks like this changed from open/seek/read/close calls to mmap, but I can't think of any reason that won't work

djwatson:
Awesome, thanks!  @compnerd also mentioned a use case for a generic find_section function

PR #5, opened 2017-02-02
build: check if m4 macro exists

djwatson:
@bertwesarg This fails for me on older versions of automake


bertwesarg:
+1

But the required automake 1.6 is ancient. On what supported platform is that old automake?

djwatson:
automake --version
automake (GNU automake) 1.11.1

autoconf --version
autoconf (GNU Autoconf) 2.63

The actual error message is 

configure.ac:20: warning: macro `AM_PROG_AR' not found in library
autoreconf: running: /usr/bin/autoconf --force
configure.ac:20: error: possibly undefined macro: AM_PROG_AR
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf: /usr/bin/autoconf failed with exit status: 1

PR #6, opened 2017-03-04
Don't link with gcc_s when compiling for Android

qmfrederik:
Currently libunwind fails to compile out of the box for Android because it will link with libgcc_s. That library doesn't exist on Android, so linking will fail.

With this change, you can cross-compile libunwind for Android using the NDK toolchain and sysroot:

```
./autogen.sh
./configure CC=$CC --with-sysroot=$SYSROOT --host=aarch64-linux-androideabi --target=aarch64-linux-androideabi --disable-tests --disable-coredump
```

with `$CC` set to `[toolchain]bin/aarch64-linux-android-clang` and `$SYSROOT` set to `[toolchain]sysroot/`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -105,6 +105,19 @@ SET_ARCH([$build_cpu],[build_arch])
 SET_ARCH([$host_cpu],[host_arch])
 SET_ARCH([$target_cpu],[target_arch])
 
+# Check for Android
+AC_MSG_CHECKING([for Android])
+android="no"
+case "$host_os" in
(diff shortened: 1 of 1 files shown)


qmfrederik:
This PR would unblock .NET Core building for aarch64 Android from the regular libunwind repository; as of dotnet/coreclr#9940 .NET Core is building from a fork (which I hope to remove soon).

djwatson:
looks good to me, thanks.

PR #7, opened 2017-03-22
x86_64: fix mincore_validate and msync_validate

JoZie:
The calls to mincore() or msync() are not checking for actual accessibility this could lead to SIGSEGV if the address from a mapped page with the PROT_NONE property occurs on the stack.
Hence an attempt to write one byte from the checked address to a pipe will fail if the address is not readable.


JoZie:
Sorry for the fuzz. The PR was intended for internal review.

djwatson:
I'd love to merge a patch like this.  sendmsg has msg_nosignal, might make it a bit cleaner.  Also x86 support would be great.

JoZie:
@djwatson You are right, blocking SIGPIPE is a bit messy. But using by send() you have to setup and connect a socket. IMHO this would add more code whose necessity is hard to understand. It's also possible to set and reset the signal mask right before and after the write() call.

Regarding x86 support:  Unfortunately I have no running 32Bit system at hand. But since it's posix the code should work too.

djwatson:
changing signal mask is probably ok, we do that elsewhere.

src/x86_64/Ginit.c
```diff
@@ -72,10 +72,57 @@ get_dyn_info_list_addr (unw_addr_space_t as, unw_word_t *dyn_info_list_addr,
 #define PAGE_SIZE 4096
 #define PAGE_START(a)   ((a) & ~(PAGE_SIZE-1))
 
+static int mem_validate_pipe[2];
+
+static inline void
+open_pipe (void)
+{
+  /* ignore errors for closing invalid fd's */
+  close (mem_validate_pipe[0]);
```
djwatson: I don't think you can indiscriminately close here, the first time through both fd's will be zero, so you will close fd 0 (stdout)

djwatson:
Thanks for the update! A couple tests look broken, might be that issue.

src/x86_64/Ginit.c
JoZie: Yes, you're right, thank you!

JoZie:
An user initialization of the fd's did the job. I also added your test case.
Are there any plans for removing `mincore()` and `msync()`?

PR #8, opened 2017-03-27
Bring back support for UNW_CACHE_PER_THREAD.

bertwesarg:
Needs to be build with --enable-per-thread-cache.

@JoZie can you please test this (performance only), though we need a call to
```
unw_set_caching_policy (unw_local_addr_space, UNW_CACHE_PER_THREAD);
```

somewhere in our init-phase.

### DIFF
--- a/src/dwarf/Gparser.c
+++ b/src/dwarf/Gparser.c
@@ -537,11 +537,11 @@ dwarf_flush_rs_cache (struct dwarf_rs_cache *cache)
     cache->log_size = DWARF_DEFAULT_LOG_UNW_CACHE_SIZE;
   } else {
     if (cache->hash && cache->hash != cache->default_hash)
-      munmap(cache->hash, DWARF_UNW_HASH_SIZE(cache->prev_log_size)
-                           * sizeof (cache->hash[0]));
+      mi_munmap(cache->hash, DWARF_UNW_HASH_SIZE(cache->prev_log_size)
+                              * sizeof (cache->hash[0]));
     if (cache->buckets && cache->buckets != cache->default_buckets)
-      munmap(cache->buckets, DWARF_UNW_CACHE_SIZE(cache->prev_log_size)
-	                      * sizeof (cache->buckets[0]));
+      mi_munmap(cache->buckets, DWARF_UNW_CACHE_SIZE(cache->prev_log_size)
+	                         * sizeof (cache->buckets[0]));
     GET_MEMORY(cache->hash, DWARF_UNW_HASH_SIZE(cache->log_size)
                              * sizeof (cache->hash[0]));
     GET_MEMORY(cache->buckets, DWARF_UNW_CACHE_SIZE(cache->log_size)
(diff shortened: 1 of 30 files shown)


bertwesarg:
Now the default caching policy is per-thread, if build with --enable-per-thread-cache. No need for the unw_set_caching_policy() anymore.

bertwesarg:
Sorry, that should be an PR to another branch.

djwatson:
This looks pretty good, only minor comments.  Looks like the openmp stuff was added on accident and the pull request closed though.  Clean it up and we'll merge it?

tests/Gperf-trace.c
```diff
@@ -235,6 +235,8 @@ main (int argc, char **argv)
 
   measure_init ();
 
+  doit ("default         ");
```
djwatson: debugging leftovers?

bertwesarg: no, the default changes depending on configuring with --enable-per-thread-cache, thus showing how the default performs make sense, doesn't it?

djwatson: Oh I see.  Sure.

bertwesarg:
Having a per-thread-cache should be exercised when running with threads, right. But the per-thread-cache also saves the locking, which also saves the sigprocmask syscalls. Thus it is also a win for single threaded apps.

If the the TLS size is still an issue, than maybe we need to provide two libraries, one for one-shot-usage, like the backtrace or other uses, and one for performance-needs, where the cache is actually in use.

djwatson:
We could only make the cache pointer TLS, and allocate the cache on first thread use using mmap/mempool?  We'd have to have a pthread_key to unregister it I guess.

The bigger issue for making it the default is that under the covers, __thread may actually allocate.  At least in glibc.

src/dwarf/Gparser.c
```diff
@@ -574,6 +584,8 @@ get_rs_cache (unw_addr_space_t as, intrmask_t *saved_maskp)
   if ((atomic_read (&as->cache_generation) != atomic_read (&cache->generation))
        || !cache->hash)
     {
+      /* cache_size is only set in the global_cache, copy it over before flushing */
+      cache->log_size = as->global_cache.log_size;
```
djwatson: Hmm, setting the cache size may not work quite right with a tls cache - the GET_MEMORY calls will never get cleaned up ever, and will just leak if a thread is destroyed after setting a larger cache size?

bertwesarg: Thats most likely true, I did not dive into the 'set cache size' thing much yet, though I don't have an idea how to handle it.

djwatson: We could punt for now and disallow set cache size for UNW_CACHE_PER_THREAD if you want.  Otherwise I think it requires a pthread key destructor.

bertwesarg: I could live without a custom per-thread cache size

djwatson:
I disabled unw_set_cache_size if this is enabled, and merged as 57257060c9d56cf9d1a499d71e3ca1eb51cd279d

PR #18, opened 2017-05-05
Expose register states

unkadoug:
I need to use libunwind to build my own cache of register states, so I've added to the API a function that uses a callback function to pass register states back to me, along with the address ranges that bound them, so that I can cache them, and also a function that lets me pass in a register state, and a cursor, and update the registers of the cursor with that state.


unkadoug:
I have no means to reproduce the error reported here, so I'm just going to have to start over, post this change in smaller pieces until I can identify the piece that introduces the error.

PR #23, opened 2017-06-14
Dwarf cache nodes are allocated in a round-robin fashion, despite the

unkadoug:
'lru' prefix used in several data fields.  Drop the unnecessary fields,
and just use a simple counter to track the next cache entry to be recycled.

### DIFF
--- a/src/dwarf/Gparser.c
+++ b/src/dwarf/Gparser.c
@@ -583,7 +583,7 @@ dwarf_flush_rs_cache (struct dwarf_rs_cache *cache)
     cache->prev_log_size = cache->log_size;
   }
 
-  cache->lru_head = 0;
+  cache->rr_head = 0;
 
   for (i = 0; i < DWARF_UNW_CACHE_SIZE(cache->log_size); ++i)
     {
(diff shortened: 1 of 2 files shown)


src/dwarf/Gparser.c
```diff
@@ -682,11 +679,7 @@ rs_new (struct dwarf_rs_cache *cache, struct dwarf_cursor * c)
   unsigned short head;
 
   head = cache->lru_head;
-  cache->lru_head = cache->links[head].lru_chain;
-
-  /* re-insert rs at the tail of the LRU chain: */
-  cache->links[cache->lru_tail].lru_chain = head;
-  cache->lru_tail = head;
+  cache->lru_head = (head + 1) % DWARF_UNW_CACHE_SIZE(cache->log_size);
```
djwatson: Can this be a bitshift instead, since it is a log_size?

djwatson:
Nice.  Yea the 'lru' prefix is super confusing.

djwatson:
squashed and merged, thanks

PR #24, opened 2017-06-15
Arm code uses dwarf before arm for finding proc info, but unlike

unkadoug:
dwarf code, ignores error return from dwarf_callback.  Change arm
code to behave like dwarf code in that case.


djwatson:
merged, thanks.

PR #32, opened 2017-07-06
Ben Avison (bavison@riscopen.org) has observed that when a synthetic

unkadoug:
eh_frame_hdr is generated, there is no space in it for the eh_frame,
so the eh_frame value is written to, and later read from, memory that
is not assigned to this purpose, with unpredictable results.

This change adds a new field to the dwarf_eh_frame_hdr type, to
make room for that value, and adds the (packed) attribute to the
struct defintion to avoid a problem with unused space in the struct.

### DIFF
--- a/include/dwarf-eh.h
+++ b/include/dwarf-eh.h
@@ -106,16 +106,16 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 
 #define DW_EH_VERSION           1       /* The version we're implementing */
 
-struct dwarf_eh_frame_hdr
+struct __attribute__((packed)) dwarf_eh_frame_hdr
   {
     unsigned char version;
     unsigned char eh_frame_ptr_enc;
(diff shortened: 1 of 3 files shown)


bavison:
I'm the reporter... just to say this fix looks good to me. I can't see any instances of pointer arithmetic on the struct other than the ones you've already removed, so I can't see any harm in extending it like this.

djwatson:
Is the single_fde patch still needed in addition to this, or no?

unkadoug:
I know that the patch I've submitted fixes something that is
unquestionably, absolutely, broken.  The changes related to single_fde, I'm
not as confident about because I don't know what the point of single_fde
is.  But I'm likely to push that patch on you after this one is accepted.

Doug

bavison:
I have checked and can confirm that this patch is not by itself enough to prevent the segfaults I expeienced - I need the single_fde patch as well.

djwatson:
Merged, thanks.

PR #34, opened 2017-07-06
Usage of the single_fde field in cb_data suggests that it should be

unkadoug:
set only when dwarf_extract_proc_info_from_fde has completed successfully,
but instead it is set before the linear search for the matching ip has
begun.  Set it only when that search has completed successfully, and
has thus extracted the proc_info.


djwatson:
merged, thanks.

PR #36, opened 2017-07-19
Fix ppc build problem

unkadoug:
Add register state and state iterate functions to ppc32 and ppc64 directories.  They were added to other arch directories in change 502ba27753... but not to these, for some reason.

The result is that libuwind can be made again on a ppc64le, and 'make check' fails as badly as it did before the build problem was introduced, but no worse.


djwatson:
merged, thanks

PR #37, opened 2017-08-06
Update dwarf ip cache on more archs

unkadoug:
For architectures that initialize the dwarf ip cache, but do not update when the ip register is set, add those updates.


djwatson:
merged, thanks.

PR #38, opened 2017-08-14
dwarf: Fix cache size calculation

djwatson:
The and mask trick only works for power-of-two sized things,
but must be computed using the full size.  This incorrectly
resulted in a very small cache size.

Found using bisect and 'make perf' in tests directory.

blame rev: 0b51f5892df0691fc7b3b947647222ab8e57dd54


djwatson:
@unkadoug

unkadoug:
Sorry about that.

djwatson:
merged

PR #40, opened 2017-08-20
Use `ucontext_t` instead of `struct ucontext`

yuyichao:
Ref https://sourceware.org/git/?p=glibc.git;a=commit;h=251287734e89a52da3db682a8241eb6bccc050c9
And this is what other part of the code uses.

------------------------------

~~There's also `struct ucontext_t` from `<sys/ucontext.h>` that should be ABI compatible (and is what I usually use elsewhere). I'm not sure why the kernel version is used or which one is preferred here.~~


yuyichao:
From the travis failure it seems that this was meant to be the user space one but was just using the wrong name. Since `ucontext_t` was used elsewhere in the code I changed it to that and updated the commit message, etc. accordingly.

djwatson:
merged, thank you.

Issue #43, opened 2017-10-16
libunwind crashes during stack unwinding


kbaladurin:
Following example crashes with libunwind-1.2 (commit 5f354cb7b9c84dae006f0ebd8ad7a78d7e2aad0c v1.2-stable branch) on x86:
```cpp
#include <vector>
#include <cstdio>

#define UNW_LOCAL_ONLY
#include <libunwind.h>

__attribute__((noinline))
void show_backtrace (void)
{
    unw_cursor_t cursor; unw_context_t uc;
    unw_word_t ip, sp, esi;

    unw_getcontext(&uc);
    unw_init_local(&cursor, &uc);
    while (unw_step(&cursor) > 0) {
        unw_get_reg(&cursor, UNW_REG_IP, &ip);
        unw_get_reg(&cursor, UNW_REG_SP, &sp);
        unw_get_reg(&cursor, UNW_X86_ESI, &esi);
        printf ("ip = %lx, sp = %lx, esi = %lx\n", (long) ip, (long) sp, (long) esi);
    }
}

__attribute__((noinline))
void func(int a, std::vector<char*> &v)
{
    show_backtrace();
    printf("%d %d\n", a, v.size());
}

int main(int argc, char **argv)
{
    std::vector<char*> v(2);

    if (argc > 1)
    {
        if (argc > 100)
            return 1;

        v[0] = argv[argc - 2];
        v[1] = argv[argc - 1];

        func(argc - 1, v);

        return 2;
    }
    else
    {
        v[0] = argv[0];

        func(argc - 2, v);

        return 3;
    }

    return 0;
}
```
**Steps to reproduce**
1. Compile libunwind-1.2:
```
$ git branch 
  master
* v1.2-stable
$ ./autogen.sh
./configure --enable-debug --host=i386-linux-gnu "CFLAGS=-m32" "CXXFLAGS=-m32" "LDFLAGS=-m32"
$ make
```
2. Compile example:
```
$ g++ --version
g++ (Ubuntu 5.4.0-6ubuntu1~16.04.5) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

[...]

PR #46, opened 2017-10-16
dwarf/Gparser: fix crash during unwinding

kbaladurin:
We should update locations of the registers after all of them will be restored. Otherwise some locations will be incorrect.

For example if function stores ebp, edi, esi and ebx registers on
stack in its prologue, compiler can generate following unwind info:
DW_CFA_expression: r5 (ebp) (DW_OP_breg5 (ebp): 0)
DW_CFA_expression: r7 (edi) (DW_OP_breg5 (ebp): -4)
DW_CFA_expression: r6 (esi) (DW_OP_breg5 (ebp): -8)
DW_CFA_expression: r3 (ebx) (DW_OP_breg5 (ebp): -12)

In this case locations of the ebx and ebp will be calculated using current ebp but locations of the esi and edi will be calculated using previous (restored) one. Due to it their locations will be incorrect and it could lead to crash if we will try to get esi or edi value.

This patch fixes this problem.

### DIFF
--- a/src/dwarf/Gparser.c
+++ b/src/dwarf/Gparser.c
@@ -826,39 +826,44 @@ apply_reg_state (struct dwarf_cursor *c, struct dwarf_reg_state *rs)
       cfa = DWARF_GET_LOC (cfa_loc);
     }
 
+  dwarf_loc_t new_loc[DWARF_NUM_PRESERVED_REGS];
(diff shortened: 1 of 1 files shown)


kbaladurin:
This patch fixes #43

PR #57, opened 2017-11-20
Fix an include guard in src/mi/backtrace.c

qzmfranklin:


qzmfranklin:
I see. That was actually intended.

PR #59, opened 2017-11-24
dwarf: Fix size of state to avoid corrupting rs_stack

mundaym:
DW_CFA_remember_state used memcpy to overwrite state with the value
of rs_current. Unfortunately rs_current was slightly larger than state
potentially resulting in a buffer overrun. On my system this corrupted
rs_stack on a different thread.

Fix this by making the type of state match the type of rs_current. I've
also changed the memcpy call into an assignment which should ensure that
the sizes of state and rs_current match in future.

### DIFF
--- a/include/dwarf.h
+++ b/include/dwarf.h
@@ -260,7 +260,7 @@ dwarf_reg_state_t;
 typedef struct dwarf_stackable_reg_state
   {
     struct dwarf_stackable_reg_state *next;       /* for rs_stack */
-    dwarf_reg_only_state_t state;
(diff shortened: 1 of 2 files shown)


mundaym:
I'm currently in the process of porting libunwind to IBM Z (s390x) and I ran into this bug when testing Gtest-concurrent on the platform. helgrind identified the issue. Let me know if there is anything further you would like me to do. Thanks!

djwatson:
Merged, thanks!

Happy to merge any new ports as well.

PR #60, opened 2017-12-12
Add port to Linux on IBM Z (s390x)

mundaym:
The changes to shared files are minimal so this might be safe for 1.3 (famous last words!).
Thanks for taking a look.

-----

This adds a port to Linux on the IBM Z platform (a.k.a s390x). It only
supports the 64-bit ABI. Most functionality is working and all the tests
pass with the exception of the coredump tests*.

Unwinding is only supported if DWARF unwind information is present.
libunwind can't currently make use of the backchain (if present).

The getcontext/setcontext functions only preserve/restore a subset of
registers. Currently this only consists of callee-saved registers and
some parameter registers.

Vector registers and access registers are not saved (and aren't callee-
saved) by getcontext and cannot currently be modified. They will however
be restored unmodified after resuming a context from a signal handler.

There is no special libunwind support for setjmp, the functionality is
emulated using glibc (I think all the ports do this for modern Linux
kernels).

* Unwinding on s390x requires floating point register access which the
coredump library doesn't currently support.


djwatson:
I gave it a once over and didn't notice anything out of place.  Thanks!

mundaym:
Hi, this seems to have disappeared from the master branch, should I recreate this pull request? Thanks!

djwatson:
Sorry about that, my mistake!  Should be back now.  I also removed PROTECTED visibility.

mundaym:
No worries, thanks for updating and re-committing it.

PR #67, opened 2018-02-26
Fix crasher test for gcc >= 8 when using -O2 or -O3.

Romain-Geissler-1A:
Hi,

This should fix the crasher tests with gcc >= 8 and -O2/-O3:

gdb backtrace with -O2/-O3 and without this patch:
```
Program received signal SIGSEGV, Segmentation fault.
a () at crasher.c:100
100       *(volatile int *)32 = 1;
(gdb) bt
#0  a () at crasher.c:100
#1  0x00000000004008d0 in b (x=x@entry=0) at crasher.c:111
#2  0x00000000004006b9 in main (argc=<optimized out>, argv=<optimized out>) at crasher.c:122
```

with this patch, we get the expected output:
```
Program received signal SIGSEGV, Segmentation fault.
a () at crasher.c:100
100       *(volatile int *)32 = 1;
(gdb) bt
#0  a () at crasher.c:100
#1  0x00000000004008ad in b (x=<optimized out>) at crasher.c:111
#2  0x00000000004008c2 in b (x=x@entry=0) at crasher.c:113
#3  0x00000000004008ec in main (argc=<optimized out>, argv=<optimized out>) at crasher.c:122
```

Cheers,
Romain

### DIFF
--- a/tests/crasher.c
+++ b/tests/crasher.c
@@ -87,6 +87,11 @@ write_maps(char *fname)
 #endif
 
 #ifdef __GNUC__
+#ifndef __clang__
+// Gcc >= 8 became too good at inlining aliase c into b when using -O2 or -O3,
+// so force -O1 in all cases, otherwise a frame will be missing in the tests.
+#pragma GCC optimize "-O1"
+#endif
 int c(int x) NOINLINE ALIAS(b);
 #define compiler_barrier() asm volatile("");
 #else


tests/crasher.c
```diff
@@ -87,6 +87,9 @@ write_maps(char *fname)
 #endif
 
 #ifdef __GNUC__
+// Gcc >= 8 became too good at inlining aliase c into b when using -O2 or -O3,
+// so force -O1 in all cases, otherwise a frame will be missing in the tests.
+#pragma GCC optimize "-O1"
```
yuyichao: Not sure about the policy here but I believe this will cause a unknown pragma warning on clang.

Romain-Geissler-1A: Do you want me to protect that with a #ifndef __clang__ ?

yuyichao: That's the not sure about policy port ;-p

Another approach I've used that may be slightly more compiler friendly is to use a page allocated as `PROT_NONE`.

Romain-Geissler-1A: I am not sure I understand the part with the page allocated as `PROT_NONE`.

yuyichao:
```c++
    volatile uint32_t *addr = (uint32_t*)mmap(0, page_size /* 4096 */, PROT_READ, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
    *addr = 1;
```

tests/crasher.c
Romain-Geissler-1A: Well creating a segfault is not the problem here. It correctly segfaults.

Ok I realize I have badly posted my initial report. I should have added the first line:
```
Program received signal SIGSEGV, Segmentation fault.
```
as well for the first case. But it is really here.

The problem I am trying to fix in this pull request is not the segfault itself, but the unwinding part. Until now, gcc used not to inline the call to `c` (which is an alias of `b`) while now it does.

yuyichao: Ah, sorry, I somehow thought you mean that the GCC optimized out the UB....... The clang warning should still be there though and I'm not the one to decide what's the best way for this project.
(Would be nice if the test/this test can just be always compilied with `-O1` or `-O0`.)

yuyichao: > I should have added the first line:
>
> ```
> Program received signal SIGSEGV, Segmentation fault.
> ```
>
> as well for the first case

Ah, yeah, that's why I was confused..............

djwatson:
lgtm, thanks!

PR #71, opened 2018-04-03
aarch64: tune down size of unw_context_t and unw_cursor_t

djwatson:
aarch64 defines a huge __reserved field in sigcontext.  Cut it down
to only the used FP fields.

unw_cursor_t can also be cut down a bit, while still maintaining some reserved space.


djwatson:
@tracelog

PR #72, opened 2018-04-04
dwarf: do not allocate in load_debug_frame

djwatson:
load_debug_frame calls malloc() in a couple spots, use mmap via
GET_MEMORY instead.  These call paths are infrequent, and are never
freed.


djwatson:
@tracelog

bertwesarg:
:+1:

PR #74, opened 2018-04-11
Include wait-related headers for Ltest-mem-validate

ararslan:
Without these, I was getting errors from this test set claiming that `wait`, `WIFCONTINUED`, et al. were undefined on FreeBSD 11.1.


djwatson:
lgtm, thanks!

PR #76, opened 2018-04-24
fix various dataraces

milianw:
This is a follow up to the mail I sent to the list. Contained in this patch series are two fixes for data races found by TSAN. Additionally, a valgrind warning is fixed where close was called with `fd == -1`. Then one compiler warning is silenced by removing an unused variable.

Tests still seem to pass fine on my system. More importantly, the TSAN / valgrind warnings are not shown anymore.


djwatson:
> Do we want to keep my approach when the thread local data cache is available and enabled?

Yea that would be fine

djwatson:
code looks good to me, although travis has build failures

djwatson:
merged, thanks.

PR #77, opened 2018-05-04
Wip/per thread cache size

milianw:
see the mail I send to the mailing list. I'm unsure what I'm breaking here...

### DIFF
--- a/src/dwarf/Gparser.c
+++ b/src/dwarf/Gparser.c
@@ -541,8 +541,7 @@ dwarf_flush_rs_cache (struct dwarf_rs_cache *cache)
 {
   int i;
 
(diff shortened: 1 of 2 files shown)


djwatson:
cache->hash, buckets, and links for the default log_size are stored in TLS mem.  If you call set_cache_size, we allocate a new cache using mmap.  If you're using threads, this memory is never unmapped unless there is a pthread key destructor.  Which I think is fine to add and use if you're using a per-thread cache anyway.

djwatson:
No updates for a year, closing.  Feel free to reopen if you want to work on it. 

IIRC, this needs a pthread destructor to clean up each thread's cache, otherwise it leaks memory.

PR #78, opened 2018-05-04
RFC: Remove tsl_model(initial-exec)

milianw:
Use the default tsl_model for the per-thread dwarf cache.
This allows using libunwind via dlopen even when it is configured
with --enable-per-thread-cache.

We'll need to wait for a result in the mailing list discussion on how to handle this.

### DIFF
--- a/src/dwarf/Gparser.c
+++ b/src/dwarf/Gparser.c
@@ -597,7 +597,7 @@ get_rs_cache (unw_addr_space_t as, intrmask_t *saved_maskp)
 #if defined(HAVE___THREAD) && HAVE___THREAD
   if (likely (caching == UNW_CACHE_PER_THREAD))
     {
-      static __thread struct dwarf_rs_cache tls_cache __attribute__((tls_model("initial-exec")));
+      static __thread struct dwarf_rs_cache tls_cache;
       Debug (16, "using TLS cache\n");
       cache = &tls_cache;
     }


src/dwarf/Gparser.c
```diff
@@ -597,7 +597,7 @@ get_rs_cache (unw_addr_space_t as, intrmask_t *saved_maskp)
 #if defined(HAVE___THREAD) && HAVE___THREAD
   if (likely (caching == UNW_CACHE_PER_THREAD))
     {
-      static __thread struct dwarf_rs_cache tls_cache __attribute__((tls_model("initial-exec")));
```
djwatson: I think we should make this configurable - my understanding is the __thread cache is signal-safe for initial-exec, but not in a dynamically loaded library.

milianw: @djwatson So that then implies we'll need to set the initial-exec model on *all* `__thread` variables, not just this one, no?

What I still don't understand is the interaction with `--per-thread-cache` and signal safety. I.e. in https://github.com/libunwind/libunwind/pull/77 you write, that adding the pthread destructor "is fine to add and use if you're using a per-thread cache anyway." But Bert said that the main reason for the destructor not being available is that it's not signal safe. So if we add the destructor, `--per-thread-cache` would become signal-unsafe...

So, long story short: Do we really need to make `--per-thread-cache` signal-safe?

bertwesarg: The original comment regarding signal safety was from Dave. I'm not sure about it. Thread cancellation should not happen inside a signal handler, but maybe a signal handler may still be triggered while canceling a thread. But I doubt that this is allowed.

For our use-case, a per-thread cache and signal-safety is very important. Though we know that we are not there yet. But that is not restricted to libunwind.

djwatson: > So that then implies we'll need to set the initial-exec model on all __thread variables, not just this one, no?

Right.

> I.e. in #77 you write, that adding the pthread destructor "is fine to add and use if you're using a per-thread cache anyway."

I'm not sure if a pthread destructor is signal-safe or not, and I don't think we care, only that __thread usage is signal-safe, or you explicitly choose a configure option to make it otherwise.

> So, long story short: Do we really need to make --per-thread-cache signal-safe?

Yes, we would love per-thread-cache to be signal-safe.

milianw:
OK, got it. I'll update both patch sets once I get the time.

Cheers

djwatson:
Closing with no updates, feel free to reopen if you want to work on it. 

IIRC this needs to be a configuration-time option.

PR #81, opened 2018-06-12
s390x: fix build when long double is more than 64 bits

mundaym:
The floating point registers on s390x are 64-bits wide so type
them as double precision rather than long double.

Also, GCC on Ubuntu 18.04 complains about the sizeof calls so fix
those too.


djwatson:
merged, thanks.

Issue #68, opened 2018-03-02
Segfault on aarch64 with clang


vicencb:
Hello,
I am using libunwind on arm64 with clang and the musl-libc.
The issue is also reproducible with the glibc.
The resulting program built with libunwind segfaults.
It has been tested with clang versions 5.0.1 and 6.0.0-rc3 and libunwind versions v1.3-rc1 and 7d6cc6696ab8a808da3dbe23ca2493ddf2799b56.

This script reproduces the issue:
```sh
#!/bin/sh

clang --version
# clang version 5.0.1 (tags/RELEASE_501/final)
# Target: aarch64-unknown-linux-gnu
# Thread model: posix
# InstalledDir: /usr/bin

if false ; then
  export PATH=/opt/llvm/bin:/usr/bin
  clang --version
  # clang version 6.0.0 (tags/RELEASE_600/rc3)
  # Target: aarch64-unknown-linux-gnu
  # Thread model: posix
  # InstalledDir: /opt/llvm/bin
fi

git clone git://git.sv.gnu.org/libunwind.git

cd libunwind
CC=clang ./autogen.sh \
  --enable-static          \
  --disable-shared         \
  --disable-documentation  \
  --disable-tests          \
  --disable-coredump       \
  --disable-ptrace         \
  --disable-setjmp         \
  --disable-debug          \
  --disable-cxx-exceptions \
  --disable-msabi-support  \
  --disable-minidebuginfo
make

cat << EOF > uw.c
#define UNW_LOCAL_ONLY
#include <libunwind.h>
#include <stdio.h>

void show_backtrace(int x, int y) {
  unw_cursor_t cursor; unw_context_t uc;
  unw_word_t ip, sp;

  unw_getcontext(&uc);
  unw_init_local(&cursor, &uc);
  while (unw_step(&cursor) > 0) {
    unw_get_reg(&cursor, UNW_REG_IP, &ip);
    unw_get_reg(&cursor, UNW_REG_SP, &sp);
    printf ("ip = %lx, sp = %lx\n", (long) ip, (long) sp);
  }
}

int main() {
  show_backtrace(2, 3);
  return(0);
}
EOF

clang -Iinclude -O0 -g -c uw.c -o uw.o
clang uw.o src/.libs/libunwind.a -o uw
./uw
# ip = 3, sp = ffffd37c2bf0
# Segmentation fault
```

Regards,
&nbsp;&nbsp;Vicenç.

djwatson:
[...]

PR #83, opened 2018-07-18
Add FreeBSD/aarch64 support

myfreeweb:
I'm actually quite confused about how aarch64 contexts are implemented…

- there's `unw_tdep_context_t` / `unw_sigcontext`, a private copy of Linux structs
  - filled by the inline assembly in the `unw_tdep_getcontext` macro??
  - but there was a reference to `fpsimd_context` from Linux system headers??
  - the assembly does not seem to save the float/simd registers though??

So I added an internal Linux-compatible fpsimd struct and that compiles…

I'm not sure how to properly test, `make check` tries to link the library twice for some reason:

```
/usr/bin/ld: error: duplicate symbol: dwarf_search_unwind_table_int
>>> defined at Gfind_proc_info-lsb.c:809 (dwarf/Gfind_proc_info-lsb.c:809)
>>>            Gfind_proc_info-lsb.o:(_Uaarch64_dwarf_search_unwind_table) in archive ../src/.libs/libunwind-aarch64.a
>>> defined at Gfind_proc_info-lsb.c:809 (dwarf/Gfind_proc_info-lsb.c:809)
>>>            Lfind_proc_info-lsb.o:(_ULaarch64_dwarf_search_unwind_table) in archive /mnt/greg/src/github.com/libunwind/libunwind/src/.libs/libunwind.a
```

Tried to run something manually:

- `run-coredump-unwind` (and `-mdi`) expect GNU binaries: `mktemp: illegal option -- -`
- `test-ptrace-misc` returns `sum = 0`
- `test-strerror` and `test-proc-info` exit 0 without printing anything
- `forker` segfaults in `strtol_l`
- the example from issue #68 prints something reasonable (`ip = 20394, sp = ffffffffdaf0`) and does not segfault! :)

### DIFF
--- a/include/libunwind-aarch64.h
+++ b/include/libunwind-aarch64.h
@@ -196,6 +196,17 @@ typedef struct
 	struct unw_sigcontext uc_mcontext;
   } unw_tdep_context_t;
 
+typedef struct
+  {
+	uint32_t _ctx_magic;
+	uint32_t _ctx_size;
+	uint32_t fpsr;
+	uint32_t fpcr;
+	uint64_t vregs[64];
+  } unw_fpsimd_context_t;
+
+
+
 #include "libunwind-common.h"
 #include "libunwind-dynamic.h"
 
--- a/src/aarch64/unwind_i.h
+++ b/src/aarch64/unwind_i.h
@@ -59,6 +59,6 @@ extern int aarch64_local_resume (unw_addr_space_t as, unw_cursor_t *cursor,
   } while (0)
 #endif
 
-#define GET_FPCTX(uc) ((struct fpsimd_context *)(&uc->uc_mcontext.__reserved))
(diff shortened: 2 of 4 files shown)


djwatson:
This patch seems to implement _UPT_access_fpreg for aarch64, is that correct? 

It looks fine to me for that, but your PR title and description seem to be completely unrelated to what the patch is doing.

myfreeweb:
@djwatson that's one part of it. (yeah I didn't even notice that wasn't OS-specific) there's also aarch64 support in `_UCD_access_reg_freebsd.c`, and `unw_fpsimd_context_t` instead of using that struct directly from Linux headers to make it work on FreeBSD

djwatson:
Gotchya.  Updating the commit message and merging, thanks.

PR #88, opened 2018-09-13
Use mmap for debug_frames

unkadoug:
Adopt two changes by Sergey Korolev to use mmap and a shell sort, rather than malloc and qsort, to process debug frames and reduce problems with
non signal-safe functions in supposedly signal-safe libunwind procedures.


djwatson:
Merged, thank you!

PR #93, opened 2018-10-29
dwarf: Push correct CFA onto stack for dwarf expression evaluation.

stephen-roberts-work:
This change fixes a bug where stale CFAs were pushed onto the dwarf expression stack before expression evaluation. Some optimising compilers emit CFI which relies on this being correct.

Please note: I had to do some guesswork when adding the testcase, so there may be mistakes there and I'll be happy to do any rework required.


stephen-roberts-work:
Also, I wasn't really sure what licensing boilerplate the testcases would require, so I left that blank.

PR #102, opened 2019-01-08
dwarf: clang doesn't respect the static alias

djwatson:
Clang doesn't respect the static alias, resulting in global instead
of local symbols, and name collisions.  Work around for clang by always
going through the PLT (at a small perf hit)

### DIFF
--- a/src/dwarf/Gfind_proc_info-lsb.c
+++ b/src/dwarf/Gfind_proc_info-lsb.c
@@ -46,12 +46,17 @@ struct table_entry
 #include "os-linux.h"
 #endif
 
+#ifndef __clang__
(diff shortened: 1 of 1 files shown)


djwatson:
@ararslan Can you test?  Thanks

ararslan:
Works perfectly with Clang on both Linux and FreeBSD!

ararslan:
For good measure I tested with GCC on Linux as well. Looking good there too.

ararslan:
Thanks for addressing this! Will this change be made to the 1.4 branch as well?

djwatson:
> Will this change be made to the 1.4 branch as well?
done!

PR #106, opened 2019-03-13
x86_64: Add fixup code if previous RIP was invalid

yaspoon:
Previous RIP could become invalid if a bad function pointer
was followed. Calling unwind from a signal frame in this case
skips the frame that called the bad pointer.

Check if the value at RBP-8 looks to be valid in this case. It
should be the value of RIP pushed to the stack by the CALL instruction.
If it looks valid use this value as the next frames RIP else fallback
to the previous method.

As requested in [1] I've updated/cleaned up my patch and submitted a pull request for review

[1] http://lists.nongnu.org/archive/html/libunwind-devel/2019-03/msg00002.html

### DIFF
--- a/src/x86_64/Gstep.c
+++ b/src/x86_64/Gstep.c
@@ -158,43 +158,83 @@ unw_step (unw_cursor_t *cursor)
             }
           else
             {
-              unw_word_t rbp1 = 0;
-              rbp_loc = DWARF_LOC(rbp, 0);
-              rsp_loc = DWARF_NULL_LOC;
-              rip_loc = DWARF_LOC (rbp + 8, 0);
-              ret = dwarf_get (&c->dwarf, rbp_loc, &rbp1);
-              Debug (1, "[RBP=0x%lx] = 0x%lx (cfa = 0x%lx) -> 0x%lx\n",
-                     (unsigned long) DWARF_GET_LOC (c->dwarf.loc[RBP]),
-                     rbp, c->dwarf.cfa, rbp1);
-
-              /* Heuristic to determine incorrect guess.  For RBP to be a
-                 valid frame it needs to be above current CFA, but don't
-                 let it go more than a little.  Note that we can't deduce
-                 anything about new RBP (rbp1) since it may not be a frame
-                 pointer in the frame above.  Just check we get the value. */
-              if (ret < 0
-                  || rbp < c->dwarf.cfa
-                  || (rbp - c->dwarf.cfa) > 0x4000)
+              /*
+               * Check if previous RIP was invalid
+               * This could happen if a bad function pointer was
+               * followed and so the stack wasn't updated by the
+               * preamble
+               */
+              unw_word_t not_used = 0;
+              int rip_fixup_success = 0;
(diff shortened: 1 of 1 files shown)


yaspoon:
Unwind.txt is a C program I used to test patch I've submitted. It purposefully calls a bad function pointer which exhibits the bug.

The following is the output from attached test program.
What the output looks like from master:
Handler called
ip = 55987c2c4324, sp = 7ffe51cad6d0 offset = 29 name = handle_sigsegv
ip = 7f5963145e00, sp = 7ffe51cad700 offset = 40 name = killpg
ip = 1, sp = 7ffe51cadcb8 offset = 40 name =
ip = 55987c2c43bb, sp = 7ffe51cadcc8 offset = 78 name = main
ip = 7f5963132223, sp = 7ffe51cadd90 offset = f3 name = __libc_start_main
ip = 55987c2c410e, sp = 7ffe51cade50 offset = 2e name = _start

What it "should" look like with the missing "crash" function, this is
with the patch and what a gdb backtrace shows:
Handler called
ip = 55ceda4c5324, sp = 7ffc963ff810 offset = 29 name = handle_sigsegv
ip = 7fc00c47be00, sp = 7ffc963ff840 offset = 40 name = killpg
ip = 1, sp = 7ffc963ffdd8 offset = 40 name =
ip = 55ceda4c5340, sp = 7ffc963ffde0 offset = 12 name = crash
ip = 55ceda4c53bb, sp = 7ffc963ffdf0 offset = 78 name = main
ip = 7fc00c468223, sp = 7ffc963ffeb0 offset = f3 name = __libc_start_main
ip = 55ceda4c510e, sp = 7ffc963fff70 offset = 2e name = _start

To build the test program I used the following command:
gcc -o unwind unwind.c -g -lunwind -lunwind-x86_64

[unwind.txt](https://github.com/libunwind/libunwind/files/2960488/unwind.txt)
[debug-with-patch.log](https://github.com/libunwind/libunwind/files/2960476/debug-with-patch.log)
[debug-master.log](https://github.com/libunwind/libunwind/files/2960481/debug-master.log)

src/x86_64/Gstep.c
```diff
@@ -158,43 +158,83 @@ unw_step (unw_cursor_t *cursor)
...
+                      rbp_loc = DWARF_NULL_LOC;
+                    }
+
+                  c->frame_info.frame_type = UNW_X86_64_FRAME_GUESSED;
+                  c->frame_info.cfa_reg_rsp = 0;
+                  c->frame_info.cfa_reg_offset = 16;
+                  c->frame_info.rbp_cfa_offset = -16;
+                  c->dwarf.cfa += 16;
+
+                  /* Mark all registers unsaved */
```
djwatson: The ptrace test is failing because I believe this codeblock needs to be called unconditionally (pulled up two indent levels).

yaspoon: I'll add a test to the test suite this weekend and try and fixup the ptrace test at the same time. Thanks for the quick feedback.

yaspoon:
Closed and continued in pull request #120 as I screwed up and rebased and can't push without a messy merge.

PR #110, opened 2019-04-09
mips: make _step_n64 as a static function

snyh:
1. hidden _step_n64 to avoid a strange bug which
   will reproduce under `perf report`.
   The value `tdep_init_done` of suddenly be changed to zero
   when entering `_step_n64` from `unw_step`.

2. refactor code to improve the readability.
  use `FP_REG`, `SP_REG` and `RA_REG` to abstract the guessing
  logical.

### DIFF
--- a/src/mips/Gstep.c
+++ b/src/mips/Gstep.c
@@ -111,85 +111,91 @@ mips_handle_signal_frame (unw_cursor_t *cursor)
 }
 
 
-int _step_n64(struct cursor *c)
-{
-  //TODO:handle plt entry
-  struct dwarf_loc fp_loc, pc_loc;
-  int ret;
-  unw_word_t fp = 0;
-  unw_word_t ra = 0;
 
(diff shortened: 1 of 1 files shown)


snyh:
@djwatson 

There is a segment fault when run `perf report` with https://github.com/libunwind/libunwind/pull/100 .
The problem is the content of `c.dwarf.loc` and `tdep_init_done`  are right on `unw_step`, but they are changed to invalidate after entered the `_step_n64`.

When the problem happened,
```
  >_step_n64: tdep_init_done == 0
  >_step_n64: c->dwarf.loc[SP_REG] == 2
  >_step_n64: c->dwarf.loc[FP_REG] == f
  >_step_n64: c->dwarf.loc[RA_REG] == 2
```

The workaround is make `_step_n64` as a static function.  I don't know why, but it works. 

This PR just add a `static` keyword and refactor the original code to make more readable.  Would you please take a look this PR when you have time?  Thanks.

djwatson:
I can't see why this would fix the issue, but I don't see any problems with changing it to static.  Will merge, thanks.

PR #112, opened 2019-04-15
Added braces to suppress empty if/else warnings

stephen-roberts-work:
This change fixes all empty if/else block warnings from gcc.
It does not change the functionality of libunwind in any way.


djwatson:
Thanks!

PR #117, opened 2019-04-30
Fix format specifier for int64_t:29

am11:
When width is less than 32 (i.e. 29 in this case), the preferred format specifier is `%i`.

clang:
> x86_64/Gstash_frame.c:113:47: warning: format specifies type 'long' but the argument has type 'int' [-Wformat]

gcc:
> x86_64/Gstash_frame.c:113:15: warning: format ‘%li’ expects argument of type ‘long int’, but argument 3 has type ‘int’ [-Wformat=]


djwatson:
Thanks!

PR #120, opened 2019-06-03
x86_64: Add fixup code if previous RIP was invalid

yaspoon:
Previous RIP could become invalid if a bad function pointer
was followed. Calling unwind from a signal frame in this case
skips the frame that called the bad pointer.

I added a test and fixed up the code so that it works with the test-ptrace test as advised in pull request #106 finally after 3 months.
I've been rebasing my code without thinking. So now I can't push to the original branch in the pull request. So I've instead created this pull request. All original information can be found in pull request #106 

Any advise on whether I'm setting up the cursor values correctly would be greatly appreciated as it's quite confusing if I've doing it correctly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,5 +76,6 @@ tests/[GL]ia64-test-stack
 tests/ia64-test-dyn1
 tests/ia64-test-sig
 tests/[GL]x64-test-dwarf-expressions
+tests/x64-unwind-badjmp-signal-frame
 tests/*.log
 tests/*.trs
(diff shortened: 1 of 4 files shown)


yaspoon:
I can see the build is failing how do I get the test-suite.log output to see what went wrong? I'm assuming that some frame names are different to what is expected. Do you have any suggestions? Maybe I should do a backtrace before following the jump and comparing the before and after backtraces as a better test?

yaspoon:
I think I've figured out the problem. The version of gcc used in Ubuntu 14.04 (Or maybe it's just because it's Ubuntu and now Arch) doesn't seem to use or initialise RBP at all. Which means the code path I've added never gets hit so the test fails. I think I've got a fix for it by moving the invalid previous RIP check up higher.

PR #121, opened 2019-06-05
RFC: README: add libc requirement description

yousong:
I made this table when trying to fix [recent packaging issues in OpenWrt](https://github.com/openwrt/packages/issues/8548#issuecomment-498266597) .  It could help track progress on musl-libc support and may also help other distros.

The reason why subject of this pull request is prefixed with a RFC is that I am not sure about ppc32/ppc64's dependency on setcontext().

 - both of them provides _UI_setcontext assembly implmenetation, but I cannot see how this symbol could be get used in the code
 - ppc32 has no refs to setcontext() call
 - ppc64 calls setcontext() explicitly

### DIFF
--- a/README
+++ b/README
@@ -22,6 +22,32 @@ several architecture/operating-system combinations:
 | FreeBSD | x86          | ✓      |
 | FreeBSD | AArch64      | ✓      |
 
+## Libc Requirements
+
+libunwind depends on getcontext(), setcontext() functions which are missing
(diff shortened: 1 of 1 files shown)


djwatson:
I believe all the 'setcontext' calls are only for users to use as a faster version (that doesn't save signals) vs. glibc's implementation.  I think libunwind only uses getcontext. 

I'm happy to merge this, more documentation is always nice.

PR #122, opened 2019-06-07
Older systems (e.g. RHEL5) do not have pipe2(). Add a check and compatibility code.

plopresti:


plopresti:
This is the only patch I need to get libunwind working on RHEL5, which I (unfortunately) still have a need to support.

djwatson:
Merged, thanks!

PR #131, opened 2019-09-10
arm: clear thumb bit in $PC before setting it to $IP

simpleton:
PR #55 doesn't skip non-signal frame correctly. In general, arm32 has two instruction modes. The least-significant bit of the address is used to indicate which mode the function is using.

if we're using thumb mode, `-1` is not enough for moving to previous frame

### DIFF
--- a/src/dwarf/Gparser.c
+++ b/src/dwarf/Gparser.c
@@ -440,8 +440,15 @@ fetch_proc_info (struct dwarf_cursor *c, unw_word_t ip)
      continue, and it's important we get this right, as 'ip' could be
      right at the function entry and hence FDE edge, or at instruction
      that manipulates CFA (push/pop). */
+
   if (c->use_prev_instr)
-    --ip;
+    {
+#if defined(__arm__)
+      /* On arm, the least bit denotes thumb/arm mode, clear it. */
+      ip &= ~(unw_word_t)0x1;
+#endif
+      --ip;
+    }
 
   memset (&c->pi, 0, sizeof (c->pi));
 
--- a/src/arm/Gstep.c
+++ b/src/arm/Gstep.c
@@ -46,7 +46,8 @@ arm_exidx_step (struct cursor *c)
   c->dwarf.loc[UNW_ARM_R15] = DWARF_NULL_LOC;
   unw_word_t ip = c->dwarf.ip;
   if (c->dwarf.use_prev_instr)
-    --ip;
+    /* The least bit denotes thumb/arm mode, clear it. */
+    ip = (ip & ~(unw_word_t)0x1) - 1;
 
   /* check dynamic info first --- it overrides everything else */
   ret = unwi_find_dynamic_proc_info (c->dwarf.as, ip, &c->dwarf.pi, 1,
--- a/src/mi/Gget_proc_name.c
+++ b/src/mi/Gget_proc_name.c
@@ -106,7 +106,15 @@ unw_get_proc_name (unw_cursor_t *cursor, char *buf, size_t buf_len,
   ip = tdep_get_ip (c);
 #if !defined(__ia64__)
   if (c->dwarf.use_prev_instr)
-    --ip;
+    {
+#if defined(__arm__)
+      /* On arm, the least bit denotes thumb/arm mode, clear it. */
+      ip &= ~(unw_word_t)0x1;
+#endif
+      --ip;
+    }
+
+
 #endif
   error = get_proc_name (tdep_get_as (c), ip, buf, buf_len, offp,
                          tdep_get_as_arg (c));


src/dwarf/Gparser.c
```diff
@@ -440,8 +440,13 @@ fetch_proc_info (struct dwarf_cursor *c, unw_word_t ip)
      continue, and it's important we get this right, as 'ip' could be
      right at the function entry and hence FDE edge, or at instruction
      that manipulates CFA (push/pop). */
-  if (c->use_prev_instr)
+  if (c->use_prev_instr) {
+#if defined(__arm__)
```
milianw: afaik this won't work when unwinding an arm stack copy on a x86 machine, or? this is required for perf record on arm followed by perf report on x86 e.g.

i.e. instead of a defined, this need to check the source ABI somehow?

simpleton: Good to know! thanks @milianw 
I will clean up the thumb flag in other place where only effect arm device

simpleton:
Hi @milianw ,how about this new way to clear thumb bit?

simpleton:
Hi @milianw , thanks for detailed explanation. On Android ARM device, libunwind needs this PR to unwind correctly, otherwise it might stick at one frame can't unwind to previous one. such as:
```
Z10abort_hookv+24 in /data/app/xxx/lib/arm/libaborthooks.so+3AA5
 _ZN15google_breakpad16ExceptionHandler16getMinidumpFlagsEv+0 in /data/app/xxx/lib/arm/libbreakpad.so+1E52F
 _ZN15google_breakpad16ExceptionHandler16getMinidumpFlagsEv+0 in /data/app/xxx/lib/arm/libbreakpad.so+1E52F
 _ZN15google_breakpad16ExceptionHandler16getMinidumpFlagsEv+0 in /data/app/xxx/lib/arm/libbreakpad.so+1E52F
 _ZN15google_breakpad16ExceptionHandler16getMinidumpFlagsEv+0 in /data/app/xxx/lib/arm/libbreakpad.so+1E52F
 _ZN15google_breakpad16ExceptionHandler16getMinidumpFlagsEv+0 in /data/app/xxx/lib/arm/libbreakpad.so+1E52F
 _ZN15google_breakpad16ExceptionHandler16getMinidumpFlagsEv+0 in /data/app/xxx/lib/arm/libbreakpad.so+1E52F
 _ZN15google_breakpad16ExceptionHandler16getMinidumpFlagsEv+0 in /data/app/xxx/lib/arm/libbreakpad.so+1E52F
```

I saw Google fixed this issue via [adjust_ip](https://android.googlesource.com/platform/external/libunwind/+/efb75a0b84fed700173700a2e4fee153ba491b32%5E%21/src/arm/Gstep.c), but I think it's good to clear up thumb mode when set $ip

yuyichao:
Does this mean that the return value from `unw_step` will be missing the instruction set info? Won't that cause an issue if the caller actually want to use this info to continue execution?

(Disclaimer, I don't use libunwind that way and I'm not 100% if this is the function used in that case)

simpleton:
@yuyichao good point! I search the code and don't find anyone use in that way. But it's better to keep all the information.

I think do `adjust_ip` when we try to `--ip;` is much safer. `& ~1` should be ok even the address is not for Thumb mode.

simpleton:
Hey @milianw , can you educate me or point some code snippet on how to check the source ABI? thanks a lot.

I'm not sure which(two commits in this PR) is the best way, but over all adjust before move to previous instruction is safer in my mind. 

cc @djwatson , @kostikbel for put more eyeball on this.

src/dwarf/Gparser.c
kostikbel: Generally, when the libunwind is built for specific target architecture, the UNW_TARGET_<something> is defined.  Look into the source for its appearance to see how to use it.

That said (and this is probably a generic note for most load/store arches), wouldn't it be better to decrement by 4, and by 2 for thumb mode, on arm ?

simpleton: @kostikbel yup, minus 2/4 look better to me too. But `-1` looks like is not enough for all the arch, but it actually move the `$IP` to the right address range. It's kinda of enough. I'm quite neutral on this.

djwatson:
Bump - do we still want this, or should I close it out?

simpleton:
I think libunwind needs it, even there're several forked version of libunwind have similar fixing. such as `-2` or clear thumb mode. cc @milianw @kostikbel

kostikbel:
I think that '-1' patch is good enough.  No need to go with precise -2/-4.

simpleton:
> I think that '-1' patch is good enough. No need to go with precise -2/-4.

Not really. `-1` is not enough for going back to previous frame in thumb mode. It will only clear up the thumb mode flag. There're two choice here:
1. clear up thumb mode in [`unw_step`](https://github.com/libunwind/libunwind/pull/131/commits/7d5b195ed918f953a4bf8d249314488e60476dde). 

2. minus one more or clear thumb mode before `--ip;`

kostikbel:
> > I think that '-1' patch is good enough. No need to go with precise -2/-4.
> 
> Not really. `-1` is not enough for going back to previous frame in thumb mode. It will only clear up the thumb mode flag. There're two choice here:
> 
>     1. clear up thumb mode in [`unw_step`](https://github.com/libunwind/libunwind/pull/131/commits/7d5b195ed918f953a4bf8d249314488e60476dde).
> 
>     2. minus one more or clear thumb mode before `--ip;`

By '-1' I mean the patch preent in the pull request we are discussing.  I believe it is correct and good enough.

simpleton:
Hi @kostikbel , sorry for misunderstanding your comment. I rebase and delete one commit. Current PR is ready for reviewing. cc @djwatson @milianw

djwatson:
lgtm

PR #132, opened 2019-09-19
style: use space to instead of tab in Gparser.c

simpleton:
I'm not sure the code style of libunwind, but mixing tab and space is not good practice.
It will display code like:

Snip20190919_7

 I transfer tab to space due to the most lines in `Gparser.c` are using space.


milianw:
I'm in favor, @djwatson has to decide. in general, it would probably make a lot of sense to incorporate a .clang-format with some predefined standard and reformat the source, then add a pre-commit hook to prevent it from going astray in the future

simpleton:
@milianw Total agree!

djwatson:
Yea, let’s do a global clang format instead.

PR #133, opened 2019-09-19
arm: initialize use_prev_instr to 0 in handle_signal_frame

simpleton:
ref: `c->use_prev_instr` use in `fetch_proc_info`
  /* The 'ip' can point either to the previous or next instruction
     depending on what type of frame we have: normal call or a place
     to resume execution (e.g. after signal frame).

     For a normal call frame we need to back up so we point within the
     call itself; this is important because a) the call might be the
     very last instruction of the function and the edge of the FDE,
     and b) so that run_cfi_program() runs locations up to the call
     but not more.

     For signal frame, we need to do the exact opposite and look
     up using the current 'ip' value.  That is where execution will
     continue, and it's important we get this right, as 'ip' could be
     right at the function entry and hence FDE edge, or at instruction
     that manipulates CFA (push/pop). */

if I read this correctly, all the archs need this initialization. But
I heavily use and test on arm & aarch64. I'm not very confidence on
other arch.

### DIFF
--- a/src/arm/Gos-linux.c
+++ b/src/arm/Gos-linux.c
@@ -118,6 +118,7 @@ arm_handle_signal_frame (unw_cursor_t *cursor)
   dwarf_get (&c->dwarf, c->dwarf.loc[UNW_ARM_R15], &c->dwarf.ip);
 
   c->dwarf.pi_valid = 0;
+  c->dwarf.use_prev_instr = 0;
 
   return 1;
 }


src/arm/Gos-linux.c
```diff
@@ -118,6 +118,7 @@ arm_handle_signal_frame (unw_cursor_t *cursor)
   dwarf_get (&c->dwarf, c->dwarf.loc[UNW_ARM_R15], &c->dwarf.ip);
 
   c->dwarf.pi_valid = 0;
+  c->dwarf.use_prev_instr = 0;
```
milianw: so far the init seems to be done in the platform-specific init.h header. either this code here is duplicating that, or the other code can now be removed?

simpleton: Hi @milianw , thanks for reply. 
> so far the init seems to be done in the platform-specific init.h header

Can you point me the code? I only see the api support pass the `use_prev_instr` as a parameter.

I think this's necessary if the library can handle this case, there's no strong easy to let user to handle this. Btw, I saw the `aarch64` has done this in [`handle_signal_frame`](https://github.com/libunwind/libunwind/blob/55ef85c50e4184c98c03e0c52ffb1c39b5201fab/src/aarch64/Gstep.c#L128)

milianw: see e.g. `src/arm/init.h:70` or `src/x86_64/init.h:81` etc.

simpleton: @milianw thanks for pointing the code. This initialization makes me a little bit confusing.

The local unwind need caller to tell libunwind whether the frame is signal frame via passing `UNW_INIT_SIGNAL_FRAME` as extend parameter. My question is: if we can determine the frame via `unw_is_signal_frame(1)`, why we don't hide this detail for users?

milianw: I dunno, I have no clue about this code. I'm just saying that we don't need to init `use_prev_instr` like done in your patch, since it seems to be done already elsewhere?

simpleton: Thx for reply. No, not really. For remote mode, it's true. Because it never run into signal handler. But for local mode, it just let user to pass a flag to determine the value of f`use_prev_instr `

https://github.com/libunwind/libunwind/blob/d32956507cf29d9b1a98a8bce53c78623908f4fe/src/arm/Ginit_local.c#L62-L71

djwatson: Yea I agree this looks unnecessary.  Is there some argument for keeping it?

simpleton:
It's redundant with `UNW_INIT_SIGNAL_FRAME` flag.

PR #153, opened 2020-01-21
arm: cache cb_data which contains libunwind

clamp03:
- Too many calls to arm_phdr_cb even though it gets the same data
- Cache ARM_CB_DATA for shared object which contains libunwind, then reuse

### DIFF
--- a/include/tdep-arm/libunwind_i.h
+++ b/include/tdep-arm/libunwind_i.h
@@ -254,6 +254,7 @@ dwarf_put (struct dwarf_cursor *c, dwarf_loc_t loc, unw_word_t val)
 #define tdep_init                       UNW_OBJ(init)
 #define arm_find_proc_info              UNW_OBJ(find_proc_info)
 #define arm_put_unwind_info             UNW_OBJ(put_unwind_info)
+#define arm_cb_cache_data_init          UNW_OBJ(cb_cache_data_init)
 /* Platforms that support UNW_INFO_FORMAT_TABLE need to define
    tdep_search_unwind_table.  */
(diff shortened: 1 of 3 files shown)


djwatson:
It looks like this is a cache for the dl_iterate_phdr info? I'm probably okay with this - it needs to be hooked up to clear cache also, right?  I think we already cache info in some places such that dlclose doesn't really work correctly

clamp03:
@djwatson Thanks for your comment. Yes, it is a cache for the dl_iterate_phdr info. I don't think it needs cache clear because it caches a data for a shared object which built with libunwind like dotnet runtime. If I misunderstand, please let me know. Thank you!

djwatson:
What happens if someone calls dlclose() after a dlopen() of something not from the .net runtime?

clamp03:
It caches data for a file that contains a libunwind.
- If someone calls dlclose() for A.so file (not .net runtime), cached data is valid because it cached .net runtime.
- If someone calls dlclose() for .net runtime(.so file), cached data of libunwind as well as libunwind in .net runtime are removed. (libunwind source is built with .net runtime.)
- When libunwind is used as .so file and B.so file use libunwind.so, this caches data of a libunwind.so. So if someone calls dlclose()  B.so or libunwind.so, it is okay.
One disadvantage is that it always checks with cached data in dll_iterate_phdr. I think it is not a huge overhead in most cases.

Sorry for my broken English. If I misunderstand, please let me know.
Thank you.

clamp03:
I think it is better to apply for my own project.
Thank you for the review.

PR #157, opened 2020-02-06
Fix compilation with gcc 10 (which defaulted to -fno-common)

### LINKED ISSUE #154: libunwind fails to build on gcc-10 (or on CC="gcc -fno-common")
It's an upstream version of downstream bug: https://bugs.gentoo.org/706560

gcc-10 will change the default from -fcommon to fno-common: https://gcc.gnu.org/PR85678.
The error also happens if CFLAGS=-fno-common passed explicitly.

On gcc-10 (and gcc-9 -fno-common) libunwind build fails as:

```
libtool: link: gcc -shared  -fPIC -DPIC  .libs/os-linux.o mi/.libs/init.o mi/.libs/flush_cache.o mi/.libs/mempool.o mi/.libs/strerror.o x86_64/.libs/is_fpreg.o x86_64/.libs/regname.o x86_64/.libs/Los-linux.o mi/.libs/backtrace.o mi/.libs/dyn-cancel.o mi/.libs/dyn-info-list.o mi/.libs/dyn-register.o mi/.libs/Ldyn-extract.o mi/.libs/Lfind_dynamic_proc_info.o mi/.libs/Lget_accessors.o mi/.libs/Lget_proc_info_by_ip.o mi/.libs/Lget_proc_name.o mi/.libs/Lput_dynamic_unwind_info.o mi/.libs/Ldestroy_addr_space.o mi/.libs/Lget_reg.o mi/.libs/Lset_reg.o mi/.libs/Lget_fpreg.o mi/.libs/Lset_fpreg.o mi/.libs/Lset_caching_policy.o mi/.libs/Lset_cache_size.o x86_64/.libs/setcontext.o x86_64/.libs/Lapply_reg_state.o x86_64/.libs/Lreg_states_iterate.o x86_64/.libs/Lcreate_addr_space.o x86_64/.libs/Lget_save_loc.o x86_64/.libs/Lglobal.o x86_64/.libs/Linit.o x86_64/.libs/Linit_local.o x86_64/.libs/Linit_remote.o x86_64/.libs/Lget_proc_info.o x86_64/.libs/Lregs.o x86_64/.libs/Lresume.o x86_64/.libs/Lstash_frame.o x86_64/.libs/Lstep.o x86_64/.libs/Ltrace.o x86_64/.libs/getcontext.o  -Wl,--whole-archive ./.libs/libunwind-dwarf-local.a ./.libs/libunwind-elf64.a -Wl,--no-whole-archive  -lc -lgcc_s -llzma -lz  -g -O2 -nostartfiles -nostdlib   -Wl,-soname -Wl,libunwind.so.8 -o .libs/libunwind.so.8.0.1
/usr/lib/gcc/x86_64-pc-linux-gnu/10.0.1/../../../../x86_64-pc-linux-gnu/bin/ld: x86_64/.libs/Linit.o:(.bss+0x0): multiple definition of `_U_dyn_info_list'; mi/.libs/dyn-info-list.o:(.bss+0x0): first defined here
collect2: error: ld returned 1 exit status
```

Looks like `_U_dyn_info_list` definition site is tricky even on amd64

```
$ git grep _U_dyn_info_list | fgrep list_t
include/libunwind_i.h:extern unw_dyn_info_list_t _U_dyn_info_list;
src/mi/dyn-info-list.c:HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
src/x86/Ginit.c:HIDDEN unw_dyn_info_list_t _U_dyn_info_list;

src/x86/Linit.c:#include "Ginit.c"
```

Should it perhaps be consolidated in a single place?

stac47:
I propose this to fix #154

### DIFF
--- a/src/mi/dyn-info-list.c
+++ b/src/mi/dyn-info-list.c
@@ -25,7 +25,7 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 
 #include "libunwind_i.h"
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
+extern HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
 
 unw_word_t
 _U_dyn_info_list_addr (void)


yuyichao:
I always find it confusing which file is included in which but is this a x86 only file? If not, I assume the one in x86 should be turned to a declaration instead.

stac47:
I would say each definition are done in the Ginit.c for the targeted architecture. The modified file is used to get the address of the structure and not to define it.

Romain-Geissler-1A:
Hi,

I think the extern shall be in libunwind_i.h I don't know if it also needs to be in this .c file, but for sure the .h file is included from at least two places in x86 leading to such kind of errors:

```
ld.lld: error: duplicate symbol: _U_dyn_info_list
>>> defined at libunwind_i.h:265 (../include/libunwind_i.h:265)
>>>            Ginit.o:(_U_dyn_info_list) in archive /data/mwrep/res/osp/Libunwind/Generic/20-0-0-0/lib/x86_64-v20/Debug/libunwind-x86_64.a
>>> defined at libunwind_i.h:265 (../include/libunwind_i.h:265)
>>>            Linit.o:(.bss+0x0) in archive /data/mwrep/res/osp/Libunwind/Core/20-0-0-0/lib/x86_64-v20/Debug/libunwind.a
collect2: error: ld returned 1 exit status
```

Cheers,
Romain

Romain-Geissler-1A:
Hi Laurent,

It looks like Linit.c can in some cases actually include Ginit.c, and have partially the same code exposed. So I think we should instead keep this symbol declared "extern" in Ginit.c, and really define it in dyn-info-list.c

When I try this, I get this symbol visibility error:

> libtool: link: gcc -shared  -fPIC -DPIC  coredump/.libs/_UCD_accessors.o coredump/.libs/_UCD_create.o coredump/.libs/_UCD_destroy.o coredump/.libs/_UCD_access_mem.o coredump/.libs/_UCD_elf_map_image.o coredump/.libs/_UCD_find_proc_info.o coredump/.libs/_UCD_get_proc_name.o coredump/.libs/_UPT_elf.o coredump/.libs/_UPT_access_fpreg.o coredump/.libs/_UPT_get_dyn_info_list_addr.o coredump/.libs/_UPT_put_unwind_info.o coredump/.libs/_UPT_resume.o coredump/.libs/_UCD_access_reg_linux.o    -ggdb3 -nostartfiles -fno-lto -ggdb3 -Wl,--as-needed   -Wl,-soname -Wl,libunwind-coredump.so.0 -o .libs/libunwind-coredump.so.0.0.0
> 16:24:44  /opt/1A/toolchain/x86_64-v20.0.12/lib/gcc/x86_64-1a-linux-gnu/10.0.1/../../../../x86_64-1a-linux-gnu/bin/ld: x86_64/.libs/Ginit.o: in function `get_dyn_info_list_addr':
> 16:24:44  /home/jenkins/workspace/OSP_osp-libunwind_release_20.0.0/BUILD/libunwind-1.3.1/src/x86_64/Ginit.c:69: undefined reference to `_U_dyn_info_list'
> 16:24:44  /opt/1A/toolchain/x86_64-v20.0.12/lib/gcc/x86_64-1a-linux-gnu/10.0.1/../../../../x86_64-1a-linux-gnu/bin/ld: x86_64/.libs/Ginit.o: relocation R_X86_64_PC32 against undefined hidden symbol `_U_dyn_info_list' can not be used when making a shared object
> 16:24:44  /opt/1A/toolchain/x86_64-v20.0.12/lib/gcc/x86_64-1a-linux-gnu/10.0.1/../../../../x86_64-1a-linux-gnu/bin/ld: final link failed: bad value
> 16:24:44  collect2: error: ld returned 1 exit status
> 16:24:44  make[2]: *** [Makefile:2409: libunwind-x86_64.la] Error 1
> 16:24:44  make[2]: *** Waiting for unfinished jobs....

So I think we shall also remove the "HIDDEN" keyword. In the end, since the symbol is already declared "extern" in libunwind_i.h, already included in Ginit.c, I see no point of re-declaring it in Ginit.c.

Internally in Amadeus I modified the patched you did checked it initially (following this PR you did), and checked out this patch in libunwind instead:

```
--- a/src/x86_64/Ginit.c
+++ b/src/x86_64/Ginit.c
@@ -49,7 +49,6 @@

 unw_addr_space_t unw_local_addr_space = &local_addr_space;

-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;

 /* XXX fix me: there is currently no way to locate the dyn-info list
        by a remote unwinder.  On ia64, this is done via a special
--- a/src/mi/dyn-info-list.c
+++ b/src/mi/dyn-info-list.c
@@ -25,7 +25,7 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */

 #include "libunwind_i.h"

-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
+unw_dyn_info_list_t _U_dyn_info_list;

 unw_word_t
 _U_dyn_info_list_addr (void)
```

I checked that in Amadeus the build of libunwind works with that patch. Also, internal Amadeus program linked against the static version of libunwind now do not raise any error about duplicated symbol.

If this fix flies for @yuyichao then please Laurent update this contribution to make it acceptable upstream (ie do not do the fix only for x64, but for all targets).

Cheers,
Romain

yuyichao:
The location where you decide to put it makes sense though I have to say I haven't really traced all the include paths so I'm taking your word for `dyn-info-list.c` being compiled only once...

I don't think we should remove "hidden" though. The visibililty here basically is hiding it from external users of libunwind and I think that is still desirable.

Romain-Geissler-1A:
Not removing "hidden" means that it won't work when using shared libraries (libunwind itself doesn't even build). As soon as the same symbol is used cross libraries, when using shared libraries, it can't be hidden.

yuyichao:
Is that variable ever used in other library? AFAICT this is an internal variable for libunwind and no externall user should be using it. In fact, I believe it's declared in internal header rather than a public header.

yuyichao:
I searched my local `libunwind` installation and I believe `_U_dyn_info_list` is only available as a local symbol in `libunwind-x86_64.so` and is not used anywhere else. This agrees with what I understand the code is intended to do.

When you see the error above, did it get fixed by removing the hidden everywhere. If you did that, did you check that the library actually works, that `_U_dyn_info_list` is actually defined somewhere, and that it's used in more than one places?

Romain-Geissler-1A:
Well I honestly fail to understand how generally libunwind is designed from a lib point of view (ie which lib shall have which symbol).

I see that for many symbol you actually prefix them by platform, and sometimes the "local" version is also prefixed by the letter "L" leading to no symbol conflict. For `_U_dyn_info_list` it doesn't look to be the case thought, I don't know if this is expected or not.

For this specific symbol, when I do "nm" our own deliveries of libunwind, I see the symbol defined both in `libunwind.so` and `libunwind-x86_64.so` (also called `libunwind-generic.so`), the latter depends on the first one:

```
rgeissler@ncerndobedev6097:/projects/mwdeldev/osp/Libunwind/Core/19-0-0-10/lib/x86_64-v19/Debug> for i in *; do echo $i; nm $i |grep "\<_U_dyn_info_list\>"; done
libunwind.a
                 U _U_dyn_info_list
0000000000000010 C _U_dyn_info_list
                 U _U_dyn_info_list
0000000000000010 C _U_dyn_info_list
libunwind.so
00000000000181f0 b _U_dyn_info_list
libunwind.so.19.0.0.10
00000000000181f0 b _U_dyn_info_list
libunwind.so.8   
00000000000181f0 b _U_dyn_info_list
libunwind.so.8.0.1
00000000000181f0 b _U_dyn_info_list

rgeissler@ncerndobedev6097:/projects/mwdeldev/osp/Libunwind/Core/19-0-0-10/lib/x86_64-v19/Debug> cd -
/projects/mwdeldev/osp/Libunwind/Generic/19-0-0-10/lib/x86_64-v19/Debug

rgeissler@ncerndobedev6097:/projects/mwdeldev/osp/Libunwind/Generic/19-0-0-10/lib/x86_64-v19/Debug> for i in *; do echo $i; nm $i |grep "\<_U_dyn_info_list\>"; done
libunwind-generic.a
0000000000000010 C _U_dyn_info_list
libunwind-generic.so
0000000000020870 b _U_dyn_info_list
libunwind-x86_64.a
0000000000000010 C _U_dyn_info_list
libunwind-x86_64.so
0000000000020870 b _U_dyn_info_list
libunwind-x86_64.so.19.0.0.10
0000000000020870 b _U_dyn_info_list
libunwind-x86_64.so.8
0000000000020870 b _U_dyn_info_list
libunwind-x86_64.so.8.0.1
0000000000020870 b _U_dyn_info_list
```

Since both libraries seem to define (and access) the same symbol, I don't see how you can't have it public rather than hidden. But maybe it's just not expected that both libraries access the same symbol ?

yuyichao:
> I see that for many symbol you actually prefix them by platform, and sometimes the "local" version is also prefixed by the letter "L" leading to no symbol conflict. For _U_dyn_info_list it doesn't look to be the case thought, I don't know if this is expected or not.

I believe this is so that different version (for different arch) does not have conflicting name. Anything that doesn't have an arch prefix are local variables to that arch/build that shouldn't be shared with other build. This is consistent with the fact that `_U_dyn_info_list` being hidden. In fact, I believe it means that `_U_dyn_info_list` must be hidden.

> Since both libraries seem to define (and access) the same symbol, I don't see how you can't have it public rather than hidden. But maybe it's just not expected that both libraries access the same symbol ?

They do not define and access the same symbol. They merely have defined and access local symbol with the same name. (same symbol as in same addreess after linking).

Also, the fact that it works like this before, while it doesn't mean the variable can't be shared, it at least mean it's definitely not the problem in your compilation. (`-fno-common` should't change the behavior regardingn missing local symbol)

Romain-Geissler-1A:
> They do not define and access the same symbol. They merely have defined and access local symbol with the same name. (same symbol as in same addreess after linking).

Which is okay for dynamic linking, but it not for static linking (in which symbol visibility is not taken into account). And here in static mode it used to work with`-fcommon` (which was the default until gcc 10), but not never with `-fno-common`. If you really want to have two distinct symbols as you seem to write it, is it ok to simply rename them to avoid name conflict ? Right now, before the gcc change you have different behavior:
 - in static linking, the linker merged the two distinct symbols into one
 - in dynamic linking, the linker created two distinct private symbols with the same name in two distinct libraries.
which seems a bit odd. Since I have no idea on what is the actual behavior you expected (two symbols vs one shared), I don't know how to fix it.

PS: It happens that we use lib unwind both in static and shared mode (Laurent and I work in the team responsible for the whole C/C++ build in Amadeus, we have several use cases, some of them using shared linking, others using static linking.).

yuyichao:
> is it ok to simply rename them to avoid name conflict ? Right now, before the gcc change you have different behavior:

I think that's fair. It should not change any behavior for dynamically linked library (which honestly is the only one I personally care about). It might break someone's code using static linking if hey were illegally accessing this variable before or sth like that. You seems to have cases with static linking so it seems that you are in a good position to test this.

I still think the `HIDDEN` should be there though, since it seems that the variable is not meant to be accessed elsewhere.

yuyichao:
> in dynamic linking, the linker created two distinct private symbols with the same name in two distinct libraries. which seems a bit odd.

Note that this isn't particularly odd. It happens all the time with `static` symbols. After linking the hidden symbol bascially looks the same as a static (local) one IIRC.

Romain-Geissler-1A:
I tested the following patch both in static and shared link, it raises no symbol error of any kind with gcc 10 and ld.lld:

```
--- a/src/x86_64/Ginit.c
+++ b/src/x86_64/Ginit.c
@@ -49,7 +49,7 @@

 unw_addr_space_t unw_local_addr_space = &local_addr_space;

-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
+static unw_dyn_info_list_t _Ux86_64_dyn_info_list;

 /* XXX fix me: there is currently no way to locate the dyn-info list
        by a remote unwinder.  On ia64, this is done via a special
@@ -66,7 +66,7 @@
 get_dyn_info_list_addr (unw_addr_space_t as, unw_word_t *dyn_info_list_addr,
                         void *arg)
 {
-  *dyn_info_list_addr = (unw_word_t) &_U_dyn_info_list;
+  *dyn_info_list_addr = (unw_word_t) &_Ux86_64_dyn_info_list;
   return 0;
 }

```

I can ask Laurent to submit a real pull request changing all targets the same way if you are ok with that.

Romain-Geissler-1A:
I has to be noted that we do use libunwind in the test library I used, only to use things like unw_backtrace, unw_get_accessors and address_space->get_proc_name(...)

yuyichao:
No that is not correct. It simply make the variable local to this file and cause all other use of the variable to use the other instance. You now have two version of this variable in each shared library for the same arch and build. That is not what you want.

To do this correctly, you need to make a patch without renaming that make the build pass without `-fcommon`. This will make sure new version behaves exactly as before, static or dynamic. Then you can rename the variable, using the macro (`UNW_OBJ`?) to do the symbol renaming rather than doing it manually for each arch, which should change the behavior for static library to match that of dynamic library without channging the behavior for dynamic library. `HIDDEN` should never be removed in any steps and nothing should be changed to `static` either.

yuyichao:
```patch
diff --git a/src/Makefile.am b/src/Makefile.am
index ff977446..f89ad409 100644
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -99,7 +99,8 @@ libunwind_la_LIBADD =
 # libraries:
 libunwind_la_SOURCES_common =					\
 	$(libunwind_la_SOURCES_os)				\
-	mi/init.c mi/flush_cache.c mi/mempool.c mi/strerror.c
+	mi/init.c mi/dyn-info-list.c mi/flush_cache.c		\
+	mi/mempool.c mi/strerror.c
 
 # List of arch-independent files needed by generic library (libunwind-$ARCH):
 libunwind_la_SOURCES_generic =						\
@@ -131,7 +132,7 @@ endif
 libunwind_la_SOURCES_local_nounwind =					\
 	$(libunwind_la_SOURCES_os_local)				\
 	mi/backtrace.c							\
-	mi/dyn-cancel.c mi/dyn-info-list.c mi/dyn-register.c		\
+	mi/dyn-cancel.c mi/dyn-register.c				\
 	mi/Ldyn-extract.c mi/Lfind_dynamic_proc_info.c			\
 	mi/Lget_accessors.c						\
 	mi/Lget_proc_info_by_ip.c mi/Lget_proc_name.c			\
diff --git a/src/aarch64/Ginit.c b/src/aarch64/Ginit.c
index dec235c8..b0b5be89 100644
--- a/src/aarch64/Ginit.c
+++ b/src/aarch64/Ginit.c
@@ -61,8 +61,6 @@ tdep_uc_addr (unw_tdep_context_t *uc, int reg)
 
 # endif /* UNW_LOCAL_ONLY */
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
-
 /* XXX fix me: there is currently no way to locate the dyn-info list
        by a remote unwinder.  On ia64, this is done via a special
        unwind-table entry.  Perhaps something similar can be done with
diff --git a/src/arm/Ginit.c b/src/arm/Ginit.c
index 2720d063..a843511d 100644
--- a/src/arm/Ginit.c
+++ b/src/arm/Ginit.c
@@ -57,8 +57,6 @@ tdep_uc_addr (unw_tdep_context_t *uc, int reg)
 
 # endif /* UNW_LOCAL_ONLY */
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
-
 /* XXX fix me: there is currently no way to locate the dyn-info list
        by a remote unwinder.  On ia64, this is done via a special
        unwind-table entry.  Perhaps something similar can be done with
diff --git a/src/hppa/Ginit.c b/src/hppa/Ginit.c
index 461e4b93..15c7a91d 100644
--- a/src/hppa/Ginit.c
+++ b/src/hppa/Ginit.c
@@ -64,8 +64,6 @@ _Uhppa_uc_addr (ucontext_t *uc, int reg)
 
 # endif /* UNW_LOCAL_ONLY */
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
-
 /* XXX fix me: there is currently no way to locate the dyn-info list
        by a remote unwinder.  On ia64, this is done via a special
        unwind-table entry.  Perhaps something similar can be done with
diff --git a/src/mips/Ginit.c b/src/mips/Ginit.c
index 3df170c7..68a9935e 100644
--- a/src/mips/Ginit.c
+++ b/src/mips/Ginit.c
@@ -69,8 +69,6 @@ tdep_uc_addr (ucontext_t *uc, int reg)
 
 # endif /* UNW_LOCAL_ONLY */
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
-
 /* XXX fix me: there is currently no way to locate the dyn-info list
        by a remote unwinder.  On ia64, this is done via a special
        unwind-table entry.  Perhaps something similar can be done with
diff --git a/src/ppc32/Ginit.c b/src/ppc32/Ginit.c
index ba302448..5425d619 100644
--- a/src/ppc32/Ginit.c
+++ b/src/ppc32/Ginit.c
@@ -91,9 +91,6 @@ tdep_uc_addr (ucontext_t *uc, int reg)
 
 # endif /* UNW_LOCAL_ONLY */
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
-
-
 static void
 put_unwind_info (unw_addr_space_t as, unw_proc_info_t *proc_info, void *arg)
 {
diff --git a/src/ppc64/Ginit.c b/src/ppc64/Ginit.c
index 4c88cd6e..32b87280 100644
--- a/src/ppc64/Ginit.c
+++ b/src/ppc64/Ginit.c
@@ -95,9 +95,6 @@ tdep_uc_addr (ucontext_t *uc, int reg)
 
 # endif /* UNW_LOCAL_ONLY */
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
-
-
 static void
 put_unwind_info (unw_addr_space_t as, unw_proc_info_t *proc_info, void *arg)
 {
diff --git a/src/s390x/Ginit.c b/src/s390x/Ginit.c
index f0886ac9..958c18d2 100644
--- a/src/s390x/Ginit.c
+++ b/src/s390x/Ginit.c
@@ -50,8 +50,6 @@ static struct unw_addr_space local_addr_space;
 
 unw_addr_space_t unw_local_addr_space = &local_addr_space;
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
-
 static inline void *
 uc_addr (ucontext_t *uc, int reg)
 {
diff --git a/src/sh/Ginit.c b/src/sh/Ginit.c
index 52988a72..a17dd91f 100644
--- a/src/sh/Ginit.c
+++ b/src/sh/Ginit.c
@@ -58,8 +58,6 @@ tdep_uc_addr (ucontext_t *uc, int reg)
 
 # endif /* UNW_LOCAL_ONLY */
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
-
 /* XXX fix me: there is currently no way to locate the dyn-info list
        by a remote unwinder.  On ia64, this is done via a special
        unwind-table entry.  Perhaps something similar can be done with
diff --git a/src/tilegx/Ginit.c b/src/tilegx/Ginit.c
index 7564a558..65609044 100644
--- a/src/tilegx/Ginit.c
+++ b/src/tilegx/Ginit.c
@@ -64,8 +64,6 @@ tdep_uc_addr (ucontext_t *uc, int reg)
 
 # endif /* UNW_LOCAL_ONLY */
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
-
 /* XXX fix me: there is currently no way to locate the dyn-info list
        by a remote unwinder.  On ia64, this is done via a special
        unwind-table entry.  Perhaps something similar can be done with
diff --git a/src/x86/Ginit.c b/src/x86/Ginit.c
index f6b8dc27..5e57f254 100644
--- a/src/x86/Ginit.c
+++ b/src/x86/Ginit.c
@@ -54,8 +54,6 @@ tdep_uc_addr (ucontext_t *uc, int reg)
 
 # endif /* UNW_LOCAL_ONLY */
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
-
 /* XXX fix me: there is currently no way to locate the dyn-info list
        by a remote unwinder.  On ia64, this is done via a special
        unwind-table entry.  Perhaps something similar can be done with
diff --git a/src/x86_64/Ginit.c b/src/x86_64/Ginit.c
index a865d338..5d139bff 100644
--- a/src/x86_64/Ginit.c
+++ b/src/x86_64/Ginit.c
@@ -49,8 +49,6 @@ static struct unw_addr_space local_addr_space;
 
 unw_addr_space_t unw_local_addr_space = &local_addr_space;
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
-
 /* XXX fix me: there is currently no way to locate the dyn-info list
        by a remote unwinder.  On ia64, this is done via a special
        unwind-table entry.  Perhaps something similar can be done with
```

I believe this should be the correct patch to fix the current issue. I tested by adding `-fno-common` on gcc 9.2.1. The renaming of the symbol (i.e. the change of static library behavior should be a separate issue)

yuyichao:
Also @Romain-Geissler-1A  Just to make sure I understand, the error you posted [above](https://github.com/libunwind/libunwind/pull/157#issuecomment-595911035) does not happen when you compile libunwind but only happens when you build something else linking both versions (local and remote?) of the static libunwind right?

If that is the case, I think this still won't fix you issue just yet (this should behave the same as this PR but IMHO should be cleaner for future reader...)

yuyichao:
Hold on, I think this actually changed the dynamic library behavior. Let me check...

djwatson:
bump. Do we still want this?

Romain-Geissler-1A:
Hi,

As far as I know people using gcc 10 (due to be released in one or two months) are still hitting the issue.

Yet I still don't know what the real fix should be, and would be glad if one maintainer who know better than us how the internals are meant to work would have a look. We applied a local patch which work for us, but we would prefer an upstream fix obviously. It's easy to reproduce with any gcc version, by forcing CFLAGS -fno-common.

Cheers,
Romain

djwatson:
I think @yuyichao 's fix is closest to correct, and is what the original ia64 implementation did.

I've put up pull request #166

PR #158, opened 2020-02-08
Fix SmartOS/Solaris build

am11:
` #include <sys/ptrace.h>` was neither available nor required on Solaris/SmartOS.

Also, a bit unfortunately Solaris x86_64 support has been regressed since it was added last year (https://github.com/libunwind/libunwind/commit/edc427a9eccd6db583fd0cd920e2af23b4b544a9), now only 13 out of 33 tests are passing, which was previously 23 out of 32. I have updated `README.md` to reflect the current stats. Here is the test run logs collected on SmartOS x64: [test-suite.log](https://github.com/libunwind/libunwind/files/4173580/test-suite.log).

### DIFF
--- a/README
+++ b/README
@@ -208,9 +208,15 @@ only test programs that are known to work at this time are:
 
 ### Expected results on Solaris x86-64
 
-`make check` is passing 23 out of 32 tests. The following nine tests are consistently
+`make check` is passing 13 out of 33 tests. The following nine tests are consistently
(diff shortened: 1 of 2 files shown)


am11:
@djwatson, thanks! Could you give a guesstimate on the release date? The latest tag is from over an year ago, and quite a of interesting changes are pending release. :)

PR #166, opened 2020-03-30
Fix gcc 10 build (or -fno-common)

djwatson:
Use only a single _U_dyn_info_list variable in the local libunwind library.

### DIFF
--- a/src/aarch64/Ginit.c
+++ b/src/aarch64/Ginit.c
@@ -61,13 +61,6 @@ tdep_uc_addr (unw_tdep_context_t *uc, int reg)
 
 # endif /* UNW_LOCAL_ONLY */
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
-
-/* XXX fix me: there is currently no way to locate the dyn-info list
-       by a remote unwinder.  On ia64, this is done via a special
-       unwind-table entry.  Perhaps something similar can be done with
-       DWARF2 unwind info.  */
-
 static void
 put_unwind_info (unw_addr_space_t as, unw_proc_info_t *proc_info, void *arg)
 {
@@ -78,7 +71,13 @@ static int
 get_dyn_info_list_addr (unw_addr_space_t as, unw_word_t *dyn_info_list_addr,
                         void *arg)
 {
-  *dyn_info_list_addr = (unw_word_t) &_U_dyn_info_list;
+#ifndef UNW_LOCAL_ONLY
+# pragma weak _U_dyn_info_list_addr
+  if (!_U_dyn_info_list_addr)
+    return -UNW_ENOINFO;
+#endif
+  // Access the `_U_dyn_info_list` from `LOCAL_ONLY` library, i.e. libunwind.so.
+  *dyn_info_list_addr = _U_dyn_info_list_addr ();
   return 0;
 }
 
--- a/src/arm/Ginit.c
+++ b/src/arm/Ginit.c
@@ -57,18 +57,17 @@ tdep_uc_addr (unw_tdep_context_t *uc, int reg)
 
 # endif /* UNW_LOCAL_ONLY */
 
-HIDDEN unw_dyn_info_list_t _U_dyn_info_list;
-
-/* XXX fix me: there is currently no way to locate the dyn-info list
-       by a remote unwinder.  On ia64, this is done via a special
-       unwind-table entry.  Perhaps something similar can be done with
-       DWARF2 unwind info.  */
-
 static int
 get_dyn_info_list_addr (unw_addr_space_t as, unw_word_t *dyn_info_list_addr,
                         void *arg)
 {
-  *dyn_info_list_addr = (unw_word_t) &_U_dyn_info_list;
+#ifndef UNW_LOCAL_ONLY
+# pragma weak _U_dyn_info_list_addr
+  if (!_U_dyn_info_list_addr)
+    return -UNW_ENOINFO;
+#endif
+  // Access the `_U_dyn_info_list` from `LOCAL_ONLY` library, i.e. libunwind.so.
+  *dyn_info_list_addr = _U_dyn_info_list_addr ();
   return 0;
 }
 
--- a/src/coredump/_UPT_get_dyn_info_list_addr.c
+++ b/src/coredump/_UPT_get_dyn_info_list_addr.c
@@ -74,6 +74,11 @@ get_list_addr (unw_addr_space_t as, unw_word_t *dil_addr, void *arg,
 
 #else
 
+/* XXX fix me: there is currently no way to locate the dyn-info list
+       by a remote unwinder.  On ia64, this is done via a special
+       unwind-table entry.  Perhaps something similar can be done with
+       DWARF2 unwind info.  */
+
 static inline int
 get_list_addr (unw_addr_space_t as, unw_word_t *dil_addr, void *arg,
                int *countp)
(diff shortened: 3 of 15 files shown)


djwatson:
@yuyichao @Romain-Geissler-1A can you review?  Thanks

Romain-Geissler-1A:
Hi,

I applied this patch on release 1.3.1, regenerated src/Makefile.in (which I initially forgot), and it worked for us both in static and dynamic link.

Thanks ;)
Romain

yuyichao:
I think this might not be the correct patch since there seems to be cases where one copy is used in the other via the function. Hopefully I can check in the next day or two. If not thus can probably be merged and the regression will probably come only in some unusual corner cases...

yuyichao:
Warning: long post ahead....

-----

So I dug through the use of the dynamic unwind info list as well as the history of this. This is as much as I can figure out. There are still quite a few things that doesn't make much sense most of which are from code from 16+ years ago with commit messages that either doesn't make much sense or are not very informative.

I also took this chance to get a better understanding of the local vs remote in libunwind and I'll start with a summary of that. Other people might have a better understanding of it already but up to this point, I've only dealt with one or two cases...

# TL;DR

Whatever we do should be fine here, the duplicated `_U_dyn_info_list` isn't even used. We should be able to change the other arch to be consistent with IA64, which is the cleanest and does not have this problem. The patch is [here](https://github.com/libunwind/libunwind/compare/master...yuyichao:common).

# Long version

So first of all, to set the stage, there are three different versions a source file can be compiled under and they fall under two possible configurations.

1. `REMOTE_ONLY`

    This imply `!LOCAL_ONLY` and is built if the host arch is different from the target arch.

2. `!REMOTE_ONLY`

    This can have both `LOCAL_ONLY` (`libunwind.so`) and `!LOCAL_ONLY` (`libunwind-<arch>.so`)
    In this configuration, a lot of the the source files will be compiled twice one with
    `!REMOTE_ONLY && LOCAL_ONLY` and once with `!REMOTE_ONLY && !LOCAL_ONLY`.
    The `!LOCAL_ONLY` one (`libunwind-<arch>.so`) is linked to the
    `LOCAL_ONLY` one (`libunwind.so`).

For the variable we care about `_U_dyn_info_list` and all the related ones.
It is for dynamically registering unwind info.
The registration function for it is only available in the `LOCAL_ONLY` library (`libunwind.so`)
so only the version of variable for that should be used by both libraries.
Since `mi/dyn-info-list.c` always defines it for the `LOCAL_ONLY` version, that should be the
one we always use in all cases, one way or another.

Here's a list of defs/uses that I can find for the few interesting variables/functions.

* `_U_dyn_info_list`:

    * `<arch>/Ginit.c` (except IA64):

        Defined and used (by `get_dyn_info_list_addr`) for `!REMOTE_ONLY` (both `LOCAL_ONLY` and `!LOCAL_ONLY`)
        As mentioned above, this definition should not be used.
        The use is OK if it is using the one from the `LOCAL_ONLY` library.
        This is impossible right now since `_U_dyn_info_list` is a hidden symbol.

    * `mi/dyn-info-list.c`:

        Defined for `LOCAL_ONLY`. This is the one that should be used.

    * `mi/dyn-canel.c` and `mi/dyn-register.c`:

        Used for `LOCAL_ONLY`. These are correct.

* `_U_dyn_info_list_addr()`:

    This is the function that exports the address of `_U_dyn_info_list` from `mi/dyn-info-list.c`.
    This should only be defined once and it is currently the correct way to access `_U_dyn_info_list` from the `!LOCAL_ONLY` library (`libunwind-<arch>.so`).

    * `mi/dyn-info-list.c`:

        Defined for `LOCAL_ONLY`

    * `mi/Gfind_dynamic_proc_info.c` and `ia64/Ginit.c`:

        Used as weak symbol for `!REMOTE_ONLY && !LOCAL_ONLY` (i.e. `libunwind-<arch>.so`).

        Used as normal symbol for `!REMOTE_ONLY && LOCAL_ONLY` (i.e. `libunwind.so`).

        The use in `ia64/Ginit.c` is the same `get_dyn_info_list_addr` that uses
        `_U_dyn_info_list` on other archs.

So basically the only user that is accessing the wrong `_U_dyn_info_list` is the
`get_dyn_info_list_addr` in `Ginit.c` for other archs and they could just be fixed
to be the same as IA64 instead and use `_U_dyn_info_list_addr()`.

However, looking at the user of `get_dyn_info_list_addr`, it is only used by
`unwi_dyn_remote_find_proc_info` in `mi/Gdyn-remote.c` which is an internal function only used by
`remote_find_proc_info` in `mi/Gfind_dynamic_proc_info.c` but that function
isn't even called for `local_addr_space` (`unwi_find_dynamic_proc_info` in `mi/Gfind_dynamic_proc_info.c`). This is why I said it doesn't really matter what we do since no one should be using the duplicated copy anyway.....

Of course, `unwi_dyn_remote_find_proc_info` is internal but is is a visible symbol
so maybe it does have some external user? If that's the case it'll make sense to fix the usage.
That's why I think the correct fix is to just copy the change and that's what I did in
https://github.com/libunwind/libunwind/compare/master...yuyichao:common

# A few side notes, future works or related questions

* Weak symbol

    The use of weak symbol started in 4e731a55729a25ef0e50f4f2aaa1bd95e4eec89e
    which is used for `!REMOTE_ONLY && !LOCAL_ONLY`.
    However, `!REMOTE_ONLY` indicates that a local version is being built and
    the remote version (`libunwind-<arch>`) is always linked to the local version (`libunwind.so`)
    even back then so I don't see why a weak symbol is needed...

    Maybe there's a flag for building `!REMOTE_ONLY && !LOCAL_ONLY` version without linking
    to `LOCAL_ONLY` version?

* Symbol names

    The comment suggests that IA64 does something that might be able to locate the
    `_U_dyn_info_list` for ptrace and coredump. (or something equivalent)
    Can't make sense of that code ATM so I don't know whether the symbol name of either
    is significant. That's why I prefer not to rename the variable to add an arch prefix
    in case it is actually important and
    in case it makes it harder to implement such thing for other archs later
    (even though it can only break IA64 ATM).

    Also, since this symbol, together with `_U_dyn_info_list_addr` is only defined
    for `LOCAL_ONLY` it is not possible to use it with any other `LOCAL_ONLY` libraries
    or be used by more than one `!REMOTE_ONLY` libraries.
    It is therefore not necessary to add arch prefix to it since there won't be any conflict
    anyway.
    In fact, all symbols in `LOCAL_ONLY` library have this property but we won't be renaming
    them any time soon for backward capatibility I assume. It might also be good to keep the
    prefix to be consistent with the `!LOCAL_ONLY` one.
    (i.e. it's better to have `_Ux86_64foo` and `_ULx86_64foo` rather than `_Ux86_64foo` and
    `foo` or `_Ufoo` even if there's no conflict in either case...)

* Why was `_U_dyn_info_list_addr` needed.

    It seems to be exactly equivalent to `_U_dyn_info_list` and it seems that we should be
    able to just use that instead. Maybe weak symbol support was different 16+ years ago?

    The only benefit of having a function that I've actually used before was to
    set a breakpoint on the use of the variable... That said, many uses doesn't go
    through that function, there are many other ways to set the breakpoint and
    this is hardly the variable that needs this the most...

* Populating `local_addr_space.dyn_info_addr`?

    It should be fine to populate it the same time we populate `local_addr_space.acc.get_dyn_info_list_addr`. It's unused, but so is `local_addr_space.acc.get_dyn_info_list_addr` = = ....
    Probably doesn't worth doing for this one as it probably just add initialization time (by only a few instructions) without fixing any issue.

djwatson:
Thanks for the extensive writeup @yuyichao  ! Patch is merged.

PR #168, opened 2020-04-01
ARM and elf fixes

egtvedt:


yuyichao:
Should't `c->dwarf.use_prev_instr` covers this?

egtvedt:
@yuyichao I suspect you are right, I have had this patch locally for years (since 2014), and finally have gotten around to start upstreaming fixes.

Will test dropping it, and re-submit the coredump-fix in a dedicated pull request.

egtvedt:
@yuyichao AFAIK `use_prev_instr` is only set if you unwind a local process. I unwind a coredump, hence `use_prev_instr` is zero.

In the old days glibc assert function used to end with `bl` function, which was why we initially discovered this need.

Perhaps the noreturn attribute solution needs to be adjusted to current master?

yuyichao:
> AFAIK use_prev_instr is only set if you unwind a local process. I unwind a coredump, hence use_prev_instr is zero.

Then it sounds like that is the problem instead. The current patch is almost certainly wrong for normal unwinding.

I also don't see any unwinding step setting `use_prev_instr` that's local specific. OTOH, though, I also don't see where it is set for exidx unwinding, or any special handling for signal frame, maybe that's the issue?

The only local specific setting of the `use_prev_instr` I could find is in the init function, but that doesn't seem to be the same problem (IIUC `use_prev_instr` is specificly set for each step rather than a property for the entire trace so the init function should only affect the first step). In any case, if that's the true problem, I believe it should be fixed there instead.

djwatson:
please update and reopen, I agree this doesn't look like the correct fix.

PR #169, opened 2020-04-01
coredump: Unmap elf image before mapping another

egtvedt:
Make sure to unmap the currently mapped elf image before mapping
another image in CD_get_proc_name. If not, the reference to that
image is lost and the memory is leaked.

Signed-off-by: Sturle Mastberg <smastber@cisco.com>
Signed-off-by: Hans-Christian Noren Egtvedt <hegtvedt@cisco.com>


djwatson:
lgtm, thanks.

PR #170, opened 2020-04-02
Add AO_REQUIRE_CAS to fix build on ARM < v6

ffontaine:
ARM earlier than ARMv6, such as ARMv4 and ARMv5 do not provide
optimize atomic operations in libatomic_ops. Since libunwind is using
such operations, it should define AO_REQUIRE_CAS before including
<atomic_ops.h> so that libatomic_ops knows it should use emulated
atomic operations instead (even though they are obviously a lot more
expensive).

Also, while real atomic operations are all inline functions and
therefore linking against libatomic_ops was not required, the emulated
atomic operations actually require linking against libatomic_ops, so
the commented AC_CHECK_LIB test in acinclude.m4 is uncommented to make
sure we link against libatomic_ops.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@free-electrons.com>
[Retrieved fom:
https://git.buildroot.net/buildroot/tree/package/libunwind/0001-Add-AO_REQUIRE_CAS-to-fix-build-on-ARM-v6.patch]
Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -22,11 +22,5 @@ fi])
 AC_DEFUN([CHECK_ATOMIC_OPS],
 [dnl Check whether the system has the atomic_ops package installed.
   AC_CHECK_HEADERS(atomic_ops.h)
-#
-# Don't link against libatomic_ops for now.  We don't want libunwind
-# to depend on libatomic_ops.so.  Fortunately, none of the platforms
-# we care about so far need libatomic_ops.a (everything is done via
-# inline macros).
-#
-#  AC_CHECK_LIB(atomic_ops, main)
+  AC_CHECK_LIB(atomic_ops, main)
 ])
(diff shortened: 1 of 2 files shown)


yuyichao:
I always assume this is intentional from the comment that you deleted. It'll be nice to limit this to <armv6 only or make it configurable.

Also, if c11 is available I think that should be used instead...

yuyichao:
Also note that some of the atomic operations seems to be used in places where lock-free might be important (signal handler). Linking to a locking implementation might break that.

(This is also where the c11 `ATOMIC_*_LOCK_FREE` and `atomic_is_lock_free()` come in handy.)

djwatson:
I agree with yuyichao's comments, please update and reopen.

PR #178, opened 2020-05-08
Fix a memory leak issue

am11:
One memleak issue was found by static analyzer.

```sh
$ cppcheck --enable=all . 2>&1 | grep -B5 -A5 -i leak
...
Checking src/tilegx/Gcreate_addr_space.c ...
src/tilegx/Gcreate_addr_space.c:52:5: error: Memory leak: as [memleak]
    return NULL;
    ^
...
```


djwatson:
lgtm, thanks

Issue #124, opened 2019-06-21
Opt out of creating a weak alias for backtrace


pedronavf:
libunwind creates a weak alias for backtrace unconditionally. On our application we write libc's backtrace and our own (previously using the _Unwind functions and now libunwind) so we have two pieces of info and to make sure our implementation doesn't have any bugs.

When calling libunwind functions from a signal handler when an invalid function dereference has happened (null pointer), libunwind crashes in `_ULarm_step`. Because of the alias the crash is happening when we call `backtrace_symbols_fd`, voiding our intent of having the system's backtrace written to disk before calling our own.

In the same way that we can configure without C++ exceptions `--disable-cxx-exceptions`(so the _Unwind functions are not aliased), I think the same should happen with `backtrace`

acmorrow:
I'd expand this to not only making it not declare a weak alias, but to not declare the `backtrace` function at all. I'd like to be able to have a program that links to the system versions of both the libc and libunwind dynamic libraries, and then be able to call both the libc `backtrace` and the libunwind `unw_backtrace` functions. But if `libunwind` exports the `backtrace` symbol at all and ends up earlier in the link order than `libc` (which is likely), then calling `backtrace` will always be intercepted by `libunwind`. This makes it impossible, for instance, to write a meaningful comparative benchmark of libc `backtrace` vs `unw_backtrace`.

pedronavf:
Yes, that's exactly what I was getting at! We want to be able to call both functions from our crash handler.

acmorrow:
@pedronavf - Yes, in full agreement. I am currently in the process of refactoring a crash handler with an eye to moving to libunwind. For the crash handler, performance is rather uninteresting. But we have other uses of libc `backtrace` than just the crash handler, such as for a heap profiler and a heap debugger that captures deleter stacks, and in such cases the performance of the backtrace capture really does matter. We would like to be able to roll out the libunwind based crash handler without affecting the performance of the other `backtrace` callers until we have investigated the performance when they are switched over to libunwind. But we can't do that if libunwind unconditionally intercepts libc `backtrace` when it is linked into the process.

pedronavf:
I ended up patching the library. I should probably work on a pull request.

This is the patch that removes the definition of `backtrace`
```
diff --git a/src/mi/backtrace.c b/src/mi/backtrace.c
index c7aa2bdc..80f6fecf 100644
--- a/src/mi/backtrace.c
+++ b/src/mi/backtrace.c
[...]

PR #185, opened 2020-06-08
Added --disable-unwind-header option

pedronavf:
When set, unwind.h won't be copied to the installation directory. Some platforms provide their own unwind.h that might conflict with the one supplied with libunwind.

As mentioned in #124, libunwind installs its own copy of unwind.h. This file externs `_Unwind_GetIPInfo` which some ARM platforms with an old libstdc++ don't have.

Depending on the order of include paths libunwind's unwind.h might be read before the system one, causing third party code that does not link against libunwind to fail with an `undefined reference to '_Unwind_GetIPInfo'` linker error.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -46,7 +46,10 @@ include_HEADERS += include/libunwind-s390x.h
 endif
 
 if !REMOTE_ONLY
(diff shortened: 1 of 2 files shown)


djwatson:
merged, thanks.

Issue #156, opened 2020-01-28
Building on QNX7


guesshe:
Hi, I am just wondering if there is any instructions on cross-compiling libunwind on qnx7 RTOS?

bregma:
CLosing as duplicate of #246

PR #188, opened 2020-06-11
Portability fixes (from QNX)

bregma:
We have been porting libunwind to the QNX Neutrino operating system (see unrelated #156).  Before we can do that, we have had to make some OS-independent changes to replace some Linuxisms.  In some cases we replaced the use of "symbols reserved for the implementation" with valid libunwind-namespaced symbols because using the reserved symbols is undefined behaviour.

These are really just a collection of minor pain points we would rather push upstream so we can all win.

Builds cleanly on Ubuntu 16.04 x86_64 with GCC 5.4 and GCC 8.3. All tests pass.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -34,8 +34,8 @@ esac
 
 dnl Checks for header files.
 AC_HEADER_STDC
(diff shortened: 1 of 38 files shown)


djwatson:
These all look reasonable to me, thanks!

Issue #190, opened 2020-06-23
build failure and warnings on aarch64 and redhat 7.x


mwkrentel:
Hi, I found a data point where libunwind 1.5-stable doesn't build.
I started looking into this and I have a couple questions or suggestions.
These apply to both master and 1.5-stable.

(1) `--enable-debug-frame` doesn't build on RedHat 7.x because elf.h is
too old.  I first noticed this on aarch64 where debug-frame is turned
on by default, but it applies anywhere with --enable-debug-frame and
RedHat 7.x.

The problem is /usr/include/elf.h is too old for libunwind 1.5 and
doesn't define SHF_COMPRESSED.  The build fails with:

```
   ../../libunwind-1.5-rc1/src/dwarf/Gfind_proc_info-lsb.c: 
   In function 'load_debug_frame':
   ../../libunwind-1.5-rc1/src/dwarf/Gfind_proc_info-lsb.c:130:24: 
   error: 'SHF_COMPRESSED' undeclared (first use in this function)
      if (shdr->sh_flags & SHF_COMPRESSED)
```

This fails on RedHat 7.x, but of course, the real problem is the
version of glibc, 2.17 on RH 7.  But I haven't done the bisection to
identify the rev where this was added.

For my purposes (I've become the maintainer for the spack libunwind
package, github.com/spack/spack), I have an easy workaround.  I just
require a recent elf headers as a prereq.

But I was thinking you might want to add a configure test (see if
SHF_COMPRESSED is defined in elf.h) and turn off debug-frame if not.

Or, maybe pull in a copy of elf.h into the code base.

----------

(2) On aarch64, there is a spew of warnings over the asm definition of
unw_tdep_getcontext().

```
In file included from ../include/libunwind.h:7:0,
                 from mi/backtrace.c:29:
mi/backtrace.c: In function 'unw_backtrace':
../include/libunwind-aarch64.h:236:5: 
warning: right-hand operand of comma expression has no effect [-Wunused-value]
 #define unw_tdep_getcontext(uc) (({     \
                                 ~~~~~~~~~
   unw_tdep_context_t *unw_ctx = (uc);     \
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   register uint64_t *unw_base __asm__ ("x0") = (uint64_t*) unw_ctx->uc_mcontext.regs;  \
[...]

PR #191, opened 2020-06-23
Make SHF_COMPRESSED use contingent on its existence

bregma:
A recent change broke builds for many non-recent and non-Linux targets (like QNX)
by adding code dependent on a non-portable preprocessor macro SHF_COMPRESSED
found in some more recent GLIBC elf.h headers.

This fixes the first part of #190.

Build and verified on x86_64 Ubuntu 16.04, All tests pass.

### DIFF
--- a/src/dwarf/Gfind_proc_info-lsb.c
+++ b/src/dwarf/Gfind_proc_info-lsb.c
@@ -127,6 +127,7 @@ load_debug_frame (const char *file, char **buf, size_t *bufsize, int is_local)
       return 1;
     }
 
+#if defined(SHF_COMPRESSED)
   if (shdr->sh_flags & SHF_COMPRESSED)
     {
       unsigned long destSize;
@@ -161,6 +162,7 @@ load_debug_frame (const char *file, char **buf, size_t *bufsize, int is_local)
     }
   else
     {
+#endif
       *bufsize = shdr->sh_size;
       GET_MEMORY(*buf, *bufsize);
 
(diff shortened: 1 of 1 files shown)


mwkrentel:
Looks good to me.  With this patch, 1.5-stable now builds on the
original aarch64 RedHat 7.x machine where I first noted the problem.

Be sure to add to both 1.5 and master.

Btw, in your patch, if you move the first `#endif` before the opening
`{` for else, I think you can eliminate the second `#if defined()`.

```
 163   else
 164 #endif
 165     {
 166       *bufsize = shdr->sh_size;
 167       GET_MEMORY(*buf, *bufsize);
           ...
 173     }
```

But that's just a style thing.

Thanks !!

yuyichao:
Is it just a platform independent constant? If it is I think it’s better to just define it and I’ll help when building “portable” binaries.

mwkrentel:
It's really a glibc version issue.  You could define `SHF_COMPRESSED`,
but then you'd find that `struct Elf64_Chdr` was undefined, etc.

If you really wanted to do that, that is support this feature on older
systems (and RH 7 isn't all that old), then you should pull in a copy
of `elf.h` from a recent glibc into the code base.

But that's a much bigger decision, especially right before 1.5 release.
It's worth thinking about, but I'm not trying to push you one way or
the other.

yuyichao:
OK if there are more than constants that needs to be conditionally defined then this should be OK for now.

mwkrentel:
Is there a time frame for merging?

bregma:
We just have to wait for someone with admin privs like @djwatson to review and commit.

pedronavf:
Oh, I didn't see this and submitted PR #194 which, taking Android as an example, defines the flag.

djwatson:
Merged to master and v1.5-stable.  Thanks!

PR #194, opened 2020-07-02
Define the SHF_COMPRESSED flags if not provided by the system

pedronavf:
Commit a36ec8cfd added support for compressed debug sections, but some system with an older elf.h or libc don't provide the SHF_COMPRESSED flag and libunwind fails to build.

Taking Android as an example (https://android.googlesource.com/platform/external/elfutils/+/android-7.0.0_r1/libelf/libelf.h) if SHF_COMPRESSED is not defined we now define it so compressed sections work.

Another approach would have been to introduce a configure time check to see if SHF_COMPRESSED is supported and disable compressed section support completely if not.

### DIFF
--- a/src/dwarf/Gfind_proc_info-lsb.c
+++ b/src/dwarf/Gfind_proc_info-lsb.c
@@ -106,6 +106,36 @@ linear_search (unw_addr_space_t as, unw_word_t ip,
    contains no .debug_frame.  */
 /* XXX: Could use mmap; but elf_map_image keeps tons mapped in.  */
 
+#ifndef SHF_COMPRESSED
+ /* Older glibc elf.h might not yet define the ELF compression types.  */
+#define SHF_COMPRESSED      (1 << 11)  /* Section with compressed data. */
+
+ /* Section compression header.  Used when SHF_COMPRESSED is set.  */
+#if ELF_CLASS == ELFCLASS32
+ typedef struct
+ {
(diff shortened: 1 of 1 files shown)


src/dwarf/Gfind_proc_info-lsb.c
```diff
@@ -106,6 +106,34 @@ linear_search (unw_addr_space_t as, unw_word_t ip,
...
+   Elf32_Word   ch_addralign;   /* Uncompressed data alignment.  */
+ } Elf32_Chdr;
+
+ typedef struct
+ {
+   Elf64_Word   ch_type;        /* Compression format.  */
+   Elf64_Word   ch_reserved;
+   Elf64_Xword  ch_size;        /* Uncompressed data size.  */
+   Elf64_Xword  ch_addralign;   /* Uncompressed data alignment.  */
+ } Elf64_Chdr;
```
pedronavf, line 127: Elf64 might not be defined in all platforms. This needs an #ifdef

pedronavf: Fixed.

djwatson:
I merged #191 instead.  Reopen if there is some reason there some advantage to this version instead

pedronavf:
Fair enough. The constant is platform independent, so with my change you could read a compressed section even if your system's libc is old, as long as you have the required libraries (ie, zlib).

PR #195, opened 2020-07-09
linux: Use /proc/[pid]/root

mdurnev:
When we are backtracing processes running in a container, we cannot rely on ELF paths
in /proc/[pid]/maps. Those paths are valid inside the container. We have to add
/proc/[pid]/root to each path to access the ELF files from outside of the container.

Signed-off-by: Mikhail Durnev <mikhail_durnev@mentor.com>


mdurnev:
Could anyone take a look into this patch please?

djwatson:
lgtm, thank you!

mdurnev:
@djwatson, thank you for merging it!

Issue #129, opened 2019-08-04
Is tests/Ltest-cxx-exceptions hiding a problem at aarch64 by not linking to libunwind?


bernhardu:
I was trying to collect some information about https://bugs.debian.org/932499
that seems to trigger just at aarch64 and seems to be caused by libunwind somehow.

That led me to the test Ltest-cxx-exceptions which seems not to link against libunwind.
Isn't it that way just testing gcc's implementation?

I found that by LD_PRELOAD it can be loaded.
At x86_64 both implementations seem to succeed the test.
But at aarch64 libunwind seems to fail.

Should Ltest-cxx-exceptions link against libunwind.so?

Could you reproduce the issue?

Failing example when using libunwind.so:
```
benutzer@debian:~/source/libunwind8/git/libunwind_v1.4-rc1-13-gb316234d_try22$ LD_PRELOAD=src/.libs/libunwind.so.8.0.1:/lib/aarch64-linux-gnu/libgcc_s.so.1 gdb -q -ex 'b main' -ex run -ex 'info share' -ex 'b _Unwind_RaiseException' -ex 'cont' -ex 'bt' -ex 'disa 2' -ex cont --args tests/Ltest-cxx-exceptions 
Reading symbols from tests/Ltest-cxx-exceptions...done.
Breakpoint 1 at 0xa00: file Ltest-cxx-exceptions.cxx, line 65.
Starting program: /home/benutzer/source/libunwind8/git/libunwind_v1.4-rc1-13-gb316234d_try22/tests/Ltest-cxx-exceptions 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/aarch64-linux-gnu/libthread_db.so.1".

Breakpoint 1, main (argc=1, argv=0xfffffffff418) at Ltest-cxx-exceptions.cxx:65
65        if (argc > 1)
From                To                  Syms Read   Shared Object Library
0x0000ffffbf6d0040  0x0000ffffbf6e6c48  Yes         /lib/ld-linux-aarch64.so.1
0x0000ffffbf691770  0x0000ffffbf699dcc  Yes         src/.libs/libunwind.so.8.0.1
0x0000ffffbf66d910  0x0000ffffbf67b568  Yes         /lib/aarch64-linux-gnu/libgcc_s.so.1
0x0000ffffbf55fdd0  0x0000ffffbf603980  Yes         /lib/aarch64-linux-gnu/libstdc++.so.6
[...]

PR #196, opened 2020-07-13
Link Ltest-cxx-exceptions with libunwind

AdrianBunk:
Bernhard Übelacker noticed that Ltest-cxx-exceptions was actually
testing the gcc implementation due to not linking with libunwind.

Fixes: #129


djwatson:
wow, nice catch!

PR #200, opened 2020-07-28
ia64: Fix dyn_info_list.S with newer binutils

jrtc27:
Using binutils 2.34 or 2.35, libunwind fails to link on ia64 with:

  /usr/bin/ld: .IA_64.unwind has both ordered [`.IA_64.unwind' in unwind/.libs/GetIPInfo.o] and unordered [`.IA_64.unwind' in ia64/.libs/dyn_info_list.o] sections

Fix this by using the right flags and type for the section, taken from binutils's gas/testsuite/gas/ia64/unwind.s.


djwatson:
Merged, thanks!

PR #201, opened 2020-09-02
Use __linux__ define everywhere instead of __linux

milianw:
The libunwind code base uses __linux__ already in many locations, but
some others relied on __linux instead. Apparently on some older
toolchain configurations, such as Ubuntu Trusty that's still used for
AppImage generation sometimes, __linux is never defined - only
__linux__ is. Fixes compiler warning on such platforms:

```
coredump/_UPT_get_dyn_info_list_addr.c: In function 'get_list_addr':
coredump/_UPT_get_dyn_info_list_addr.c:86:3: warning: #warning Implement get_list_addr(), please. [-Wcpp]
 # warning Implement get_list_addr(), please.
   ^
```

### DIFF
--- a/src/coredump/_UCD_find_proc_info.c
+++ b/src/coredump/_UCD_find_proc_info.c
@@ -35,7 +35,7 @@ get_unwind_info(struct UCD_info *ui, unw_addr_space_t as, unw_word_t ip)
 {
   unsigned long segbase, mapoff;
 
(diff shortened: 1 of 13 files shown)


milianw:
See also: https://stackoverflow.com/questions/6649936/c-compiling-on-windows-and-linux-ifdef-switch

djwatson:
Merged.  Thanks milianw!

PR #214, opened 2020-10-02
Define the FP register size on PPC to be double

### LINKED ISSUE #208: FPSCR register overwritten by random bytes on PPC64LE
I was debugging a failure on PPC which occurred only when using libunwind and only after an exception is thrown. I traced this back to the FPSCR register being changed as observed by floating point exceptions seemingly randomly enabled after unwind.

More details in my SO question: https://stackoverflow.com/questions/63959740/throwing-a-c-exception-with-libunwind-on-powerpc-loaded-sets-random-floating-p and below

Debugging this through GDB led me to `setcontext` where ` mtfsf  0xff,fp0,1,0` is executed which changes the FPSCR. `fp0` in turn is loaded by `lfd  fp0,(SIGCONTEXT_FP_REGS+(32*8))(r31)`

There are relevant comments at https://github.com/libunwind/libunwind/blob/c63c2369a5e9eefb6e5fbb240edaecae8856a02e/include/libunwind-ppc64.h#L151-L156 and https://github.com/libunwind/libunwind/blob/c63c2369a5e9eefb6e5fbb240edaecae8856a02e/src/ppc64/Gstep.c#L311-L312 and a define at https://github.com/libunwind/libunwind/blob/c63c2369a5e9eefb6e5fbb240edaecae8856a02e/src/ppc64/ucontext_i.h#L134

Checking this loop: https://github.com/libunwind/libunwind/blob/c63c2369a5e9eefb6e5fbb240edaecae8856a02e/src/ppc64/Gresume.c#L79 it seems that the last element of `fpregs` is indeed not written and hence contains possibly indeterminate values which is then accessed by glibc to retstore the FPSCR

Flamefire:
According to GLIBC headers those are in fact double (4 Byte) not long double (8 Byte)

Fixes #208 

Note that there are still problems: The vector registers are also accessed via `unw_tdep_fpreg_t` which was likely not correct before and will now read/write even less.
I'd also strongly suggest to add compile time checks that the defines for such types that are "blindly" casted to memory locations in existing structs are actually correct. I'm not sure if `static_assert` exists in C but something like this could be used to check for some types or at least same sizes wherever such a cast from one pointer type to another is made.

### DIFF
--- a/include/libunwind-ppc64.h
+++ b/include/libunwind-ppc64.h
@@ -72,7 +72,7 @@ typedef uint64_t unw_word_t;
 typedef int64_t unw_sword_t;
 #endif
 
-typedef long double unw_tdep_fpreg_t;
+typedef double unw_tdep_fpreg_t;
 
 /*
  * Vector register (in PowerPC64 used for AltiVec registers)


djwatson:
you could do something like

assert(sizeof(unw_tdep_fpreg_t) == sizeof(ucontext_t.reg)) 

or something, that's the closest C has.  

Thanks for the patch.

yuyichao:
> I'm not sure if static_assert exists in C

It does exist in c11.

PR #220, opened 2020-12-02
Fix .debug_frame load_offset calculation

oleavr:
Typically resulting in a crash during local unwinding on e.g.
Android/arm64, where the calculated load_offset would be zero.


djwatson:
looks good to me. Thanks!

PR #222, opened 2020-12-16
Gtest-exc depends on frames not being inlined

vtjnash:
The test is unwinding a precise number of frames, which assumes the
compiler will not inline these functions. Mark this explicitly to be
more reliable, and add some more verbose debugging aids.


djwatson:
Thanks!  Yea the unittests are all fairly fragile, any compiler hints here are welcome.  I've mostly been keeping x86_64 ones running, but even then they only all pass on gcc, not even clang.

PR #223, opened 2020-12-16
RFC: improve unw_getcontext on ARM/Thumb

vtjnash:
Better support unw_resume, by returning r0 from the user, instead of
hard-coding it to zero (also incidentally fixing a compiler warning
about an unused expression). This better mimics a real function call,
instead of hard-coding the return value. When in thumb mode, this also
means we also need to avoid switching between thumb and ARM upon return
to this point, so we need to set the pc accurately to point after the
return instruction. When in ARM mode, we need an extra nop also to point
at the correct return instruction (since the stored value of pc in the
context runs ahead by 8 bytes, while stmia is only 4 bytes).

### DIFF
--- a/include/libunwind-arm.h
+++ b/include/libunwind-arm.h
@@ -268,28 +268,42 @@ typedef struct unw_tdep_context
   }
 unw_tdep_context_t;
 
-/* There is no getcontext() on ARM.  Use a stub version which only saves GP
-   registers.  FIXME: Not ideal, may not be sufficient for all libunwind
-   use cases.  Stores pc+8, which is only approximately correct, really.  */
+/* FIXME: this is a stub version which only saves GP registers.  Not ideal, but
+   may be sufficient for all libunwind use cases.
+   In thumb mode, we return directly back to thumb mode on return (with bx), to
+   avoid altering any registers after unw_resume. */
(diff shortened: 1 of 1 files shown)


djwatson:
lgtm, but I'm not an arm expert.  Happy to merge if there are no other comments.

vtjnash:
As a side note on the design here: it's unclear to me whether the many platforms here should be modified to store the values (r0, x0, rax, eax, etc.) at this function's return, or preserve their value on entry as they typically do at present. Presently, unw_getcontext/unw_resume aren't quite the same as setjmp/longjmp in this detail, as it means the return result first will be 0, then on `unw_resume`, the second return value from `unw_getcontext` will be whatever meaningless garbage the compiler had left in the return register when we'd first entered this function.

For this code, it'd be the difference between:
```
     "mov r0, #0\n"
     "stmia %[base], {r0-r15}\n"
     "nop\n" /* align return address to value stored by stmia */
```
or 
```
     "stmia %[base], {r0-r15}\n"
     /* the next instruction also aligns the return address to the value stored by stmia */
     "mov r0, #0\n"
```

djwatson:
Merged, thanks!

PR #224, opened 2020-12-16
stop lying about the ARM PC location?

vtjnash:
I am not sure why it should lie about the PC register and use a
different random value (LR), when the actual value is also easily
available. The rate of passing tests seems to be the same.
(I think it may be aliasing the location of LR to IP internally
elsewhere, in unw_step, which ends up hiding the lie—mostly.)


djwatson:
lgtm, but I'm not an arm expert.  Happy to merge if there are no other comments.

djwatson:
Merged, thanks

PR #225, opened 2020-12-16
RFC: mark unw_getcontext functions as returns_twice

vtjnash:
This instructs the compiler to avoid optimizations in the callee that
would potentially interact poorly with interpreting the state, such as
frame-pointer-elimination on certain platforms, alloca lifetime
analysis/slot reuse, and so on.

It also passes along the return value from getcontext, when used (which
fortuitously is already UNW_ESUCCESS or -UNW_EUNSPEC), instead of
forcing it to be 0.

### DIFF
--- a/include/libunwind-hppa.h
+++ b/include/libunwind-hppa.h
@@ -116,7 +116,7 @@ unw_tdep_proc_info_t;
 #include "libunwind-common.h"
 
 #define unw_tdep_getcontext             UNW_ARCH_OBJ (getcontext)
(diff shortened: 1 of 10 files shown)


djwatson:
The MSVC doesn't seem to like this, the contbuild fails:

https://travis-ci.org/github/libunwind/libunwind/jobs/750077969

djwatson:
Closing.  Feel free to update so MSVC doesn't break, if you'd like.

PR #226, opened 2020-12-16
fix build on ARM?

vtjnash:
This file can get included both directly and through Lex_tables, resulting in a symbol conflict when this function then gets defined twice (with simple `./autogen.sh && make`). Marking it `static` (local to each file) lets the build succeed, though I don't know if that's the preferred strategy (vs. some combination of `#ifdef` with `UNW_LOCAL_ONLY` and `UNW_REMOTE_ONLY`).

### DIFF
--- a/src/arm/Gex_tables.c
+++ b/src/arm/Gex_tables.c
@@ -381,7 +381,7 @@ arm_exidx_extract (struct dwarf_cursor *c, uint8_t *buf)
   return nbuf;
 }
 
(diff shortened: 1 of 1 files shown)


djwatson:
If it isn't an exported interface function, then yes I believe it has to be marked static

vtjnash:
bump

djwatson:
Merged, thanks.

PR #227, opened 2020-12-16
RFC: disallow tail-call optimization in tests

vtjnash:
This disallows frame reuse for tail and other sibling calls. I'm not aware of any place where a sibling call optimization would actually be used by the compiler (the failing test I was looking at turned out to be due to inlining, not a sibling call), but I though this might help ensure future compatibility for counting the expected number of stack / unwind frames. Feel free to request this be done differently, or just close it, if it doesn't sound useful enough now.


djwatson:
Seems fine to me.

vtjnash:
bump

djwatson:
Merged, thanks!

PR #231, opened 2021-02-01
Fix Clang builds

ffainelli:
While cross-compiling strace against libunwind for ARM, the configure
script of strace failed with:

libunwind-ptrace.h:40:27: error: a parameter list without types is only
allowed in a function definition

it turns out that we do not have a definition for what pid_t should be,
so include sys/types.h to remedy that.


ffainelli:
@djwatson does that sound reasonable to you?

djwatson:
Merged, thanks!

ffainelli:
Cool, thanks!

Issue #233, opened 2021-02-10
x86_64 `unw_cursor_t` is self-referential and thus not copiable


EliaGeretto:
According to the `man` page for `libunwind`, cursors should be copiable and, apparently, valid as long as the `unw_context_t` they have been created with exists:
> ... this effect can be achieved by making copies of an unwind  cursor.

However, this is not true for the x86_64 `unw_cursor_t`. The following line makes it a self-referential struct:
https://github.com/libunwind/libunwind/blob/master/src/x86_64/Ginit_local.c#L52

For this reason, when copied, the pointer in `dwarf.as_arg` may become invalid. The problem seems to have been introduced in commit 649f1fb3449a65dd0626a709432d8b02a7c56bbc.

EliaGeretto:
The same self-reference is also present for: `s390x`, `x86`.

EliaGeretto:
The issue was solved for `x86_64` in commit https://github.com/libunwind/libunwind/commit/debb6128d17b782552d53efa8869a392d1f40a83, I have updated the title to show that the issue is not fixed for `x86` and `s390x`. From what I have seen, it should be relatively easy to port my patch to `x86_64` in https://github.com/libunwind/libunwind/pull/234 to those platforms as well, if someone wants to pick it up.

PR #234, opened 2021-02-11
[x86_64] Remove self-reference from local cursor

EliaGeretto:
When creating a local x86_64 cursor, its `dwarf.as_arg` member is set to
point to the cursor itself, making it a self-referential struct. This
does not allow to safely keep a copy of a cursor, as stated in the
documentation. In addition, the self-reference is used to access just
two members: `uc` and `validate`.

This commit modifies `dwarf.as_arg` to pack together both `uc` and
`validate`, so the x86_64 cursor is not self-referential anymore and no
additional memory allocation is performed. Since `uc` points to a
`ucontext_t`, which is at least 2-byte aligned, it is safe to store the
`validate` bit in the LSB of the `uc` pointer.

Additional checks were added to verify that the `validate` bit is not
set when the cursor is non-local.

This PR fixes https://github.com/libunwind/libunwind/issues/233.

### DIFF
--- a/include/tdep-x86_64/libunwind_i.h
+++ b/include/tdep-x86_64/libunwind_i.h
@@ -31,6 +31,7 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 /* Target-dependent definitions that are internal to libunwind but need
    to be shared with target-independent code.  */
 
+#include <stdint.h>
 #include <stdlib.h>
 #include <stdatomic.h>
 #include <libunwind.h>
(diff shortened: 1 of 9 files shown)


EliaGeretto:
If possible, I would like to have some comments from the maintainers on this solution. Packing `validate` in the `uc` pointer may not be the cleanest solution, but the alternative was to perform a heap allocation or to use a thread-local variable, which are both slower.

djwatson:
I don't have any particular comments, this looks good to me.

EliaGeretto:
@djwatson Thank you for taking a look. I still want to check if I can move the check for `unw_local_addr_space` inside `ACCESS_MEM_FAST` in  `src/x86_64/Gtrace.c`. After that, I will submit it for review. This fixes only `x86_64`, but it should be easily portable to `s390x` and `x86`, which are affected by the same issue.

EliaGeretto:
Some of the checks in `Gtrace.c` could be removed when compiling with `UNW_LOCAL_ONLY`. In particular, the checks for `d->as == unw_local_addr_space` are not necessary, since that equality is always true. These checks could be moved inside `ACCESS_MEM_FAST` which, however, requires a `validate` parameter that would not be available. As a consequence, removing those checks would require modifying `ACCESS_MEM_FAST` which is used in other architectures as well and is thus out of the scope of this PR. I am thus submitting the code for review as is.

EliaGeretto:
Ping.

djwatson:
Merged, thanks!

PR #230, opened 2021-01-28
Fix elf get_load_offset when the binary was linked by lld, which defi…

Romain-Geissler-1A:
…nes read/executable program header offsets not always aligned on a page size.

For example, the same library, linked with ld.bfd will show these PT_LOAD mappings:
  LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x000000000002b3d0 0x000000000002b3d0  R      0x1000
  LOAD           0x000000000002c000 0x000000000002c000 0x000000000002c000
                 0x000000000001b729 0x000000000001b729  R E    0x1000
  LOAD           0x0000000000048000 0x0000000000048000 0x0000000000048000
                 0x000000000000fb17 0x000000000000fb17  R      0x1000
  LOAD           0x0000000000057fb0 0x0000000000058fb0 0x0000000000058fb0
                 0x00000000000010b0 0x0000000000001508  RW     0x1000

while ld.lld, from the very same .o files, will link it like this:
  LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x000000000003b32c 0x000000000003b32c  R      0x1000
  LOAD           0x000000000003b330 0x000000000003c330 0x000000000003c330
                 0x000000000001b740 0x000000000001b740  R E    0x1000
  LOAD           0x0000000000056a70 0x0000000000058a70 0x0000000000058a70
                 0x0000000000001050 0x0000000000001050  RW     0x1000
  LOAD           0x0000000000057ac0 0x000000000005aac0 0x000000000005aac0
                 0x0000000000000060 0x00000000000004b3  RW     0x1000

Failure to identify the load_offset from iterating the program header lead to later being unable
to find the correct proc_info for a given symbol.

@maskray can you please confirm it's expected that ld.lld generates libraries.so where the RE PT_LOAD mapping is having a p_offset which is not page aligned (my example is taken from an x64 library).

PR #238, opened 2021-02-17
Fix elf get_load_offset when the binary was linked by lld: return the…

Romain-Geissler-1A:
… right offet.

Followup pull request, first part was in PR #230


Romain-Geissler-1A:
And it looks like I forgot the trailing semicolon here... sorry for that, I had it when I tested...

Issue #216, opened 2020-10-14
Link issues with ARM target 1.4.0, 1.5rc2


crzysdrs:
Building from a linux x86_64 host targeting arm has missing symbols for functions relating to 32bit elf reading.

```
$  gcc -v 2>&1 | grep -i ^Target
Target: x86_64-linux-gnu
```

```sh
PREFIX=/tmp/libunwind/
TARGET_SUFFIX=linux-gnu
HOST=x86_64-${TARGET_SUFFIX}
 
#Generic Build
 ./configure --build=${HOST} \
             --enable-coredump --enable-ptrace \
                --prefix=${PREFIX} 
make -j
make install
make check

for TARGET in \
    arm-${TARGET_SUFFIX} \
    x86_64-${TARGET_SUFFIX} \
    x86-${TARGET_SUFFIX}    
do
    ./configure --build=${HOST} \
                --target=${TARGET} \
                --enable-coredump --enable-ptrace \
                --prefix=${PREFIX} 
    make -j
    make install
    make check
done
```

During make check:
```
libtool: link: gcc -g -O2 -fexceptions -Wall -Wsign-compare -o .libs/test-proc-info test-proc-info.o  ../src/.libs/libunwind-arm.so -Wl,-rpath -Wl,/tmp/libunwind/lib
/usr/bin/ld: ../src/.libs/libunwind-arm.so: undefined reference to `_Uelf32_load_debuglink'
/usr/bin/ld: ../src/.libs/libunwind-arm.so: undefined reference to `_Uelf32_find_section'
```
```

$ nm /tmp/lib/libunwind-arm.so | grep Uelf
                 U _Uelf32_find_section
                 U _Uelf32_load_debuglink

```
```
$ nm /tmp/lib/libunwind-generic.so | grep Uelf
000000000000fb60 t _Uelf64_find_section
00000000000103c0 t _Uelf64_get_proc_name
000000000000fc70 t _Uelf64_get_proc_name_in_image
000000000000ff50 t _Uelf64_load_debuglink
000000000000f990 t _Uelf64_lookup_symbol.isra.0
```

```
$ nm /tmp/lib/libunwind.so | grep Uelf
000000000000a010 t _Uelf64_find_section
000000000000a870 t _Uelf64_get_proc_name
000000000000a120 t _Uelf64_get_proc_name_in_image
000000000000a400 t _Uelf64_load_debuglink

```
[...]

Issue #237, opened 2021-02-17
Symbols missing when using ./configure --target=host and when --enable_debug_frame is set


aletru01:
nm shows that those two symbols are undefined, it seems to happen when `./configure --target=host` is used and when `--enable_debug_frame` is set.
```
nm src/.libs/libunwind-aarch64.so.8.0.1
...
                 U _Uelf64_find_section
                 U _Uelf64_load_debuglink
...
```

This issue comes from the macro which is not set with `--target=host`
```
#ifndef UNW_REMOTE_ONLY
# include "elf64.h"
# include "elfxx.c"
#endif
```
The question is, is UNW_REMOTE_ONLY supposed to be set in --target=host mode? If that's the case maybe there should be a dependency on target==host for --enable-debug-frame. Or maybe the macro shouldn't be set?

knjmooney:
UNW_REMOTE_ONLY is only ever set when target_arch != host_arch. If --enable-debug-frame doesn't work remotely, then you could disable it from building with the following in configure.ac
```
case "${target_arch}" in
  (arm) enable_debug_frame=yes;;
  (aarch64) enable_debug_frame=yes;;
  (*)   enable_debug_frame=no;;
esac])
+ if test xremote_only == xyes
+  enable_debug_frame=no;;
+ fi
```

Boucher-Lambert:
Hi, this has been solved by https://github.com/libunwind/libunwind/pull/240/commits/d60f8b198e147d8ce67d8db65f3ad3b8897f9519
Can it be closed ?

PR #240, opened 2021-02-18
Update configure.ac to prevent missing symbols

aletru01:
nm shows that those two symbols are undefined, it happens when
./configure --target=host is used and when --enable_debug_frame is set.

nm src/.libs/libunwind-aarch64.so.8.0.1
...
                 U _Uelf64_find_section
                 U _Uelf64_load_debuglink
...

This change will prevent build issues until there is remote debug_frame
support on arm.


aletru01:
Fixes #216, #237

djwatson:
Merged, thanks!

PR #245, opened 2021-03-12
Fixes for the s390x target

uweigand:
- Fix unwinding GPRs that were saved in FPRs
  Since DWARF_FPREG_LOC locations satisfy *both* DWARF_IS_REG_LOC
  and DWARF_IS_FP_LOC, code that has to distinguish between the
  two must check DWARF_IS_FP_LOC *first*.
  This fixes a failure in test-ptrace on Ubuntu 20.04.

- Fix build using (older versions of) clang
  Use fully-qualified register names (%r15) in inline asm.
  Those work with all existing compilers.

- Add missing s390x include in include/tdep/dwarf-config.h

- Add s390x files to CMakeLists.txt to enable cmake build


djwatson:
Merged, thanks!

uweigand:
Thank you!

Issue #127, opened 2019-07-10
Should README declare support of PPC32?


ChenQi1989:
There are ppc32 codes in repo. Should README declare to support ppc32?

djwatson:
Sure, send a pull request.

PR #253, opened 2021-05-24
README: document ppc32 support

ovpanait:
PPC32 support was confirmed in #127, but this info was not added to the README.

To make things clear, add ppc32 to the supported arch/os table.


djwatson:
Merged, thanks!

PR #257, opened 2021-06-16
Declare that only DWARF CIE version 3 and below are supported

### LINKED ISSUE #256: Wrong parsing of DWARF Common Information Entry version 4 and 5
The function [`parse_cie()`](https://github.com/libunwind/libunwind/blob/19a2c01b1e8ac20871ea09d20f596d425ba53aed/src/dwarf/Gfde.c#L46) assumes that the DWARF Common Information Entry (CIE) structure being parsed follows format version 3 or below, because the CIE format version 4 introduces two new fields right after the `augmentation` (UTF-8 string) field:
- `address_size` (ubyte).
- `segment_size` (ubyte).

See section **6.4.1 Structure of Call Frame Information** of the [DWARF 4 specifications](http://www.dwarfstd.org/doc/DWARF4.pdf).

The CIE version check in [`parse_cie()`](https://github.com/libunwind/libunwind/blob/19a2c01b1e8ac20871ea09d20f596d425ba53aed/src/dwarf/Gfde.c#L120) is wrong because `DWARF_CIE_VERSION_MAX` is defined as `4` in [dwarf.h](https://github.com/libunwind/libunwind/blob/19a2c01b1e8ac20871ea09d20f596d425ba53aed/include/dwarf.h#L143).

This problem causes the parsing of a CIE version 4 structure to confuse the fields `code_alignment_factor (unsigned LEB128)` and `data_alignment_factor (signed LEB128)` for the fields `address_size` (ubyte) and `segment_size` (ubyte). This in turn results in garbage behavior when interpreting the instructions of the CIE.

The correct value of `DWARF_CIE_VERSION_MAX` should be `3`, in order to report an error when parsing CIE versions higher than 3, due to the additional fields.

koutheir:
DWARF version 4 introduces new fields in the Common Information Entry (CIE) structure. That causes wrong parsing of that version of the structure. This fix causes the library to report an error instead of silently and wrongly parsing the structure.

Closes #256.

### DIFF
--- a/include/dwarf.h
+++ b/include/dwarf.h
@@ -140,7 +140,7 @@ typedef enum
 dwarf_expr_op_t;
 
 #define DWARF_CIE_VERSION       3
-#define DWARF_CIE_VERSION_MAX   4
+#define DWARF_CIE_VERSION_MAX   3
 
 #define DWARF_CFA_OPCODE_MASK   0xc0
 #define DWARF_CFA_OPERAND_MASK  0x3f


kasperk81:
added in #56. 
@yuyichao fyi

yuyichao:
I believe #56 was because there was a version 4 frame that can't be handled without. Is it impossible to fix the parsing instead without causing previously working parsing to regress?

koutheir:
Fixing parsing to support CIE version 4 is a worthwhile feature, but it is still a feature. This PR fixes a parsing bug.

yuyichao:
But it also introduce a regression for cases that was working before. I don't have an example to show right now but I suspect due to the size being specified at the beginning it was working fine for some cases right now.

kasperk81:
i am not sure if it is a bug though. we are guarded by `sized_augmentation` condition for the stuff we don't understand. we just ignore the extra u8 address_size and segment_size. no?

yuyichao:
I think the point is that the extra field was added not to the end (not sure why). But I swear there was cases (I think generated by LLVM) that appears to work previously. I suspect it's because the values happens to match/makes sense in certain way and then we ignored things afterwards.

koutheir:
This worked before because the tests were not exhaustive enough. Yes, a structure full of zeros will probably not detect this bug, and yet it's still there, and I proved it.

This will not introduce regressions, it will reveal false tests successes.

The prefix size only helps with the parsing of the following structures, but parsing one structure is still pretty much wrong.

The new fields are not part of augmentation. They are after that. The code doesn't ignore them bit rather confuses them for other fields. I went through that code debugging it many times before I figured out the source of the wrong values that didn't correspond with the output of readelf and objdump.

yuyichao:
And shouldn't the correct fix simply be

```c++
if (version > 3)
    addr += 2;
```

?

yuyichao:
> This will not introduce regressions, it will reveal false tests successes.

What I am saying has nothing to do with tests, rather actual users, and the previous code does appear to work with some particular patterns of CIE.

koutheir:
In my particular use case, the wrong parsing caused interpretation of out-of-the-blue instructions for the parsed CIE version 4. And that caused the unwind process to halt, and caused expected landing pads to be missed, and that caused leaks and then a crash.

yuyichao:
I'm not saying that the current code works with everything. And I'm not saying that there's nothing to fix. I'm just saying that it can be fixed in a much better way without affecting any code that happens to work right now. Would the version check I posted above work for your case? Or do you need a check of the value of the two fields as well, sth like

```c++
if (version > 3) {
    if (dwarf_readu8(...) != 0 || dwarf_readu8(...) != 0 || address_size != ... || segment_size != ...)
        // error
}
```

?

yuyichao:
For actually using the values read from here I totally agree it's a new feature that should be implemented separately. However, since the immediate issue is simply because of some extra info, I feel like we should be able to ignore those with a version/value check.

koutheir:
I'm not an expert of DWARF version 4, and how exactly it differs from version 3. I simply discovered that `libunwind` does not handle DWARF version 4 because of, **at least**, this bug. I don't know if there are other bugs like this. What I know for sure is that DWARF version 4 is not supported.

Leaving `DWARF_CIE_VERSION_MAX` as `4` and skipping these two new fields is just another shot in the dark. Setting `DWARF_CIE_VERSION_MAX` as `4` should be done after an investigation and some tests, and that is, again, a worthwhile feature.

yuyichao:
If there weren’t cases that actually worked I would certainly agree with that.

koutheir:
Saying that some cases worked does not allow one to assert that DWARF version 4 is supported.

yuyichao:
But having partial support is also how to start. So not error in some cases is not claiming full support. Whereas error in cases that previously working is definitely regression.

koutheir:
I agree, but does this imply that setting `DWARF_CIE_VERSION_MAX` to `4` means "DWARF CIE version 4 is partially supported"?

If that is the case, then all code using CIE must check its version and report errors when unsupported aspects of CIE version 4 are encountered. In order to do this, a comprehensive comparison of CIE version 3 and version 4 needs to be performed, not only in terms of data format, but also in semantic changes.

By the way, partially supporting a standard is very confusing and error-prone.

djwatson:
http://dwarfstd.org/doc/DWARF5.pdf

Section 1.5 says only two Dwarf 3->4 changes are incompatible, this being one of them.  The other doesn't look like it will affect unwind, only debug.  I would lean towards rolling forwards with something like @yuyichao 's proposed change

For Dwarf 4->5 (section 1.4), it looks like we would have to support the new DW_OP_addrx (and all the other 'x') operations, but I think that's it.

koutheir:
Thank you for the analysis.

If you're certain that *"The other doesn't look like it will affect unwind, only debug."*, then I agree that keeping `DWARF_CIE_VERSION_MAX` set to `4` makes more sense after fixing of the CIE parser. For CIE version 4, I would opt to parse the two additional fields and minimally validate their values, even if the information provided by those fields is not actually used elsewhere.

If you're uncertain about the statement above, then I would opt to set `DWARF_CIE_VERSION_MAX` to `3`.

djwatson:
Does anyone happen to have any test cases?  I attempted a fix in https://github.com/libunwind/libunwind/pull/266

djwatson:
Closing, please add new discussion to #266

PR #262, opened 2021-06-22
Add missing case for IA64 in coredump

kasperk81:
https://bugs.gentoo.org/586092

this might not be correct, but at least it builds, and doesn't crash.

### DIFF
--- a/src/coredump/_UCD_access_reg_linux.c
+++ b/src/coredump/_UCD_access_reg_linux.c
@@ -57,6 +57,9 @@ _UCD_access_reg (unw_addr_space_t as,
 #elif defined(UNW_TARGET_S390X)
   if (regnum > UNW_S390X_R15)
     goto badreg;
+#elif defined(UNW_TARGET_IA64)
+  if (regnum >= ARRAY_SIZE(ui->prstatus->pr_reg))
+    goto badreg;
 #else
 #if defined(UNW_TARGET_MIPS)
   static const uint8_t remap_regs[] =


djwatson:
Merged, thanks.

src/coredump/_UCD_access_reg_linux.c
```diff
@@ -57,6 +57,9 @@ _UCD_access_reg (unw_addr_space_t as,
 #elif defined(UNW_TARGET_S390X)
   if (regnum > UNW_S390X_R15)
     goto badreg;
+#elif defined(UNW_TARGET_IA64)
```
Arfrever, line 60: Why only `defined(UNW_TARGET_IA64)` though?

[That patch in Gentoo](https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-libs/libunwind/files/libunwind-1.2-coredump-regs.patch?id=e01ced2da69c0d913e4c91467a0ca5fc3615edcd) has:
```
#elif defined(UNW_TARGET_IA64) || defined(UNW_TARGET_HPPA) || defined(UNW_TARGET_PPC32) || defined(UNW_TARGET_PPC64)
```

kasperk81: I extracted the patch from Debian which pointed to Gentoo issue. Makes sense to add those conditions.

kasperk81: Fixed in #269.

Issue #263, opened 2021-06-24
x86_64 registers have incorrect values during CFI/CFA calculations


tmshort:
Some assembly code (e.g. OpenSSL's `bn_mul_mont()` and `bn_mult_mont_gather5()`) uses CFI/CFA expressions that are somewhat complicated, in that they use multiple registers in the calculation.

In this case, running libunwind periodically via a signal is used to determine where in the code the system is spending its time. Since OpenSSL has a number of functions that can take a while, they are vulnerable to being interrupted by this signal.

While doing these calculations, it appears that some of the registers used have incorrect values. However, `gdb` is able to complete a backtrace.

Example:
In OpenSSL_1_1_1-stable, crypto/bn/asm/x86_64-mont.pl (source of `bn_mul_mont()`) there is the following CFI/CFA expression:
```
.cfi_cfa_expression        %rsp+8,$num,8,mul,plus,deref,+8
```
If `unw_backtrace()` is called (via a signal for instance) and it attempts to run the calculation over the expression, a segfault happens. Here is the output (tweaked to include values) via `UNW_DEBUG_LEVEL=15`:
(Note that `%r7` is an alias to `%rsp` and `$num` is `%r9`.)
```
              >_ULx86_64_dwarf_eval_expr: len=10, pushing initial value=0x0
               >_ULx86_64_dwarf_eval_expr: push -> 0x0
               >_ULx86_64_dwarf_eval_expr: OP_breg(r7,0x8)
               >_ULx86_64_dwarf_eval_expr: push -> 0x7ffdbe4e8c08
               >_ULx86_64_dwarf_eval_expr: OP_breg(r9,0x0)
               >_ULx86_64_dwarf_eval_expr: push -> 0x7ffdbe4e7590
               >_ULx86_64_dwarf_eval_expr: OP_lit(8)
               >_ULx86_64_dwarf_eval_expr: push -> 0x8
               >_ULx86_64_dwarf_eval_expr: OP_mul
               >_ULx86_64_dwarf_eval_expr: pop -> 0x8
               >_ULx86_64_dwarf_eval_expr: pop -> 0x7ffdbe4e7590
               >_ULx86_64_dwarf_eval_expr: push -> 0x3ffedf273ac80
               >_ULx86_64_dwarf_eval_expr: OP_plus
               >_ULx86_64_dwarf_eval_expr: pop -> 0x3ffedf273ac80
[...]

PR #265, opened 2021-06-25
x86_64/GTrace:  Fix usage of unsaved registers

djwatson:
Some cfis may access registers that aren't saved on the trace path.
Set all these registers to null locations, so that attempting to use
them causes us to fall back to the slow path.

### DIFF
--- a/src/x86_64/Gtrace.c
+++ b/src/x86_64/Gtrace.c
@@ -207,10 +207,7 @@ trace_cache_get (void)
    RIP using current CFA, RBP and RSP values.  Modifies CURSOR to
    that location, performs one unw_step(), and fills F with what
    was discovered about the location.  Returns F.
-
-   FIXME: This probably should tell DWARF handling to never evaluate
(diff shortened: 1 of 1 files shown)


tmshort:
This doesn't seem to handle the case of R9, as seen in #263.

djwatson:
Can you be more specific? When I apply this patch to your test code it seems to work for me.

d->loc[UNW_X86_64_R9] gets set to NULL_LOC in the loop, resulting in an error if it is accessed.

tmshort:
I haven't tried it yet... (running a LONG test). Does it error our, or does the stack trace complete?

djwatson:
It fails the 'fast' getcontext_trace path, and falls back to slow_backtrace, which fetches all the registers, and completes successfully.

tmshort:
Gotcha. I will try this in a bit. Thanks.

tmshort:
LGTM. I can't mark this approved.

PR #266, opened 2021-06-25
DWARF: dwarf4 fix cie parsing

djwatson:
dwarf 4 addds address_size and segment_size fields. Read these before code/data align fields.

### DIFF
--- a/src/dwarf/Gfde.c
+++ b/src/dwarf/Gfde.c
@@ -48,6 +48,7 @@ parse_cie (unw_addr_space_t as, unw_accessors_t *a, unw_word_t addr,
            int is_debug_frame, void *arg)
 {
   uint8_t version, ch, augstr[5], fde_encoding, handler_encoding;
+  uint8_t address_size, segment_size;
   unw_word_t len, cie_end_addr, aug_size;
   uint32_t u32val;
   uint64_t u64val;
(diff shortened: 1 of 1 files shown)


src/dwarf/Gfde.c
```diff
@@ -138,6 +139,15 @@ parse_cie (unw_addr_space_t as, unw_accessors_t *a, unw_word_t addr,
         augstr[i++] = ch;
     }
 
+  if (version > 3)
```
kasperk81, line 142: Are we parsing the version correctly? I am asking because, I tried adding `printf(">>> the version is: %d\n", version);` before this condition and did:

```sh
./autogen.sh && make && make check

readelf --debug-dump=info tests/.libs/Ltest-mem-validate | grep -A 2 'Compilation Unit @'
```
it shows:
```
  Compilation Unit @ offset 0x0:
   Length:        0xe36 (32-bit)
   Version:       4
```

but when executing it shows, `>>> the version is: 1`

Same result if when explicitly setting dwarf version to 5:
```sh
autoreconf -i && ./configure CFLAGS="-gdwarf-5" CXXFLAGS="-gdwarf-5" && make -j && make check
```
readelf shows `Version:       5` but our dwarf parser still gets version 1.

and same thing happens if `CC=$(command clang) CXX=$(command clang++)`

does that mean the dwarf version parsing is wrong?

djwatson: Yea I noticed this as well, I always see 1.  I'll dig in to it when I get a chance

kasperk81: If this version is supposed to be the same as Compilation Unit version from the readelf output above, then the first difference I noticed is that we are reading one byte for version, while binutils uses two bytes (`unsigned short`).

Definition: https://github.com/bminor/binutils-gdb/blob/3451a2d7a3501e9c3fc344cbc4950c495f30c16d/binutils/dwarf.h#L59
Reading: https://github.com/bminor/binutils-gdb/blob/3451a2d7a3501e9c3fc344cbc4950c495f30c16d/binutils/dwarf.c#L3561

mstorsjo:
FWIW, I ran into #256, and I can confirm that this PR fixes the issue.

djwatson:
I haven't yet found time to work on this further.  Since it seems it fixes some issues and doesn't cause any more, I'm going to commit it.  

Feel free to send PRs for any address_size fixes, etc.

PR #270, opened 2021-06-28
Use the BUILD / HOST convention in .travis.yml

### LINKED ISSUE #268: Cross compilation is actually broken
If we look at the CI logs, e.g. s390x https://travis-ci.com/github/libunwind/libunwind/jobs/519138896, we see:

> checking whether we are cross compiling... no

so we end up compiling everything with host (x86_64) compiler.
the reason for that is; HOST is not set in default environment (stare at .travis.yml for a while; it's a tricky one).

But what really happens when we set HOST and TARGET correctly is we get

> checking if libtool supports shared libraries... **no**
> checking whether to build shared libraries... **no**

and as a result, we only get src/.libs/libunwind.a archive rather than an SO file.

Compelete examples:

1. cross compile arm-linux-gnueabihf on Ubuntu x64:

```sh
docker run -it ubuntu:latest

apt update
apt install -y git automake make libtool-bin g++-arm-linux-gnueabihf

git clone https://github.com/libunwind/libunwind
cd libunwind
autoreconf -i

./configure --enable-shared \
    CC=arm-linux-gnueabihf-gcc \
    CXX=arm-linux-gnueabihf-g++ \
    AR=arm-linux-gnueabihf-ar \
    RANLIB=arm-linux-gnueabihf-ranlib \
    STRIP=arm-linux-gnueabihf-strip \
    --target=arm-unknown-linux-gnueabihf \
    --host=x86_64-linux-gnu > fulllog 2>&1

# check ./fulllog which says:

# checking whether the arm-linux-gnueabihf-gcc linker (/usr/arm-linux-gnueabihf/bin/ld -m elf_i386) supports shared libraries... /usr/arm-linux-gnueabihf/bin/ld: unrecognised emulation mode: elf_i386
# Supported emulations: armelf_linux_eabi armelfb_linux_eabi
# no
# checking dynamic linker characteristics... GNU/Linux ld.so
# checking how to hardcode library paths into programs... unsupported
# checking whether stripping libraries is possible... yes
# checking if libtool supports shared libraries... no
# checking whether to build shared libraries... no
# checking whether to build static libraries... yes


# we expect all shared libraries lines to be "yes"
```

2. cross compile IA64 on gentoo x64

```sh
docker run -it gentoo/stage3
# setup and update packages repository
emerge-webrsync
emerge sys-devel/crossdev app-eselect/eselect-repository dev-vcs/git
mkdir -p /etc/portage/repos.conf/
eselect repository create localrepo

# emerge ia64 cross compiler
crossdev -t ia64-unknown-linux-gnu --with-headers

# build libunwind
git clone https://github.com/libunwind/libunwind
cd libunwind
autoreconf -i
./configure --enable-shared \
    CC=/usr/x86_64-pc-linux-gnu/ia64-unknown-linux-gnu/gcc-bin/11.1.0/ia64-unknown-linux-gnu-gcc \
    CXX=/usr/x86_64-pc-linux-gnu/ia64-unknown-linux-gnu/gcc-bin/11.1.0/ia64-unknown-linux-gnu-g++ \
    --host=x86_64-pc-linux-gnu \
    --target=ia64-unknown-linux-gnu > fulllog 2>&1

# now check the "fulllog" file, same situation that shared libraries lines are "no" instead of "yes"
```

kasperk81:
Fixes #268

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,20 +2,20 @@ sudo: required
 language: c
 compiler: gcc
 env:
-- TARGET=x86_64-linux-gnu
-- TARGET=x86-linux-gnu
-- TARGET=arm-linux-gnueabihf
-- TARGET=aarch64-linux-gnu
-- TARGET=mipsel-unknown-linux-gnu
+- HOST=x86_64-linux-gnu
+- HOST=x86-linux-gnu
+- HOST=arm-linux-gnueabihf
+- HOST=aarch64-linux-gnu
+- HOST=mipsel-linux-gnu
 # Currently experiencing build failures here
-#- TARGET=powerpc64-linux-gnu
+#- HOST=powerpc64-linux-gnu
 
 linux-s390x: &linux-s390x
   os: linux
   arch: s390x
-  env: TARGET=s390x-linux-gnu
+  env: BUILD=s390x-linux-gnu HOST=s390x-linux-gnu
   script:
-    - ./autogen.sh
+    - autoreconf -i
     - ./configure
     - make -j32
     - ulimit -c unlimited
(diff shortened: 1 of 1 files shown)


kasperk81:
@trofi, @djwatson, main diff of this change (compared to current master), from `arm-linux-gnueabihf` leg:
```diff
--- before (master)
+++ after (this PR)

checking build system type... x86_64-pc-linux-gnu
- checking host system type... x86_64-pc-linux-gnu
+ checking host system type... arm-unknown-linux-gnueabihf
checking target system type... arm-unknown-linux-gnueabihf
```

and from  `HOST=x86-linux-gnu`, we now get:
````
$ file src/.libs/libunwind.so.8.0.1

src/.libs/libunwind.so.8.0.1: ELF 32-bit LSB shared object, Intel 80386, version 1 (SYSV), dynamically linked, BuildID[sha1]=59e5ec6b4143a9d3348d5a9c04d109e0906ed1cc, not stripped
````

File: .travis.yml [hunk 1]
```diff
@@ -2,20 +2,20 @@ sudo: required
...
 
 linux-s390x: &linux-s390x
   os: linux
   arch: s390x
-  env: TARGET=s390x-linux-gnu
+  env: BUILD=s390x-linux-gnu HOST=s390x-linux-gnu
   script:
-    - ./autogen.sh
+    - autoreconf -i
     - ./configure
```
kasperk81, line 19: when we run ./autogen.sh, it does autoreconf then run configure. so it was redundant. in cross compilation case (below), on top of running configure twice, it gives us confusing logs.

trofi:
The `--build=/--host=` change looks very reasonable to me! Not familiar with rest of CI stuff.

.travis.yml
djwatson: Yea we can probably dump autogen.sh, I never use it, anyway.

PR #272, opened 2021-06-30
Show contents of test-suite.log on test failure

kasperk81:
With intetntional failure, the bottom part of failing leg looks like this: https://travis-ci.com/github/kasperk81/libunwind/jobs/520324857#L1485 (click on it and it shows contents of `test-suite.log`). Only applicable to the job which failed due to test failure.

SEcond commit is removing autogen.sh wrapper and updating README.md.


djwatson:
Thanks!

PR #267, opened 2021-06-26
Add port for Linux on RISC-V (riscv)

zhaofengli:
This PR adds basic support for Linux on RISC-V (riscv), with only 64-bit supported for now. All tests in the test suite are passing.

<del>All of the tests pass except for Ltest-mem-validate, for which DWARF-based unwinding fails when trying to access the mprotected region (more details below). I'm having trouble understanding what the correct way to solve this issue is and some pointers will be really appreciated.</del> See https://github.com/libunwind/libunwind/pull/267#issuecomment-869204461

Fixes #99.

```
PASS: test-proc-info
PASS: test-static-link
PASS: test-strerror
PASS: Gtest-bt
PASS: Ltest-bt
PASS: Gtest-exc
PASS: Ltest-exc
PASS: Gtest-init
PASS: Ltest-init
PASS: Gtest-concurrent
PASS: Ltest-concurrent
PASS: Gtest-resume-sig
PASS: Ltest-resume-sig
PASS: Gtest-resume-sig-rt
PASS: Ltest-resume-sig-rt
PASS: Gtest-trace
PASS: Ltest-trace
PASS: Ltest-init-local-signal
PASS: Ltest-mem-validate
PASS: test-async-sig
PASS: test-flush-cache
PASS: test-init-remote
PASS: test-mem
PASS: test-reg-state
PASS: Ltest-varargs
PASS: Ltest-nomalloc
PASS: Ltest-nocalloc
PASS: Lrs-race
PASS: test-ptrace
PASS: test-setjmp
PASS: Ltest-cxx-exceptions
PASS: run-check-namespace
XFAIL: run-ptrace-mapper
XFAIL: run-ptrace-misc
[124995.866301] crasher[32193]: unhandled signal 11 code 0x1 at 0x0000000000000020 in crasher[2adb244000+1000]
[124995.869723] CPU: 3 PID: 32193 Comm: crasher Tainted: G            E     5.10.0-7-riscv64 #1 Debian 5.10.40-1
[124995.870271] epc: 0000002adb244906 ra : 0000002adb244918 sp : 0000003fffa8b460
[124995.870596]  gp : 0000002adb246800 tp : 0000003ff38df250 t0 : 0000003ff37d80b0
[124995.870928]  t1 : 0000003ff37f3f4c t2 : 0000003ff38d9ea0 s0 : 0000002adb24494c
[124995.871577]  s1 : 0000000000000000 a0 : 0000000000000001 a1 : 0000002b01773048
[...]

PR #276, opened 2021-07-02
Don't force {exec_prefix}/lib64 libdir on ppc64

matthiasdiener:
See also https://sources.debian.org/patches/libunwind/1.3.2-2/20140507-ppc_lib64.patch/

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -213,12 +213,6 @@ fi
 AM_CONDITIONAL(USE_DWARF, [test x$use_dwarf = xyes])
 AC_MSG_RESULT([$use_dwarf])
 
-if test x$target_arch = xppc64; then
-        libdir='${exec_prefix}/lib64'
-        AC_MSG_NOTICE([PowerPC64 detected, lib will be installed ${libdir}]);
-        AC_SUBST([libdir])
-fi
-
 AC_MSG_CHECKING([whether to restrict build to remote support])
 if test x$target_arch != x$host_arch; then
   CPPFLAGS="${CPPFLAGS} -DUNW_REMOTE_ONLY"


kasperk81:
:+1: most (if not all) the Debian and Gentoo patches have been upstreamed in one form or the other. Would be nice to cut the v1.6 soon with RISC-V support (#267), so we can get a robust/patch-free libunwind available in packages.

matthiasdiener:
The CI failure seems to be spurious

kasperk81:
Yes, that intermittent test hang in s390x CI leg is something which @uweigand might be able to help with. :eyes:

uweigand:
> Yes, that intermittent test hang in s390x CI leg is something which @uweigand might be able to help with. eyes

I notice that the single test that is not reported as PASS in the travis log is test-async-sig.   However, I'm not able to reproduce any hang on my machine (Ubuntu 20.04) even after many re-runs.

That said, the test-async-sig test seems to test the system (e.g. proper unwind info everywhere in glibc) just as much as the actual libunwind code base, so this may simply be a problem in the environment.  Do you know which OS is installed on the travis builder?

matthiasdiener:
Closing/reopening to restart CI

kasperk81:
@uweigand, from [logs](https://app.travis-ci.com/github/libunwind/libunwind/jobs/522056699#L7) this build info is shown:

```
Build image provisioning date and time
Mon Jan 20 10:45:28 UTC 2020
Operating System Details
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.6 LTS
Release:	16.04
Codename:	xenial
Linux Version
5.3.0-26-generic
```

uweigand:
@kasperk81 thanks, I did not notice this.   Ubuntu 16.04 is quite old - I believe it is already out of support unless you have a paid-for extended maintenance contract, so even if we did identify a bug there, it's unlikely we would be able to fix it at this stage.

I don't know exactly how this works with Travis - Is there a way to move to a more recent OS version?

kasperk81:
Indeed 16.04 has reached EOL. It is possible to move the builds to GitHub Actions, but GitHub Actions does not support s390x architecture natively. We can run the build in qemu/docker, but testing libunwind would not yield the same result as baremetal.

BTW, yesterday libunwind's TravisCI account ran out of credits, so it's not a free option anymore. Another reason to consider GitHub Actions with some compromises (that we cannot reliable run s390x tests there, and only reliably cross-compile the code on that arch for validation).

zhaofengli:
> We can run the build in qemu/docker, but testing libunwind would not yield the same result as baremetal.

I don't think there is any difference between testing in qemu and on baremetal for this. It's just that setting it up is a bit tricky (we cannot use qemu-user and have to use full-system emulation).

kasperk81:
BTW, run-coredump-unwind and run-coredump-unwind-mdi both tests consistently fail on Ubuntu 18.04 and 20.04 x86_64. It is either the old version of gcc/g++ (some tests are in C++) or something else in that environment libc/kernel that makes them succeed on 16.04 (which has already reached its end of life and CI vendors are moving on from it). We should disable that test or fix the reason why https://github.com/libunwind/libunwind/blob/c720133faa27293a0e8e075ddf9ccefa55a2efa1/tests/test-coredump-unwind.c#L377-L380 condition satisfies (test_cur is 1 instead of 4, test_start_ips[1] is 1 and test_start_ips[2] is -1 on ubuntu focal with -O0, -O2 as well as -O3)...

PR #279, opened 2021-07-06
Test fixes

zhaofengli:
A few fixes to the tests that were left out of #267. Tested on riscv64 and x86-64.

Pinging @jim-wilson since the last commit was made in your name (https://github.com/libunwind/libunwind/pull/267#issuecomment-871743659).

### DIFF
--- a/tests/Ltest-mem-validate.c
+++ b/tests/Ltest-mem-validate.c
@@ -47,8 +47,11 @@ void * stack_start;
 
 void do_backtrace (void)
 {
-  void* buffer[1024];
-  int size = 1024;
+  /*
+    We make the assumption that we are able to rewind far enough
+    (steps > 5) before touching the forbidden region in the stack,
+    at which point the unwinding should stop gracefully.
+  */
   mprotect((void*)((uintptr_t)stack_start & ~(PAGE_SIZE - 1)),
            PAGE_SIZE, PROT_NONE);
 
@@ -68,6 +71,7 @@ void do_backtrace (void)
       unw_get_reg (&cursor, UNW_REG_SP, &sp);
 
       ret = unw_step (&cursor);
+	  printf("ip=%lx, sp=%lx -> %d\n", ip, sp, ret);
       if (ret < 0)
 	{
 	  unw_get_reg (&cursor, UNW_REG_IP, &ip);
@@ -78,14 +82,16 @@ void do_backtrace (void)
 
   if (steps < 5)
     {
+      printf("not enough steps: %d, need 5\n", steps);
       exit(-1);
     }
+  printf("success, steps: %d\n", steps);
 
   mprotect((void*)((uintptr_t)stack_start & ~(PAGE_SIZE - 1)),
            PAGE_SIZE, PROT_READ|PROT_WRITE);
 }
 
-void consume_and_run (int depth)
+void NOINLINE consume_and_run (int depth)
 {
   unw_cursor_t cursor;
   unw_context_t uc;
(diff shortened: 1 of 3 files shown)


zhaofengli:
I was able to reproduce the spurious failure in an Ubuntu 16.04 (same as CI, glibc 2.23) container on my end, and it was due to libc segfaulting in `malloc/arena.c` trying to allocate buffers for the printf debugging aids. It was trying to access something in the mprotected region, beyond the stack frame for main(). This is internal to the libc implementation and we shouldn't make the assumption that it will never be accessed. Let's "fix" it by making the frame for main() larger.

jim-wilson:
I was looking at the Ltest-mem-validate failure too.  PIE plus ASLR means the area protected by the mprotect call varies from run to run as the starting stack address varies from run to run.  Thus sometimes it works and sometimes it doesn't work.  Making the main stack size at least one PAGE_SIZE big looks like a good solution.

I see that travis is running ./configure CFLAGS="" so everything is being compiled without optimization when running the tests, so the NOINLINE isn't part of the problem here.

tests/Ltest-mem-validate.c
```diff
@@ -108,6 +114,9 @@ main (int argc, char **argv UNUSED)
 
   stack_start = &start;
 
+  char string[4096];
```
jim-wilson, line 117: I'd suggest using PAGE_SIZE here to match the code above.  And maybe a comment mentioning that we need to allocate as much as is being protected by the mprotect call.

jim-wilson (approved):

kasperk81:
> I see that travis is running ./configure CFLAGS="" so everything is being compiled without optimization when running the tests, so the NOINLINE isn't part of the problem here.

I think we should just update .travis.yml by adding `CFLAGS=" -O2" CXXFLAGS=" -O2"` (cxx is used in test), so it uses -O2 on all platforms. e.g. on s390x, it is using the default (also unspecified in .travis.yml) `-O2` (https://app.travis-ci.com/github/libunwind/libunwind/jobs/522226826#L365). As "without optimization" testing is less interesting than the commonly used -O2.

zhaofengli:
> > I see that travis is running ./configure CFLAGS="" so everything is being compiled without optimization when running the tests, so the NOINLINE isn't part of the problem here.
> 
> I think we should just update .travis.yml by adding `CFLAGS=" -O2" CXXFLAGS=" -O2"` (cxx is used in test), so it uses -O2 on all platforms. e.g. on s390x, it is using the default (also unspecified in .travis.yml) `-O2` (https://app.travis-ci.com/github/libunwind/libunwind/jobs/522226826#L365). As "without optimization" testing is less interesting than the commonly used -O2.

Maybe we can test both? Also while we are it, we should consider switching to GitHub Actions since it has better integration and [Travis CI seems to be going downhill](https://daniel.haxx.se/blog/2021/06/14/bye-bye-travis-ci/).

kasperk81:
Yes both would be good (we would need to add another dimension to the matrix for linux).
GitHub Action does not provide IBM archs like s390x, which is the main reason of using TravisCI afaict.

zhaofengli:
> GitHub Action does not provide IBM archs like s390x, which is the main reason of using TravisCI afaict.

That's a good point. We could run qemu-system with Ubuntu's prebuilt images, or just perform cross compile-only tests like what we do for arm/mips at the moment. qemu-user is out of the equation for libraries like this. Another option is to keep the native s390x tests on Travis CI.

Anyways, we should discuss more and these changes belong to a different PR.

zhaofengli:
See also OSU Open Source Lab's IBM Z CI offering at https://osuosl.org/services/ibm-z.

PR #283, opened 2021-07-10
Include asm/ptrace.h in _UPT_reg_offset.c

kasperk81:
This was needed for musl.


kasperk81:
upstreamable variant of https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/main/libunwind/ppc64-musl.patch
@TBK for your info, i have upstreamed few patches from varios distros this way for upcoming 1.6.

PR #284, opened 2021-07-12
Add optimization options in .travis.yml

kasperk81:
Currently CI only tests (implicit) -O0. This adds -O0, -O2 and -O3 explicitly. CI builds the whole matrix less than three minutes so it won't be a hassle but keep us up to date.

cc @zhaofengli


kasperk81:
This also enables ppc64le build, which works fine.

PR #295, opened 2021-08-28
Fix the build on FreeBSD

ararslan:
The first commit here fixes building the library and the second fixes building `test-ptrace.c`. See individual commit messages for explanation. Note that `test-ptrace` still fails on FreeBSD, but at least now it's able to build.

Fixes #294.

PR #296, opened 2021-08-29
x86_64: Stop aliasing RSP and CFA

ararslan:
This commit is by @Keno. We've been using it as a local patch for Julia (available [here](https://github.com/JuliaLang/julia/commit/55d7571a5dc2686c98a804d95ba36f5548f53f18)), applied to libunwind 1.3.2. I've adapted it to current libunwind master, hopefully correctly.

> RSP and CFA are different concepts. RSP refers to the physical register, CFA is a virtual register that serves as the base address for various other saved registers. It is true that in many frames these are set to alias, however this is not a requirement. For example, a function that performs a stack switch would likely change the rsp in the middle of the function, but would keep the CFA at the original RSP such that saved registers may be appropriately recovered.
>
> We are seeing incorrect unwinds in the Julia runtime when running julia under rr. This is because injects code (with correct CFI) that performs just such a stack switch [1]. GDB manages to unwind this correctly, but libunwind incorrectly sets the rsp to the CFA address, causing a misunwind.
>
> Tested on x86_64, patches for other architectures are ported, but not tested.
>
> [1] https://github.com/rr-debugger/rr/blob/469c22059a4a1798d33a8a224457faf22b2c178c/src/preload/syscall_hook.S#L454

### DIFF
--- a/include/dwarf.h
+++ b/include/dwarf.h
@@ -231,6 +231,7 @@ typedef enum
     DWARF_WHERE_REG,            /* register saved in another register */
     DWARF_WHERE_EXPR,           /* register saved */
     DWARF_WHERE_VAL_EXPR,       /* register has computed value */
+    DWARF_WHERE_CFA,            /* register is set to the computed cfa value */
   }
 dwarf_where_t;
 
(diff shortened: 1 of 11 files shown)


ararslan:
CI failure is the same as #295, stalled build on s390x.

djwatson:
Thanks!

PR #297, opened 2021-08-29
Prefer EXTBL unwinding on ARM

ararslan:
This commit is by @yuyichao. We've been using it as a local patch for Julia (available [here](https://github.com/JuliaLang/julia/commit/8624e4ad9a5a4717ecd169899090ccdf5f8a5aed#diff-5e23f944dcfb9e5b600d942c6c3dba6b24ec73e638741ec3cc192cfdefb29c5d)) since 2016, currently applied to libunwind 1.3.2. I've adapted it to current libunwind master, hopefully correctly.

> It is part of the C++ ABI so a EXTBL unwind info that's not `CANT_UNWIND` should always be reliable/correct. Ignore `ESTOPUNWIND` so that a `CANT_UNWIND` info can fallback to unwinding using the debug info instead.

### DIFF
--- a/include/tdep-arm/libunwind_i.h
+++ b/include/tdep-arm/libunwind_i.h
@@ -256,6 +256,7 @@ dwarf_put (struct dwarf_cursor *c, dwarf_loc_t loc, unw_word_t val)
 #define tdep_init_done                  UNW_OBJ(init_done)
 #define tdep_init                       UNW_OBJ(init)
 #define arm_find_proc_info              UNW_OBJ(find_proc_info)
+#define arm_find_proc_info2             UNW_OBJ(find_proc_info2)
 #define arm_put_unwind_info             UNW_OBJ(put_unwind_info)
 /* Platforms that support UNW_INFO_FORMAT_TABLE need to define
    tdep_search_unwind_table.  */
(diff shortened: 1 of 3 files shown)


ararslan:
Yichao, let me know if you'd like to make any changes to this patch and I'll give you commit access to my fork. Or if you think this patch isn't appropriate upstream, I can close this PR and we can continue to carry the patch for Julia.

yuyichao:
Someone had a comment about not wanting this order or something on the mailing list IIRC.
Also I just realized that the debug message needs to be updated too. It currently report about dwarf failing which isn't the case anymore.

yuyichao:
Updated the logging message and fixed the retry logic (which was already included in the commit message somehow so maybe it was lost during rebase or something?)

IIRC, the original concern is that there are cases where the exidx unwind info says stop unwind (CANT_UNWIND) but the dwarf info can do so. In the original version, this will unwind successfully since the dwarf will be tried first whereas the first version of the patch would not since it sees the stop unwind first and stops there. In the new version of the patch, the cant_unwind will not prevent us from trying dwarf anymore but it'll still prevent the frame pointer fallback just like before.

djwatson:
Seems reasonable to me, thanks!

PR #299, opened 2021-08-30
Expose function-pointer-to-name API

### LINKED ISSUE #288: Expose function-pointer-to-name API
The functionality to get a function name given a function pointer is implemented but not exposed. Would it be acceptable to expose a function for this, e.g. `get_proc_name_by_ip()`?

Our use case is logging, in situations where we have a pointer and want to log a readable function name for later analysis.

While searching, I found two old mail threads:
* https://lists.nongnu.org/archive/html/libunwind-devel/2012-01/msg00002.html
* https://lists.nongnu.org/archive/html/libunwind-devel/2014-06/msg00017.html

The latter contains a patch. I could try to rebase (or rewrite) that patch, if there is a change this can be accepted. Is it?

bjosv:
The added API `unw_get_proc_name_by_ip()` returns the name of a procedure just like `unw_get_proc_name()`,
except that the name is looked up by an instruction-pointer instead of a cursor.

This enables users to get the name of a procedure in a corresponding way as to get the auxiliary
information using `unw_get_proc_info_by_ip()`.

`doc/unw_get_proc_name_by_ip.man` is generated using latex2man [v1.24](https://www.informatik-vollmer.de/software/latex2man-1.24.tar.gz).

Fixes #288

### DIFF
--- a/doc/Makefile.am
+++ b/doc/Makefile.am
@@ -8,6 +8,7 @@ man3_MANS = libunwind.man libunwind-dynamic.man libunwind-ia64.man	\
 	unw_get_proc_info.man						\
 	unw_get_proc_info_by_ip.man					\
 	unw_get_proc_name.man						\
+	unw_get_proc_name_by_ip.man					\
 	unw_get_fpreg.man						\
 	unw_get_reg.man							\
 	unw_getcontext.man						\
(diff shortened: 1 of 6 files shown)


djwatson:
Sorry for the late response - sure, seems reasonable to me.  Thanks!

PR #300, opened 2021-08-31
Implement tdep_get_elf_image for Solaris/illumos

NanXiao:
Current `tdep_get_elf_image()` uses Linux APIs, and it doesn't work. Use code from Solaris/illumos pmap to implement `tdep_get_elf_image()`.

I used [libunwind_backtrace](https://github.com/NanXiao/code-for-my-blog/tree/master/2021/08/libunwind_backtrace) for test:

a) The result of master branch:
```
$ cat /tmp/backtrace.log
0x401b3b: -- error(unspecified (general) error): unable to obtain symbol name for this frame
0x401b47: -- error(unspecified (general) error): unable to obtain symbol name for this frame
0x401b5e: -- error(unspecified (general) error): unable to obtain symbol name for this frame
0x401757: -- error(unspecified (general) error): unable to obtain symbol name for this frame
0x4016b8: -- error(unspecified (general) error): unable to obtain symbol name for this frame
```

b) The result of my code:
```
$ cat /tmp/backtrace.log
0x401b3b: (foo+0x9)
0x401b47: (bar+0x9)
0x401b5e: (main+0x14)
0x401757: (_start_crt+0x87)
0x4016b8: (_start+0x18)
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -351,7 +351,7 @@ AC_MSG_RESULT([$intel_compiler])
 
 AC_MSG_CHECKING([if building on Solaris then define __EXTENSIONS__ macro])
 if $OS_SOLARIS; then
-  CFLAGS="${CFLAGS} -D__EXTENSIONS__"
+  CFLAGS="${CFLAGS} -D__EXTENSIONS__ -lproc"
   AC_MSG_RESULT([yes])
 else
   AC_MSG_RESULT([no])
--- a/src/os-solaris.c
+++ b/src/os-solaris.c
@@ -27,38 +27,24 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 #include <stdio.h>
 
 #include "libunwind_i.h"
-#include "os-linux.h" // using linux header for map_iterator implementation
+#include "os-solaris.h"
 
 int
 tdep_get_elf_image (struct elf_image *ei, pid_t pid, unw_word_t ip,
                     unsigned long *segbase, unsigned long *mapoff,
                     char *path, size_t pathlen)
 {
-  struct map_iterator mi;
-  int found = 0, rc;
-  unsigned long hi;
-
-  if (maps_init (&mi, pid) < 0)
-    return -1;
-
-  while (maps_next (&mi, segbase, &hi, mapoff, NULL))
-    if (ip >= *segbase && ip < hi)
-      {
-        found = 1;
-        break;
-      }
-
-  if (!found)
+  char *elf_path = get_elf_path(pid, ip);
+  if (elf_path == NULL)
     {
-      maps_close (&mi);
       return -1;
     }
   if (path)
     {
-      strncpy(path, mi.path, pathlen);
+      strncpy(path, elf_path, pathlen);
     }
-  rc = elf_map_image (ei, mi.path);
-  maps_close (&mi);
+  int rc = elf_map_image (ei, elf_path);
+  free(elf_path);
   return rc;
 }
 
--- a/src/os-solaris.h
+++ b/src/os-solaris.h
@@ -1,0 +1,134 @@
+/*
+ * CDDL HEADER START
+ *
+ * The contents of this file are subject to the terms of the
+ * Common Development and Distribution License (the "License").
+ * You may not use this file except in compliance with the License.
+ *
+ * You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+ * or http://www.opensolaris.org/os/licensing.
+ * See the License for the specific language governing permissions
+ * and limitations under the License.
+ *
+ * When distributing Covered Code, include this CDDL HEADER in each
+ * file and include the License file at usr/src/OPENSOLARIS.LICENSE.
+ * If applicable, add the following below this CDDL HEADER, with the
+ * fields enclosed by brackets "[]" replaced with your own identifying
+ * information: Portions Copyright [yyyy] [name of copyright owner]
+ *
+ * CDDL HEADER END
+ */
+
+/*
+ * Copyright 2008 Sun Microsystems, Inc.  All rights reserved.
+ * Use is subject to license terms.
+ */
+
+#ifndef os_solaris_h
+#define os_solaris_h
+
+#include <stdio.h>
+#include <string.h>
+#include <libproc.h>
+#include <limits.h>
+
+typedef struct
+{
+  prmap_t    md_map;
+  char    *md_objname;
+} mapdata_t;
+
+static  mapdata_t  *maps;
+static  int    map_count;
+static  int    map_alloc;
+
+static mapdata_t *
+nextmap(void)
+{
+  if (map_count == map_alloc)
+  {
+    int next;
+    if (map_alloc == 0)
+      next = 16;
+    else
+      next = map_alloc * 2;
+
+    mapdata_t *newmaps = realloc(maps, next * sizeof (mapdata_t));
+    if (newmaps == NULL)
+      return NULL;
+
+    memset(newmaps + map_alloc, '\0', (next - map_alloc) * sizeof (mapdata_t));
+
+    map_alloc = next;
+    maps = newmaps;
+  }
+
+  return &maps[map_count++];
+}
+
+static int
+gather_map(void *ignored, const prmap_t *map, const char *objname)
+{
+  mapdata_t *data = nextmap();
+  if (data == NULL)
+    return 1;
+
+  data->md_map = *map;
+  data->md_objname = objname ? strdup(objname) : NULL;
+  return 0;
+}
+
+static char *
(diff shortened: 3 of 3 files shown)


am11:
If it is fixing any tests (reported by `autoreconf -i && make && make test`), please update the README accordingly: https://github.com/libunwind/libunwind#expected-results-on-solaris-x86-64.

NanXiao:
@am11 Actually I didn't have any access to `Solaris` physical machine and I only have one `illumos` virtual machine. TBH, I don't know whether there are some differences between results of physical and virtual machines. 
(1) There are `8` failed test cases of master branch on `illumos` virtual machine:
```
Gtest-resume-sig
Ltest-resume-sig
Gtest-resume-sig-rt
Ltest-resume-sig-rt
Ltest-init-local-signal
test-setjmp
x64-unwind-badjmp-signal-frame
run-check-namespace
```
(2) There are `9` failed test cases of my PR on `illumos` virtual machine (`test-async-sig` also fails):
```
Gtest-resume-sig
Ltest-resume-sig
Gtest-resume-sig-rt
Ltest-resume-sig-rt
Ltest-init-local-signal
test-async-sig
test-setjmp
x64-unwind-badjmp-signal-frame
run-check-namespace
```
You can see quite different from [https://github.com/libunwind/libunwind#expected-results-on-solaris-x86-64](https://github.com/libunwind/libunwind#expected-results-on-solaris-x86-64): E.g.: `Gtest-concurrent` and `Ltest-concurrent` no longer fail.

So should I add a new section for test results of `illumos`? Or just update on `Solaris x86-64` section? Thanks in advance!

am11:
> run-check-namespace

For this one, I temporarily move `nm(1)` and `grep(1)` and replace it with GNU variants, to make this test pass:

```sh
# temporary move
mv /usr/bin/nm /usr/bin/nm2 && mv /usr/bin/gnm /usr/bin/nm
mv /usr/bin/grep /usr/bin/grep2 && mv /usr/bin/ggrep /usr/bin/grep

# run tests
make check

# revert
mv /usr/bin/nm /usr/bin/gnm && mv /usr/bin/nm2 /usr/bin/nm
mv /usr/bin/grep /usr/bin/ggrep && mv /usr/bin/grep2 /usr/bin/grep
```
(would be nice to fix this test permanently without usage of `-u` with nm etc. at some point, but it's beyond the scope of this PR)

> test-async-sig

I also tested your patch on OmniOS r151038 (2021) and compared with master, this seems like a real regression in this PR.

Here is the backtrace (notice that sigsegv is emerging from `os-solaris.h:104`):

```gdb
$ LD_LIBRARY_PATH=src/.libs gdb ./tests/.libs/test-async-sig

(gdb) r
(gdb) thread apply all bt

Thread 2 (Thread 1 (LWP 1)):
#0  0xfffffc7fe8d9f00a in _lwp_kill () from /lib/64/libc.so.1
#1  0xfffffc7fe8d958b0 in thr_kill () from /lib/64/libc.so.1
#2  0xfffffc7fe8d32b9e in raise () from /lib/64/libc.so.1
#3  0xfffffc7fe8d0c818 in abort () from /lib/64/libc.so.1
#4  0xfffffc7fe8d0cc42 in _assert_c99 () from /lib/64/libc.so.1
#5  0xfffffc7fef2f0c9f in lookup_lofs_mount_point () from /lib/64/libproc.so.1
#6  0xfffffc7fef2f12c8 in Plofspath () from /lib/64/libproc.so.1
#7  0xfffffc7fef2f1a0c in Pzonepath () from /lib/64/libproc.so.1
#8  0xfffffc7fef2f1b7a in Pfindobj () from /lib/64/libproc.so.1
#9  0xfffffc7fef2d794b in Pexecname_live () from /lib/64/libproc.so.1
#10 0xfffffc7fef2e23bf in Pexecname () from /lib/64/libproc.so.1
#11 0xfffffc7fef2f1c8f in Pfindmap () from /lib/64/libproc.so.1
#12 0xfffffc7fef2e7ce3 in map_set () from /lib/64/libproc.so.1
#13 0xfffffc7fef2e7de6 in load_static_maps () from /lib/64/libproc.so.1
#14 0xfffffc7fef2e81f8 in Prd_agent () from /lib/64/libproc.so.1
#15 0xfffffc7fef2ec1c7 in i_Pmapping_iter () from /lib/64/libproc.so.1
#16 0xfffffc7fef2ec2ab in Pmapping_iter_resolved () from /lib/64/libproc.so.1
#17 0xfffffc7fef165519 in get_elf_path (pid=<optimized out>, ip=ip@entry=4200856) at os-solaris.h:104
#18 0xfffffc7fef1657d0 in _Ux86_64_get_elf_image (ei=ei@entry=0xfffffc7fffdfde50, pid=<optimized out>, ip=ip@entry=4200856, segbase=segbase@entry=0xfffffc7fffdfde40, mapoff=mapoff@entry=0xfffffc7fffdfde48, path=path@entry=0xfffffc7fffdfde60 "\200\336\337\377\177\374\377\377\333\255;\357\177\374\377\377)\002@", pathlen=1024) at os-solaris.c:37
#19 0xfffffc7fef16db0b in _Uelf64_get_proc_name (as=0xfffffc7fef182a40 <local_addr_space>, pid=<optimized out>, ip=4200856, buf=0xfffffc7fffdfe370 "", buf_len=256, offp=0xfffffc7fffdfe368) at elfxx.c:323
#20 0xfffffc7fef166549 in get_proc_name (arg=0xfffffc7fffdfe670, offp=0xfffffc7fffdfe368, buf_len=256, buf=0xfffffc7fffdfe370 "", ip=4200856, as=0xfffffc7fef182a40 <local_addr_space>) at mi/Gget_proc_name.c:93
#21 _ULx86_64_get_proc_name (cursor=cursor@entry=0xfffffc7fffdfe9d0, buf=buf@entry=0xfffffc7fffdfe370 "", buf_len=buf_len@entry=256, offp=offp@entry=0xfffffc7fffdfe368) at mi/Gget_proc_name.c:119
#22 0x0000000000401ab2 in do_backtrace (may_print=may_print@entry=0, get_proc_name=get_proc_name@entry=1) at test-async-sig.c:94
#23 0x0000000000401e52 in main (argc=<optimized out>, argv=<optimized out>) at test-async-sig.c:185
```

NanXiao:
@am11 IMHO, the reason is the `libproc` APIs are not async-signal safe (`libproc` uses locks internally, and it may be very difficult to implement async-signal safe), though I can't find the exact description from `illumos` manual. The `illumos` manual only states it's thread safe (https://illumos.org/man/3LIB/libproc#LOCKING):

> Most functions operate on a handle to a process in the form of a struct
     ps_prochandle *.  Unless otherwise indicated, the library does not provide
     any synchronization for different routines that are operating on the same
     libproc library handle.  It is up to the caller to ensure that only a
     single thread is using a handle at any given time.  Multiple threads may
     call libproc library routines at the same time as long as each thread is
     using a different handle.
> Each individual function notes its MT-Level section.  The MT-Level of a
     routine that matches the above description will refer to this manual page.
     If it does not, then it refers to the standard attributes in attributes(5).

I also wrote a simple test for multi-thread version (https://gist.github.com/NanXiao/c55f6a299a9739518ac69733430986c4), and it works:
```
$ ./libunwind_backtrace
0x401872: (foo+0x9)
0x40187e: (bar+0x9)
0xfffffc7fef1815ac: -- error(no unwind info found): unable to obtain symbol name for this frame
0xfffffc7fef1818c0: -- error(no unwind info found): unable to obtain symbol name for this frame
0x401872: (foo+0x9)
0x40187e: (bar+0x9)
0xfffffc7fef1815ac: -- error(no unwind info found): unable to obtain symbol name for this frame
0xfffffc7fef1818c0: -- error(no unwind info found): unable to obtain symbol name for this frame
```

IMHO, this should be OK to not support `get_proc_name()` in signal handler on `illumos`.

Thanks!

am11:
libunwind is a general-purpose library. TBH, regressing a working functionality to fix a bug in demo-like code does not sound like a good tradeoff as it is breaking other real-world use cases which are relying on `get_proc_name` in the handler.

> it may be very difficult to implement

I believe for a general-purpose library like libunwind, it is worthwhile to explore and discuss options to avoid any regression.

> The `illumos` manual only states it's thread safe

illumos manpages are sometimes not up-to-date, so we used source tree (https://github.com/illumos/illumos-gate) as the 'current' point of reference during the initial ports of libunwind and dotnet/runtime (where this illumos port was first used and tested with real-world workload).

Though I am not an expert on illumos kernel libraries. I think this approach is something we can discuss on IRC channel #illumos @libera.chat for insights if there is a better, lock-free facility available on illumos to implement it; without us resorting to maintaining manual synchronization between the threads (which I agree, could get verbose).

cc @jasonbking, @sjorge

djwatson:
It seems like this patchset still needs work for the async case?  Feel free to update & reopen.

PR #314, opened 2021-11-27
Add more POSIX polyfills for nix cross-build on win

am11:
A few definitions required by v1.6.0 were missing in abstraction layer.


am11:
@djwatson, this patch is needed by .NET runtime to consume v1.6. Could you please take a look?

djwatson:
Thanks.  This will end up in master & 1.6.2.

Issue #313, opened 2021-11-26
improper use of == in configure


Hello71:
https://github.com/libunwind/libunwind/commit/563b0ff3144c0e60a9aa63fdafb4dea66ecd3d50, https://github.com/libunwind/libunwind/commit/7b498c21029effd85dc1d92cc93ada1d6c373dea

test a == b is not implemented in dash; one equals sign must be used. it causes `./configure: 18607: test: xx86_64: unexpected operator` and `./configure: 19002: test: xno: unexpected operator` and the tests will always return false. there is also no need to use the test command or x workaround; all modern shells support the more readable `if [ "$remote_only" = yes ]` syntax. it doesn't hurt anything but readability though, whereas using == will make the test always return false on Debian-derived systems.

PR #315, opened 2021-11-27
configure.ac: avoid bashisms

thesamesam:
configure scripts need to be written in POSIX-compliant shell; using
'==' is a Bashism (doesn't work for POSIX-compliant shells). Let's
use '=' as we do elsewhere to fix compatibility.

Fixes: #313
Signed-off-by: Sam James <sam@gentoo.org>


djwatson:
Thanks.  Does this need to go in 1.6, or only master?

thesamesam:
> Thanks. Does this need to go in 1.6, or only master?

Thanks for merging! 1.6 needs the fix too.

PR #317, opened 2021-12-01
Skip init_mem_validate calls during remote unwind

am11:
This was failing linux arm64 cross build on windows with remote unwinding as `tdep_init_mem_validate` is not defined when `UNW_REMOTE_ONLY` is.

The fix is to simply follow x86_64/Gglobal.c approach:
https://github.com/libunwind/libunwind/blob/4c01eb4107a0cc69bd35a5f3ed70bfb754fdc311/src/x86_64/Gglobal.c#L96-L99

### DIFF
--- a/src/aarch64/Gglobal.c
+++ b/src/aarch64/Gglobal.c
@@ -47,9 +47,9 @@ tdep_init (void)
 
     dwarf_init ();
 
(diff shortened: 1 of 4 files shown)


djwatson:
Thanks!

PR #320, opened 2021-12-15
DWARF: avoid a data race

compnerd:
This race condition was identified by TSAN.  The following is a
truncated report from TSAN:

~~~
Write of size 1 at ...
  #0 rs_new libunwind/src/dwarf/Gparser.c
  #1 find_reg_state libunwind/src/dwarf/Gparser.c
  ...

Previous read of size 1 at ...
  #0 find_reg_state libunwind/src/dwarf/Gparser.c
  ...
~~~

The cache is accessed via `get_rs_cache` which will acquire the lock,
and then at the end of the second `if (cache)` block, will release the
lock via the `put_rs_cache`.  We subsequently will read the cache entry
for the parameter for `tdep_reuse_frame`.

Rather than hoisting the `tdep_reuse_frame`, we could re-acquire the
lock, however, this seems unnecessary and would increase lock contention
for just the error path handling.

Thanks to @wmi0 for help with this issue.

### DIFF
--- a/src/dwarf/Gparser.c
+++ b/src/dwarf/Gparser.c
@@ -960,13 +960,11 @@ find_reg_state (struct dwarf_cursor *c, dwarf_state_record_t *sr)
 	  cache->links[c->prev_rs].hint = index + 1;
 	  c->prev_rs = index;
 	}
+      if (ret >= 0)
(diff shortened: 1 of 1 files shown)


djwatson:
Thanks!

PR #330, opened 2022-02-18
Updated to determine PAGE_SIZE dynamically.

### LINKED ISSUE #328: libunwind hardcodes PAGE_SIZE
In #212, I had started to take a look at libunwind on various arm platforms, and I found that libunwind was hardcoding a standard PAGE_SIZE, which was not working on one of my systems which had a different page size.

I found I could determine the page size dynamically at runtime using `sysconf(_SC_PAGESIZE)` via  [sysconf](https://man7.org/linux/man-pages/man3/sysconf.3.html), which was working across my various test platforms. 

I have a [small dev patch]( https://github.com/mongodb-forks/libunwind/commit/8e357e280cf5b2bd76520c5b8a96b8d8599986f8) I was using, but I would like the maintainer to take a look and fix this issue as they see fit. Since the arm code was ported from x86, I think this affects several other platforms.

dmoody256:
fixes #328

### DIFF
--- a/include/libunwind_i.h
+++ b/include/libunwind_i.h
@@ -55,6 +55,8 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 #include <string.h>
 #include <unistd.h>
 #include <sys/mman.h>
+#include <errno.h>
+#include <stdio.h>
 
 #if defined(HAVE_ELF_H)
 # include <elf.h>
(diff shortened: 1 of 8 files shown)


tpwrules:
I built libunwind on an AArch64 system with 16K pages and various dependents worked (Firefox, Xorg, Gstreamer, ...), unlike before where they would crash on launch. I then used the same binaries but with a 4K kernel and they still worked properly. Thanks for this work.

It may not be worth dynamically determining the page size for x86, x86_64, and ARM 32 bit (and possibly other architectures, but definitely not PPC and AArch64). As far as I know, these are fixed to have 4K pages in hardware. On the other hand, the consistency is nice.

dmoody256:
ping @djwatson any comments? We are planning to move forward with this in a fork and would like not to have to undo anything if this is not good solution.

djwatson:
This looks reasonable to me.  

Note x86_64 has multiple page sizes (hugepages), but as of yet no one has complained...

PR #332, opened 2022-02-26
Make some structs non-empty

giordano:
GCC shows warnings like this when compiling libraries or programs linking to
libunwind:

```
/buildworker/worker/package_linuxaarch64/build/usr/include/libunwind-aarch64.h:60:9: warning: empty struct has size 0 in C, size 1 in C++ [-Wc++-compat]
 typedef struct
         ^~~~~~
/buildworker/worker/package_linuxaarch64/build/usr/include/libunwind-aarch64.h:169:16: warning: empty struct has size 0 in C, size 1 in C++ [-Wc++-compat]
 typedef struct unw_tdep_save_loc
                ^~~~~~~~~~~~~~~~~
```

The pattern of using a dummy field is already used in other platforms.

CC: @vtjnash

### DIFF
--- a/include/libunwind-x86.h
+++ b/include/libunwind-x86.h
@@ -158,6 +158,7 @@ x86_regnum_t;
 typedef struct unw_tdep_save_loc
   {
     /* Additional target-dependent info on a save location.  */
+    char unused;
   }
 unw_tdep_save_loc_t;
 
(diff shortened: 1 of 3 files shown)


include/libunwind-x86.h
```diff
@@ -158,6 +158,7 @@ x86_regnum_t;
 typedef struct unw_tdep_save_loc
   {
     /* Additional target-dependent info on a save location.  */
+    char unused;
```
BillyDonahue, line 161: Maybe better for consistency between arch variants to use the same `UNW_EMPTY_STRUCT` idiom as was used in libunwind-aarch64.h and libunwind-arm.h ?

giordano: Done 👍

PR #334, opened 2022-03-09
FreeBSD also uses these structs so do not hide them

brd:


brd:
Found in the build failure here: http://ampere2.nyi.freebsd.org/data/main-arm64-default/p422f190aeba4_s23210c9f42/logs/errors/libunwind-20211201.log

MikaelUrankar:
Does unw_sigcontext has to match the FreeBSD aarch64 ucontext_t? We use a different ucontext_t than linux.

PR #348, opened 2022-03-31
DWARF: avoid invalid memory access with invalid CFI

compnerd:
In the case that the CFI is incorrect, the return address column entry
may be incorrect and point outside of the range of the program.  Add a
cheap validation to prevent the errant memory access.


djwatson:
Nice catch, thanks.

PR #354, opened 2022-04-26
avoid leaking registers across sigreturn

vtjnash:
Mostly just relevant for fp registers, which are frequently mostly just
ignored otherwise. So I don't think this should change anything noticeable normally, but only for a very unusual unwind.


djwatson:
Seems reasonable to me, thanks.

PR #355, opened 2022-04-26
restore FP registers across signals on aarch64

vtjnash:
It looked like this might be lacking, but straightforward (though likely not common to need
them to be available). Most other platforms also do not restore all of the registers, particularly since instructions like VFP and XSAVE have variable layouts based on the CPU or are undocumented.

### DIFF
--- a/src/aarch64/Gstep.c
+++ b/src/aarch64/Gstep.c
@@ -122,6 +122,8 @@ aarch64_handle_signal_frame (unw_cursor_t *cursor)
   c->dwarf.loc[UNW_AARCH64_SP]  = DWARF_LOC (sc_addr + LINUX_SC_SP_OFF, 0);
   c->dwarf.loc[UNW_AARCH64_PC]  = DWARF_LOC (sc_addr + LINUX_SC_PC_OFF, 0);
   c->dwarf.loc[UNW_AARCH64_PSTATE]  = DWARF_LOC (sc_addr + LINUX_SC_PSTATE_OFF, 0);
+  for (i = 0; i < 32; ++i)
+    c->dwarf.loc[UNW_AARCH64_V0 + i] = DWARF_LOC (sc_addr + LINUX_SC_V0_OFF + i * 0x10, 0);
 
   /* Set SP/CFA and PC/IP.  */
   dwarf_get (&c->dwarf, c->dwarf.loc[UNW_AARCH64_SP], &c->dwarf.cfa);
--- a/src/aarch64/offsets.h
+++ b/src/aarch64/offsets.h
@@ -47,3 +47,45 @@
 #define LINUX_SC_SP_OFF         0x100
 #define LINUX_SC_PC_OFF         0x108
 #define LINUX_SC_PSTATE_OFF     0x110
+#define LINUX_SC_RESERVED_OFF   0x120
+/* { u32 magic; u32 size; u32 fpsr; u32 fpcr; uint128 vregs[]; } */
+#define LINUX_SC_V0_OFF         (LINUX_SC_RESERVED_OFF + 0x010)
+#define LINUX_SC_V1_OFF         (LINUX_SC_RESERVED_OFF + 0x020)
+#define LINUX_SC_V2_OFF         (LINUX_SC_RESERVED_OFF + 0x030)
+#define LINUX_SC_V3_OFF         (LINUX_SC_RESERVED_OFF + 0x040)
+#define LINUX_SC_V4_OFF         (LINUX_SC_RESERVED_OFF + 0x050)
+#define LINUX_SC_V5_OFF         (LINUX_SC_RESERVED_OFF + 0x060)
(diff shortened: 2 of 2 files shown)


djwatson:
Looks like a merge conflict with https://github.com/libunwind/libunwind/pull/354.

Please update with whatever you think the best fix is, thanks.

src/aarch64/Gstep.c
```diff
@@ -119,6 +119,8 @@ aarch64_handle_signal_frame (unw_cursor_t *cursor)
   c->dwarf.loc[UNW_AARCH64_SP]  = DWARF_LOC (sc_addr + LINUX_SC_SP_OFF, 0);
   c->dwarf.loc[UNW_AARCH64_PC]  = DWARF_LOC (sc_addr + LINUX_SC_PC_OFF, 0);
   c->dwarf.loc[UNW_AARCH64_PSTATE]  = DWARF_LOC (sc_addr + LINUX_SC_PSTATE_OFF, 0);
+  for (reg = 0; reg < 32; ++reg)
+    c->dwarf.loc[UNW_AARCH64_V0 + reg] = DWARF_LOC (&c->dwarf, LINUX_SC_V0_OFF + reg * 0x10);
```
James-A-Clark, line 123: Do you have a test or an example that shows this working? I have a suspicion that it might not because of my previos comment and the default dwarf_get() function is for 64 bit values and these V registers are 128 bit.

James-A-Clark (changes requested):

src/aarch64/Gstep.c
vtjnash: This seems to go through `dwarf_getfp`, which is 128-bit values (unw_fpreg_t is a long double). This PR is segfaulting in that method though on a few tests (Ltest-resume-sig), so I see this PR is very much not working correctly though.

vtjnash: That test now passes for me, and it attempts to do a bit of FP state manipulation (but not these vectors). I am not sure of a particularly direct or easy way to check that it is getting it correct.

James-A-Clark: Thanks I see that the latest change looks better with the new macro. I just worry that without a before and after of something working vs not working that we might be adding something that is subtly incorrect that will just need to be fixed up later.

One other issue is that the floating point registers are only saved in the signal context if system_supports_fpsimd() is true in the kernel. To be honest I'm not sure if that's an issue in practice as these would only be accessed in the case where it was enabled anyway.

I actually have a very similar change that I'm just about to post in the next couple of days that adds support for SVE unwinding where I check for the existence of the SVE section before assigning the locations. Perhaps you could use the same search function to search for the FP section magic number, although as I said maybe in practice we can be certain that it's always there.

vtjnash: Yeah, that is fair, and I don't actually have a use-case for this, just an observation. Do you want to just take this over then, since it is likely just a sub-case of your vfp work?

Looking at `src/aarch64/Gresume.c`, we can actually be reasonably sure we don't need to care about these in most typical cases for `unw_resume` at least. In particular, normally for `unw_resume` it only restores the non-scratch register state. While for signals, it will implicitly inherit all unlisted state from signal. So as long as none of these registers are considered callee-saved, there shouldn't be an issue for code following the normal C ABI.

There is a slight hiccup with that claim, which is that `unw_setcontext` is currently a macro (rather than a proper assembly function), and therefore does not correctly force it to use the platform ABI. But that is tangential to this change.

James-A-Clark: It's been a while because I was waiting for the kernel side changes to be merged, but here's the change I was talking about: https://github.com/libunwind/libunwind/pull/423/commits/01e093b09d521d1432c40972e5533c19f69e18f3#diff-d8a7e9d46e2dc8da2f5bef4f1ba9ae27cfdecc24e8035d294b4f2783c26c523fR84

See "switch (magic)", you could add the FP magic number in that switch and then add the dwarf locs if it's found.

djwatson:
Closing due to  merge comments and broken tests and inactivity.  Please reopen as appropriate.

vtjnash:
I can rebase, but I thought James-A-Clark was going to make a PR with an even more complete version of this (https://github.com/libunwind/libunwind/pull/355#discussion_r881458733), so I stopped working on it.

vtjnash:
Thus, as far as I know, this is now fixed by https://github.com/libunwind/libunwind/pull/423

James-A-Clark:
> Thus, as far as I know, this is now fixed by #423

See the comment here: https://github.com/libunwind/libunwind/pull/355#discussion_r983384960

I added the switch statement which would be a good place to put a case for the FP registers, but I didn't actually make the change.

vtjnash:
Gotcha, this would be easy for someone to implement then given the magic number 0x46508001 from https://github.com/torvalds/linux/blob/master/arch/arm64/include/uapi/asm/sigcontext.h#L75

PR #356, opened 2022-04-26
move CI to Github Actions

vtjnash:
libunwind is out of free TravisCI credits now.


vtjnash:
This is a fairly direct port from the Travis script, plus I took some parts of https://github.com/libuv/libuv/blob/v1.x/.github/workflows/CI-unix.yml

I don't know why the tests fail, but they also seem to fail for me locally, so possibly this is some sort of configuration issue with `run-coredump-unwind` that I am unaware of.

djwatson:
Thanks for doing this!

I have merged this, but without deleting .travis.yml yet.  I have also got more free travis credits.  Might as well run both until all the kinks are worked out.

PR #357, opened 2022-04-26
move Windows CI to Github Actions

vtjnash:
libunwind is out of free TravisCI credits now.

### DIFF
--- a/.github/workflows/CI-win.yml
+++ b/.github/workflows/CI-win.yml
@@ -1,0 +1,37 @@
+name: CI-win
+
+on:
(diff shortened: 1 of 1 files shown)


vtjnash:
Fairly direct translation from the old scripts. Note that this fails because of a bug introduced recently (e85b65cec757ef589f28957d0c6c21c498a03bdf)

djwatson:
Merged, thanks.

I don't see any comments/discussion/revert of e85b65c anywhere, what's the issue?

Issue #352, opened 2022-04-25
unw_init_remote can't unwind full backtrace[lack some callchains]


footka:
hi guys! I try to collect backtrace of another process, the strategy  used is ptrace + libunwind(remote mode), everything is ok under linux x86_64, but unexpected results appears under aarch64, details as follows:

env:
kernel: Linux i85c06207.eu95sqa 4.19.91
libunwind: 1.6.2
gcc： 5.20
link mode: static

test program(test.c):
```
#include<stdio.h>
#include<stdlib.h>
#include<unistd.h>

void foo(int i) {
  if (i==0) {sleep(100); return;};
  foo(i-1);
}

int main()
{
  foo(10);
}
```

compile command:
```
gcc test.c 
```

unwind program (code fragments):
```       
 rc = ptrace(PTRACE_ATTACH, t->tid_);
        if (-1 == rc) {
          if (errno != ESRCH) {
            LOG(WARN, "ptrace attach failed, tid: %d, err: %d, errmsg: %s",
                t->tid_, errno, strerror(errno));
          }
          continue;
        }
        rc = -1;
        int wait_loops = 10;
        while (wait_loops-- > 0) {
          int st = 0;
          int w_pid = wait4(t->tid_, &st, __WALL, NULL);
          if (-1 == w_pid) {
            rc = errno;
            LOG(WARN, "wait failed, err: %d, errmsg: %s", errno, strerror(errno));
            break;
          } else if (WIFSTOPPED(st)) {
            LOG(INFO, "stopped........");
            rc = 0;
            break;
          }
          usleep(50);
        }
        if (rc != 0) {
          LOG(ERROR, "wait failed");
          continue;
        }

        struct UPT_info *ui = (struct UPT_info *)_UPT_create(t->tid_);
        if (!ui) {
          LOG(WARN, "unw_create_addr_space failed");
          rc = -1;
          continue;
        }
        unw_cursor_t c;
        unw_init_remote(&c, as, ui);
        t->n_addrs_ = 0;
        int f_limit = min((int)ARRAYSIZE(t->addrs_), CONF.frame_limits);
        do {
          unw_word_t uip;
          if ((rc = unw_get_reg(&c, UNW_REG_IP, &uip)) < 0) {
            LOG(WARN, "get reg failed, err: %d\n", rc);
            break;
          }
          unw_word_t sp;
          unw_get_reg(&c, UNW_REG_SP, &sp);
          t->addrs_[t->n_addrs_] = uip;
[...]

PR #360, opened 2022-05-09
Unwind with pointer authentication on arm64

andrewkilroyarm:
This patch adds the ability for libunwind to unwind a stack where the
return address obtained from the arm64 link register (x30) has
a pointer authentication code (PAC).

Without this patch, the unw_step function terminates early, leaving the
user the impression there is only a leaf frame when there are more.

The reason for the premature termination is that an aarch64 specific
CFI opcode 'DW_CFA_AARCH64_negate_ra_state' is not recognised, and
treated as unexpected.  Upon the next iteration's call of unw_step, it
reports the end of the stack.

For an expected callstack of

  loop   // most recent call
  call3
  call2
  call1
  main   // oldest call

The caller of libunwind observes:

  unw_step() = 1
  unw_get_reg(UNW_REG_IP) = 0, pc = 0xaaaae4720798
  unw_step() = 1
  unw_get_reg(UNW_REG_IP) = 0, pc = 0x5faaaae47207cc // ip with PAC
  unw_step() = 0  // premature end

For remote unwinding, this patch adds an optional callback function
'ptrauth_insn_mask' on unw_accessors_t so that unw_step can correctly
strip off the PAC to leave the correct return address.  This also has
the benefit that an attempt to get the name of the function with
unw_get_proc_name can succeed.

The callback is given the original unw_addr_space_t and as_arg as given
to unw_init_remote.  The application needs to ensure the mask to be
returned is somehow available on the as_arg and callback should return
it.

The callback is only needed for remote unwinding.  In local unwinding,
an instruction is used to strip the PAC.

A description of pointer authentication is in the armv8 reference manual
available here:

  https://developer.arm.com/documentation/ddi0487/ha/?lang=en

(Version H.a, section D5.1.5 Pointer authentication in AArch64 state,
page D5-4772)

Also documentation for the use of DWARF with pointer authentication, in
particular the RA_SIGN_STATE:

  https://github.com/ARM-software/abi-aa/blob/main/aadwarf64/aadwarf64.rst

### DIFF
--- a/src/dwarf/Gparser.c
+++ b/src/dwarf/Gparser.c
@@ -87,6 +87,13 @@ empty_rstate_stack(dwarf_stackable_reg_state_t **rs_stack)
     pop_rstate_stack(rs_stack);
 }
 
+#ifdef UNW_TARGET_AARCH64
+
+static void
+aarch64_negate_ra_sign_state(dwarf_state_record_t *sr);
+
+#endif
+
 /* Run a CFI program to update the register state.  */
 static int
 run_cfi_program (struct dwarf_cursor *c, dwarf_state_record_t *sr,
@@ -403,6 +410,11 @@ run_cfi_program (struct dwarf_cursor *c, dwarf_state_record_t *sr,
             set_reg (sr, regnum, DWARF_WHERE_CFAREL,
                      (regnum - 16) * sizeof (unw_word_t));
           Debug (15, "CFA_GNU_window_save\n");
+          break;
+#elif UNW_TARGET_AARCH64
+          /* This is a specific opcode on aarch64, DW_CFA_AARCH64_negate_ra_state */
+          Debug (15, "DW_CFA_AARCH64_negate_ra_state\n");
+          aarch64_negate_ra_sign_state(sr);
           break;
 #else
           /* FALL THROUGH */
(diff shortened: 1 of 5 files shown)


footka:
good job! can this patch fix issue #352 ? I had pull  github-ptrauth branch for test and verify, it seems to have no work. wish your reply!

footka:
sorry, i read detailed info， this patch just fix arm64 issue. my fault!

djwatson:
Thanks!

src/dwarf/Gparser.c
```diff
@@ -767,6 +779,61 @@ eval_location_expr (struct dwarf_cursor *c, unw_word_t stack_val, unw_addr_space
...
+
+static unw_word_t
+aarch64_strip_pac_local(unw_word_t in_addr)
+{
+  unw_word_t out_addr = in_addr;
+
+#ifdef __aarch64__
+  // Strip the PAC with XPACLRI instruction
+  register unsigned long long x30 __asm__("x30") = in_addr;
+  __asm__("hint 0x7" : "+r" (x30));
```
vtjnash, line 822: I think this `__asm__` is missing some extra `: :` that might be required by the Microsoft compiler (observed in all CI)?

andrewkilroyarm: Taking a look

PR #365, opened 2022-05-24
Add remote unwinding support for macOS

am11:
This patch adds remote unwinding support for macOS, similar to the one added earlier for Windows.
It is essentially moving files from https://github.com/dotnet/runtime/tree/9859f70/src/native/external/libunwind_extras/mac and sharing the headers with win support.

Also fixes the CI issues on Windows: it required C11 to be enabled in CMake and exclude `__asm__` block in dwarf/Gparser.c for win-arm64 (inline assembly is not supported on this platform).


am11:
cc @AaronRobinsonMSFT, @janvorli

am11:
`CI-win` workflow is passing (failed on master branch). Rest of the failures are unrelated to PR changes (also failed on master).

CMakeLists.txt
```diff
@@ -2,6 +2,8 @@ project(libunwind)
 
 cmake_minimum_required(VERSION 3.16.1)
 
+set(CMAKE_C_STANDARD 11)
```
AaronRobinsonMSFT, line 5: This seems like a big policy/support move. Is there a process in libunwind for this?

AaronRobinsonMSFT: Perhaps rev the package version?

am11: This cmake is only used by the windows CI. everything else uses autoconf build system. This line fixes the CI build on Windows.

We are using C11 since https://github.com/libunwind/libunwind/commit/45fc222a00f239a16f1420c4446113e8a730b75a.

am11:
Thanks for pointing it out, @bregma. I have updated/aligned the license header in a few files.

am11:
@bregma could you merge this too?

PR #373, opened 2022-06-22
added coredump support for s390x

tkoecker:
This change adds s390x coredump support, it contains the following:
- It adds reading the NT_FPREGSET notes, as the floating point registers are needed for unwinding on s390x.
- The fpregset is added to the thread data structure.
- The getter for floating point registers was implemented (at least for s390x).
- A register mapping for the registers in NT_PRSTATUS was added, as the register indices of the enum in libunwind are different to those in prstatus.

### DIFF
--- a/src/coredump/_UPT_access_fpreg.c
+++ b/src/coredump/_UPT_access_fpreg.c
@@ -38,7 +38,10 @@ _UCD_access_fpreg (unw_addr_space_t as, unw_regnum_t reg, unw_fpreg_t *val,
 
 #ifdef __s390x__
   if (reg < UNW_S390X_F0 || reg > UNW_S390X_F15)
-    goto badreg;
+    {
+      Debug(0, "bad regnum:%d\n", reg);
+      return -UNW_EINVAL;
+    }
 
   *val = ui->fpregset->fprs[reg - UNW_S390X_F0].d;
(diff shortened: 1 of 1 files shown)


src/coredump/_UPT_access_fpreg.c
```diff
@@ -28,7 +28,27 @@ int
...
+
+  if (write)
+    {
+      Debug(0, "write is not supported\n");
+      return -UNW_EINVAL;
+    }
+
+#ifdef __s390x__
+  if (reg < UNW_S390X_F0 || reg > UNW_S390X_F15)
+    goto badreg;
```
bregma, line 41: I think it would be more consistent with the existing codebase to just have the Debug() call and return -UNW_EINVAL here instead of jumping out of the \#ifdef to code that can only be reached from this line.

bregma (changes requested):

src/coredump/_UPT_access_fpreg.c
tkoecker: I tried to follow the way this is implemented in _UCD_access_reg() where there is a goto badreg; as error handling for each architecture. I however have to admit that it does not make much sense if only one architecture is handled. I've updated the change and moved the error handling.

PR #380, opened 2022-07-06
Autodetect presence of elf_fpregset_t

### LINKED ISSUE #378: PR #373 causes FTBFS on some target OSes
PR #373 unconditionally uses the Linux construct `elf_fpregset_t` which causes a failure to build from source on target OSes that do not have that construct. Its use should be autodetected and optioned out.

bregma:
Data of this type was added for the s390x port doesn't exists on all OSes. It
needs to be autodetected and optioned out.

Fixes #378


bregma:
Not sure what the new s390x breakage is all about but it's ocurring on other CI builds and this is blocking the borken FreeBSD and QNX builds so I'm going to go ahead and commit to unblock them.

PR #382, opened 2022-07-08
Add support for arm fp registers unwind

am11:
Upstream of this patch https://github.com/dotnet/runtime/commit/1b5719c2e3dde393531eaeb5b5cde05dabeef5b8 (shipped in .NET 6 last year and tested in `main` branch since). It was readjusted on libunwind v1.6.2 in https://github.com/dotnet/runtime/commit/231b4568059c6a6bf2819d50b99883b666de6fe7.

### DIFF
--- a/include/libunwind-arm.h
+++ b/include/libunwind-arm.h
@@ -261,6 +261,7 @@ unw_tdep_save_loc_t;
 typedef struct unw_tdep_context
   {
     unsigned long regs[16];
+    unsigned long long fpregs[16];
   }
 unw_tdep_context_t;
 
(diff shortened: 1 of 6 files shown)


am11:
@janvorli, @vtjnash, @AaronRobinsonMSFT, PTAL. I have simply cherry-picked those changes from main and committed on top of libunwind's master branch. This is the last patch which wasn't present in upstream.

AaronRobinsonMSFT:
@am11 are the failures here a real concern?

am11:
AFAIK, there is no additional failure. The baseline (master branch) has the same amount of failures.

bregma:
The failures are not due to this patch but I'm trying to get the Travis-CI build green on master before anything else goes in.

vtjnash:
The ARM CI on GA failed with a build error:
```
/tmp/ccCLi7Sa.s:157: Warning: if writeback register is in list, it must be the lowest reg in the list
[53](https://github.com/libunwind/libunwind/runs/7541287933?check_suite_focus=true#step:6:54)
/tmp/ccCLi7Sa.s:158: Error: selected processor does not support `vstmia r1,{d0-d15}' in ARM mode
[54](https://github.com/libunwind/libunwind/runs/7541287933?check_suite_focus=true#step:6:55)
```

am11:
Does that mean we need to handle `-mfpu=none` / `#ifdef __TARGET_FPU_NONE` case?

PR #397, opened 2022-07-31
Fix typos

am11:
Found these via [`typos`](https://github.com/crate-ci/typos) utility, while fixing typos in dotnet/runtime repo (where we have copy of libunwind tree).

Second commit cleans up trailing whitespaces in files modified by first commit.

### DIFF
--- a/include/tdep-ia64/libunwind_i.h
+++ b/include/tdep-ia64/libunwind_i.h
@@ -152,7 +152,7 @@ struct cursor
     unsigned int pi_is_dynamic :1; /* proc_info found via dynamic proc info? */
     unw_proc_info_t pi;         /* info about current procedure */
 
-    /* In case of stack-discontinuities, such as those introduced by
+    /* In case of stack discontiguities, such as those introduced by
        signal-delivery on an alternate signal-stack (see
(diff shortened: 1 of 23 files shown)


include/tdep-ia64/libunwind_i.h
```diff
@@ -152,7 +152,7 @@ struct cursor
     unsigned int pi_is_dynamic :1; /* proc_info found via dynamic proc info? */
     unw_proc_info_t pi;         /* info about current procedure */
 
-    /* In case of stack-discontiguities, such as those introduced by
```
bregma, line 155: This is not a spelling correction but a word choice change with subtle differences. I would keep the original word choice since it's more correct but remove the hyphen since hyphenated compound nouns are adjectives not objects.

bregma (changes requested):

include/tdep-ia64/libunwind_i.h
```diff
@@ -152,7 +152,7 @@ struct cursor
...
+    /* In case of stack-discontinuities, such as those introduced by
```
am11, line 155:
```suggestion
    /* In case of stack discontiguities, such as those introduced by
```

PR #400, opened 2022-08-05
Fix __SOFTFP__ case for arm getcontext

am11:
See https://github.com/libunwind/libunwind/pull/382#issuecomment-1204408556 for context.

### DIFF
--- a/src/dwarf/Gfind_proc_info-lsb.c
+++ b/src/dwarf/Gfind_proc_info-lsb.c
@@ -135,9 +135,9 @@ load_debug_frame (const char *file, char **buf, size_t *bufsize, int is_local,
 #if defined(SHF_COMPRESSED)
   if (shdr->sh_flags & SHF_COMPRESSED)
     {
-      unsigned long destSize;
       Elf_W (Chdr) *chdr = (shdr->sh_offset + ei.image);
 #ifdef HAVE_ZLIB
+      unsigned long destSize;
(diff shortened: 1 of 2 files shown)


am11:
cc @vtjnash, @janvorli

src/dwarf/Gfind_proc_info-lsb.c
```diff
@@ -135,9 +135,9 @@ load_debug_frame (const char *file, char **buf, size_t *bufsize, int is_local,
 #if defined(SHF_COMPRESSED)
   if (shdr->sh_flags & SHF_COMPRESSED)
     {
-      unsigned long destSize;
       Elf_W (Chdr) *chdr = (shdr->sh_offset + ei.image);
 #ifdef HAVE_ZLIB
+      unsigned long destSize;
```
am11, line 140: Unrelated change: gcc was warning that destSize is unused (it is unused outside of HAVE_ZLIB branch).

am11:
FWIW,  I compared macros using:
`echo | arm-linux-gnueabi-gcc -dM -E - | grep -iE 'arm|fpu|thumb|neon|soft'`
vs.
`echo | arm-linux-gnueabihf-gcc -dM -E - | grep -iE 'arm|fpu|thumb|neon|soft'`

vtjnash (commented):
SGTM

PR #402, opened 2022-08-05
Cross-build CI `make check` QEMU tweak

### LINKED ISSUE #401: Github cross-build actions try to run `make` on target
This cross-compile CI actions on github are trying to run `make check` using qemu. They should be running `make check` on the build host and run the *tests* using qemu.

bregma:
The cross-built CI runs using GiotHub actions were using QEMU to run `make
check`. Instead, the host should run `make check` and the actual test binary
should be run under QEMU.

This change configures the cross build to use a (new) qemu-test-driver script as
the test harness, which in turn invokes the test under QEMU.  The tests are
linked to the libunwind libraries statically to avoid complicating the search
paths and invoking libtool.

Fixes #401.


.github/workflows/CI-linux.yml
```diff
@@ -121,4 +121,9 @@ jobs:
           set -x
           sudo bash -c 'echo core.%p.%p > /proc/sys/kernel/core_pattern'
           ulimit -c unlimited
-          ${{ matrix.config.qemu }} make -j8 check
+          CROSS_LIB="/usr/${{ matrix.config.host }}"
+          make -j8 check LOG_DRIVER_FLAGS="--qemu-arch ${{ matrix.config.qemu }}" LDFLAGS="-L$CROSS_LIB/lib -static" QEMU_LD_PREFIX="$CROSS_LIB"
```
vtjnash, line 125: Thanks! It looks good to have more CI coverage now.

In very odd coincidental timing, I was also just thinking back today about how to fix this, so I found today that you could do here the same as this with the existing `libtool` script:
```suggestion
          LOG_COMPILER='../libtool --mode=execute qemu-${{ matrix.config.target }}'
          TESTS_ENVIRONMENT='LD_LIBRARY_PATH=/usr/${{ matrix.config.target }}-linux-gnu/lib'
          make -j8 check
```

This it might also need to explicitly install the expected shared libraries earlier (in the install QEMU step)
```
          sudo apt install qemu-user ${{ matrix.config.toolchain }} ${{ matrix.config.libc }} -y
          sudo ln -s /usr/${{ matrix.config.host }}/lib/ld-linux-${{ matrix.config.target }}.so.1 /lib/
```
given the new variable matrix.config.libc

bregma: The libc and dynamic loader are installed as dependencies of the qemu-user package so there's no need to explicitly install them. 

Yes, it's possible to use LOG_COMPILER+libtool instead, for tests that are simple binary executables, but we already had a hook for replacing LOG_DRIVER which is required for running tests on remote targets (like for embedded OSes) and I was familiar with it. Six of one half a dozen of the other but a customer LOG_DRIVER script will be a more elegant solution to issue #403 in the future.

PR #408, opened 2022-08-12
Added support for unwinding from the vsyscall region

kent-cheung-arm:
This change fixes unwinding from the vsyscall region. Although vsyscall has been phased out, it is still possible to call into it at an address where libunwind is unable to step out of. Here is a reproducer
```
#include <stdio.h>
#include <sys/time.h>
#include <asm/vsyscall.h>

int (*f)(struct timeval *tv, struct timezone *tz) = (void*) VSYSCALL_ADDR; 

int main() {

    struct timeval current_time;
    f(&current_time, NULL);
    printf("seconds : %ld\nmicro seconds : %ld\n",
           current_time.tv_sec, current_time.tv_usec);

    return 0;
}
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -34,10 +34,10 @@ esac
 
 dnl Checks for header files.
 AC_HEADER_STDC
-AC_CHECK_HEADERS(asm/ptrace_offsets.h asm/ptrace.h endian.h sys/endian.h sys/param.h \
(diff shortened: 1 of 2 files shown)


bregma:
Looks like this needs to be conditional on the autodetection of `<asm/vsyscall.h>` since that header is specific to the Linux kernel.

kent-cheung-arm:
Thanks for the assistance with PR.

PR #418, opened 2022-09-07
tests/run-coredump-unwind: Skip test if no coredump has been created

fweimer-rh:
In some build environments, coredumps are not created even if the corresponding ulimit is positive. This change skips the test if the coredump is missing.

### DIFF
--- a/tests/run-coredump-unwind
+++ b/tests/run-coredump-unwind
@@ -49,6 +49,10 @@ fi
     ./crasher backing_files
 ) 2>/dev/null
 COREFILE=$TEMPDIR/core*
+if ! test -f "$COREFILE"; then
+    echo "crasher process did not produce coredump, test skipped"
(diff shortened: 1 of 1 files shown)


tangyouling:
In the beginning, I also had this issue. It is not that the core file is not generated,
but the system configuration has changed the file path. You can try the following
modifications to see if it can pass the test.

Default settings (take my environment as an example):
$ cat /proc/sys/kernel/core_pattern
|/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h

Change setting:
# echo "core" > /proc/sys/kernel/core_pattern

After modification, run-coredump-unwind and run-coredump-unwind-mdi can pass the test.

fweimer-rh:
@tangyouling This only works if you can change the builders at such a fundamental level. Usually that's not possible.

PR #423, opened 2022-09-29
arm64: Add support for unwinding in SVE functions

James-A-Clark:
This PR allows remote and signal frame unwinding though an SVE function that pushes SVE registers onto the stack[1].

The remote unwinder is responsible for providing the value of the VG register at the time the sample was taken. Or in a signal frame the kernel has already placed it into the stack. Plain local unwinding is not supported yet and a warning has been added for that scenario. It's not trivial to add the current SVE vector width to the get_context() function, but it could be added later if needed.

A test has been added that demonstrates signal frame unwinding, and support has already been upstreamed to Perf which uses remote unwinding.

[1]: https://github.com/ARM-software/abi-aa/blob/main/aadwarf64/aadwarf64.rst


src/aarch64/Gregs.c
```diff
@@ -92,6 +93,41 @@ tdep_access_reg (struct cursor *c, unw_regnum_t reg, unw_word_t *valp,
...
+         Debug (1, "Writing to VG pseudo register not supported\n");
+         return -UNW_EBADREG;
+       }
+
+      if (DWARF_IS_REG_LOC(c->dwarf.loc[reg]))
+        loc = c->dwarf.loc[reg];
+      else
+        {
+          unw_addr_space_t as = c->dwarf.as;
+          unw_accessors_t *a = unw_get_accessors_int (as);;
```
bregma, line 108: Extraneous semicolon

James-A-Clark: Oops thanks will fix!

src/aarch64/init.h
```diff
@@ -30,6 +30,9 @@ common_init (struct cursor *c, unsigned use_prev_instr)
 {
   int ret, i;
 
+  for (i = UNW_AARCH64_PSTATE + 1; i < UNW_AARCH64_V0; ++i)
```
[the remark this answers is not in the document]
James-A-Clark: UNW_AARCH64_VG falls in that range so the ones that are set are sparse. It would be a pain if further ones are added to keep that loop up to date so I just moved it to the beginning so everything is zeroed before being set.

src/aarch64/Gregs.c
James-A-Clark: Done, thanks for the review

kent-cheung-arm:
LGTM

Issue #425, opened 2022-10-08
_ULx86_64_tdep_trace crash: access `cache->frames[slot]` after `cache->frames` has been unmapped 


hidva:
libunwind 1.6.2

# Reproduce

1.  je_malloc_default --> imalloc --> je_prof_backtrace --> unw_backtrace --> _ULx86_64_tdep_trace

2. Receive a SIGUSR2 before executing `if (unlikely(! f))`.

    https://github.com/libunwind/libunwind/blob/b3ca1b59a795a617877c01fe5d299ab7a07ff29d/src/x86_64/Gtrace.c#L449-L452

3. call unw_backtrace in SIGUSR2 handler. unw_backtrace --> _ULx86_64_tdep_trace --> trace_lookup --> trace_cache_expand --> `munmap(cache->frames, old_size * sizeof(unw_tdep_frame_t))`.
4. Continue to execute `if (unlikely(! f))` after the signal handler returns. Because of `munmap(cache->frames)` in step 3, accessing `f->last_frame` will trigger a SIGSEGV.

hidva:
We used the following patch as a workaround.
```diff
diff --git a/src/x86_64/Gtrace.c b/src/x86_64/Gtrace.c
index 963a858f..6ea7df9e 100644
--- a/src/x86_64/Gtrace.c
+++ b/src/x86_64/Gtrace.c
@@ -41,8 +41,9 @@ typedef struct
   unw_tdep_frame_t *frames;
   size_t log_size;
   size_t used;
-  size_t dtor_count;  /* Counts how many times our destructor has already
+  uint32_t dtor_count;  /* Counts how many times our destructor has already
                          been called. */
+  uint32_t frame_rc;
 } unw_trace_cache_t;

 static const unw_tdep_frame_t empty_frame = { 0, UNW_X86_64_FRAME_OTHER, -1, -1, 0, -1, -1 };
@@ -130,6 +131,7 @@ trace_cache_create (void)
   cache->log_size = HASH_MIN_BITS;
   cache->used = 0;
   cache->dtor_count = 0;
+  cache->frame_rc = 0;
   tls_cache_destroyed = 0;  /* Paranoia: should already be 0. */
   Debug(5, "allocated cache %p\n", cache);
   return cache;
@@ -430,7 +432,10 @@ tdep_trace (unw_cursor_t *cursor, void **buffer, int *size)
     d->stash_frames = 0;
     return -UNW_ENOMEM;
   }
[...]

PR #432, opened 2022-10-31
workaround for https://github.com/libunwind/libunwind/issues/425

hidva:


src/x86_64/Gtrace.c
```diff
@@ -430,7 +432,10 @@ tdep_trace (unw_cursor_t *cursor, void **buffer, int *size)
     d->stash_frames = 0;
     return -UNW_ENOMEM;
   }
+  if (cache->frame_rc > 0)
```
djwatson, line 435: So this just fails nested requests?

src/x86_64/Gtrace.c
```diff
@@ -430,7 +432,10 @@ tdep_trace (unw_cursor_t *cursor, void **buffer, int *size)
...
+    return -UNW_EINVAL;
 
+  cache->frame_rc = 1;
```
djwatson, line 438: Won't this fail if there are multiple nested requests? It is a 'reference count', but apparently it doesn't actually count?

src/x86_64/Gtrace.c
hidva: Yes

src/x86_64/Gtrace.c
hidva: > Won't this fail if there are multiple nested requests? 

Yes

> It is a 'reference count', but apparently it doesn't actually count?

At first, I thought of using `frame_rc` to represent the reference count of `unw_trace_cache_t.frames`, and `munmap(frames)` should be delayed until when `frame_rc` is 0. Then I realized that the current PR is just a workaround, let’s not make it so complicated. Currently, `frame_rc` is used to indicate whether `frames` is being used, if so, nested requests will not be allowed to use `frames` again.

hidva: I can rename `frame_rc` to`in_use` if needed.

hidva: @djwatson

PR #433, opened 2022-11-10
Updated to determine PAGE_SIZE dynamically for loongarch64

abushwang:
apply e85b65ce fix to loongarch64

Signed-off-by: Shuo Wang <wangshuo_1994@foxmail.com>


abushwang:
@djwatson 
hey, LoongArch also needs to updated to determine PAGE_SIZE dynamically according to commit e85b65cec

abushwang:
I think that the failure of ci-linux is ok because that this MR just fix loongarch64

PR #435, opened 2022-11-21
Add unw_backtrace2 function

gleocadie:
Today we have `uwn_backtrace` which collects instruction pointers and put them in the array (pointer) passed as parameter.

When we are in signal handler (and maybe sub-functions), the array will contain the signal handler instruction pointer (and the sub-functions) in addition to the thread `user-callstack`.

Since we are in  signal handler, we can provide the context and then get rid of the non essential frames.

This PR adds a new version of `backtrace`, which will collect instruction pointer starting from a context given by a signal handler.

### DIFF
--- a/src/mi/backtrace.c
+++ b/src/mi/backtrace.c
@@ -92,14 +92,28 @@ unw_backtrace2 (void **buffer, int size, unw_context_t* uc2)
   if (unlikely (unw_init_local2 (&cursor, &uc, UNW_INIT_SIGNAL_FRAME) < 0))
     return 0;
 
-  int n = size;
+  // get the first ip from the context
+  unw_word_t ip;
 
+  if (unw_get_reg (&cursor, UNW_REG_IP, &ip) < 0)
+    return 0;
+
+  buffer[0] =  (void *) (uintptr_t)ip;
+
+  // update buffer info to collect the rest of the IPs
+  buffer = buffer+1;
+  int remaining_size = size-1;
+
+  int n = remaining_size;
+
+  // returns the number of frames collected by tdep_trace or slow_backtrace
+  // and add 1 to it (the one we retrieved above)
   if (unlikely (tdep_trace (&cursor, buffer, &n) < 0))
     {
-      return slow_backtrace (buffer, size, &uc, UNW_INIT_SIGNAL_FRAME);
+      return slow_backtrace (buffer, remaining_size, &uc, UNW_INIT_SIGNAL_FRAME) + 1;
     }
 
-  return n;
+  return n + 1;
 }
 
 #ifdef CONFIG_WEAK_BACKTRACE
--- a/src/mi/backtrace.c
+++ b/src/mi/backtrace.c
@@ -77,7 +77,7 @@ unw_backtrace (void **buffer, int size)
 }
 
 int
-unw_backtrace2 (void **buffer, int size, unw_context_t* uc2)
+unw_backtrace2 (void **buffer, int size, unw_context_t* uc2, int flag)
 {
   if (size == 0)
     return 0;
@@ -89,7 +89,7 @@ unw_backtrace2 (void **buffer, int size, unw_context_t* uc2)
   // need to copy, because the context will be modified by tdep_trace
   unw_context_t uc = *(unw_context_t*)uc2;
 
-  if (unlikely (unw_init_local2 (&cursor, &uc, UNW_INIT_SIGNAL_FRAME) < 0))
+  if (unlikely (unw_init_local2 (&cursor, &uc, flag) < 0))
     return 0;
 
   // get the first ip from the context
@@ -110,7 +110,7 @@ unw_backtrace2 (void **buffer, int size, unw_context_t* uc2)
   // and add 1 to it (the one we retrieved above)
   if (unlikely (tdep_trace (&cursor, buffer, &n) < 0))
     {
-      return slow_backtrace (buffer, remaining_size, &uc, UNW_INIT_SIGNAL_FRAME) + 1;
+      return slow_backtrace (buffer, remaining_size, &uc, flag) + 1;
     }
 
   return n + 1;
--- a/src/mi/backtrace.c
+++ b/src/mi/backtrace.c
@@ -33,14 +33,15 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 /* See glibc manual for a description of this function.  */
 
 static ALWAYS_INLINE int
-slow_backtrace (void **buffer, int size, unw_context_t *uc)
+slow_backtrace (void **buffer, int size, unw_context_t *uc, int flag)
 {
   unw_cursor_t cursor;
   unw_word_t ip;
   int n = 0;
 
-  if (unlikely (unw_init_local (&cursor, uc) < 0))
+  if (unlikely (unw_init_local2 (&cursor, uc, flag) < 0))
     return 0;
+
 
   while (unw_step (&cursor) > 0)
     {
@@ -69,7 +70,33 @@ unw_backtrace (void **buffer, int size)
   if (unlikely (tdep_trace (&cursor, buffer, &n) < 0))
     {
       unw_getcontext (&uc);
-      return slow_backtrace (buffer, size, &uc);
+      return slow_backtrace (buffer, size, &uc, 0);
+    }
+
+  return n;
+}
+
+int
+unw_backtrace2 (void **buffer, int size, unw_context_t* uc2)
+{
+  if (size == 0)
+    return 0;
+    
+  if (uc2 == NULL)
+    return unw_backtrace(buffer, size);
+
+  unw_cursor_t cursor;
+  // need to copy, because the context will be modified by tdep_trace
+  unw_context_t uc = *(unw_context_t*)uc2;
+
+  if (unlikely (unw_init_local2 (&cursor, &uc, UNW_INIT_SIGNAL_FRAME) < 0))
+    return 0;
+
+  int n = size;
+
+  if (unlikely (tdep_trace (&cursor, buffer, &n) < 0))
+    {
+      return slow_backtrace (buffer, n, &uc, UNW_INIT_SIGNAL_FRAME);
     }
 
   return n;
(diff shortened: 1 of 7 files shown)


TannerFirl:
Hi! New watcher of the project, so bare with me. I had a few questions:

* Is it possible to add analogous signatures for the "fast backtrace" functions (e.g. for x86_64, src/x86_64/Gtrace.c tdep_trace())? This would allow users the ability to pass their own ucontext_t objects to backtrace.
* Is there an intention to extend this to other archs?

gleocadie:
> Hi! New watcher of the project, so bare with me. I had a few questions:

Hi
 
> * Is it possible to add analogous signatures for the "fast backtrace" functions (e.g. for x86_64, src/x86_64/Gtrace.c tdep_trace())? This would allow users the ability to pass their own ucontext_t objects to backtrace.

Good idea, will do.

> * Is there an intention to extend this to other archs?

Oh, that's a good point,  I did not think about it at first, I was relying on the same mechanism as `uwn_backtrace`.  Having it working for other archs would require much work 🤔 
What are the archs for which fast backtrace does not work ?

TannerFirl:
> I was relying on the same mechanism as uwn_backtrace. Having it working for other archs would require much work

Yea, probably best to leave additionals archs to another PR.

> What are the archs for which fast backtrace does not work ?

maybe @djwatson or someone could confirm, but it looks like fast backtrace (tdep_trace) is only supported for x86_64, arm, and aarch64:
```
firl@x86-64:~/libunwind> git ls-files -z | xargs -0 grep -n tdep_trace | grep -v tests
include/libunwind_i.h:333:   unw_tdep_trace() performance in particular. */
include/tdep-aarch64/libunwind_i.h:291:#define tdep_trace                      UNW_OBJ(tdep_trace)
include/tdep-aarch64/libunwind_i.h:328:extern int tdep_trace (unw_cursor_t *cursor, void **addresses, int *n);
include/tdep-arm/libunwind_i.h:273:#define tdep_trace                      UNW_OBJ(tdep_trace)
include/tdep-arm/libunwind_i.h:314:extern int tdep_trace (unw_cursor_t *cursor, void **addresses, int *n);
include/tdep-hppa/libunwind_i.h:239:#define tdep_trace(cur,addr,n)          (-UNW_ENOINFO)
include/tdep-ia64/libunwind_i.h:233:#define tdep_trace(cur,addr,n)          (-UNW_ENOINFO)
include/tdep-mips/libunwind_i.h:300:#define tdep_trace(cur,addr,n)          (-UNW_ENOINFO)
include/tdep-ppc32/libunwind_i.h:267:#define tdep_trace(cur,addr,n)          (-UNW_ENOINFO)
include/tdep-ppc64/libunwind_i.h:322:#define tdep_trace(cur,addr,n)          (-UNW_ENOINFO)
include/tdep-riscv/libunwind_i.h:266:#define tdep_trace(cur,addr,n)          (-UNW_ENOINFO)
include/tdep-s390x/libunwind_i.h:220:#define tdep_trace(cur,addr,n)          (-UNW_ENOINFO)
include/tdep-sh/libunwind_i.h:240:#define tdep_trace(cur,addr,n)          (-UNW_ENOINFO)
include/tdep-tilegx/libunwind_i.h:215:#define tdep_trace(cur,addr,n)          (-UNW_ENOINFO)
include/tdep-x86/libunwind_i.h:253:#define tdep_trace(cur,addr,n)          (-UNW_ENOINFO)
include/tdep-x86_64/libunwind_i.h:247:#define tdep_trace                      UNW_OBJ(tdep_trace)
include/tdep-x86_64/libunwind_i.h:296:extern int tdep_trace (unw_cursor_t *cursor, void **addresses, int *n);
src/aarch64/Gtrace.c:386:     if ((ret = unw_tdep_trace(&cur, addrs, &depth)) < 0)
src/aarch64/Gtrace.c:400:tdep_trace (unw_cursor_t *cursor, void **buffer, int *size)
src/arm/Gtrace.c:387:     if ((ret = unw_tdep_trace(&cur, addrs, &depth)) < 0)
src/arm/Gtrace.c:401:tdep_trace (unw_cursor_t *cursor, void **buffer, int *size)
src/mi/backtrace.c:69:  if (unlikely (tdep_trace (&cursor, buffer, &n) < 0))
src/x86_64/Gtrace.c:384:     if ((ret = unw_tdep_trace(&cur, addrs, &depth)) < 0)
src/x86_64/Gtrace.c:398:tdep_trace (unw_cursor_t *cursor, void **buffer, int *size)
```

gleocadie:
> add a doc like `doc/unw_backtrace.man` and link it in `doc/Makefile.am` also add a test in `tests/check-namespace.sh.in` and myabe in `tests/Gtest-trace.c`

Will do. thx

gleocadie:
> * Is it possible to add analogous signatures for the "fast backtrace" functions (e.g. for x86_64, src/x86_64/Gtrace.c tdep_trace())? This would allow users the ability to pass their own ucontext_t objects to backtrace.

Today, if I understood it correct, "fast backtrace" is accessible only by `unw_backtrace` so I do not see what would be the gain in adding the context to the `tdep_trace`.
Or did I miss something ?

TannerFirl:
> > * Is it possible to add analogous signatures for the "fast backtrace" functions (e.g. for x86_64, src/x86_64/Gtrace.c tdep_trace())? This would allow users the ability to pass their own ucontext_t objects to backtrace.
> 
> Today, if I understood it correct, "fast backtrace" is accessible only by `unw_backtrace` so I do not see what would be the gain in adding the context to the `tdep_trace`. Or did I miss something ?

You are right, my bad

src/mi/backtrace.c
```diff
@@ -69,12 +70,52 @@ unw_backtrace (void **buffer, int size)
...
+unw_backtrace2 (void **buffer, int size, unw_context_t* uc2)
+{
+  if (size == 0)
+    return 0;
+    
+  if (uc2 == NULL)
+    return unw_backtrace(buffer, size);
+
+  unw_cursor_t cursor;
+  // need to copy, because the context will be modified by tdep_trace
```
TannerFirl, line 89: uc looks like a shallow copy of uc2. Based on the comment, does this need to be a deep copy?

src/mi/backtrace.c
```diff
@@ -69,12 +70,52 @@ unw_backtrace (void **buffer, int size)
...
       unw_getcontext (&uc);
-      return slow_backtrace (buffer, size, &uc);
+      return slow_backtrace (buffer, size, &uc, 0);
     }
 
   return n;
 }
 
+int
+unw_backtrace2 (void **buffer, int size, unw_context_t* uc2)
```
TannerFirl, line 80: could we handle any `unw_context_t *` object by adding a `int flag` to the signature, similar to `unw_init_local2`? Use case: foo calls bar calls baz. bar calls unw_getcontext, passes the uc2 to baz, and baz calls unw_backtrace2 which returns the callstack from bar's context.

I think only change would be to signature and changing:
```
if (unlikely (unw_init_local2 (&cursor, &uc, flag) < 0))
```
In the case of signal handler, call this function with flag==UNW_INIT_SIGNAL_FRAME. In the case of no signal handler, call with flag==0

TannerFirl: Here is an example diff between e831ee22d7356d46f65ad7a93553f54e235d9c1e and the proposed change in the comment above:
```
diff --git a/src/mi/backtrace.c b/src/mi/backtrace.c
index 8fea3444..0f2863fc 100644
--- a/src/mi/backtrace.c
+++ b/src/mi/backtrace.c
@@ -77,7 +77,7 @@ unw_backtrace (void **buffer, int size)
 }
 
 int
-unw_backtrace2 (void **buffer, int size, unw_context_t* uc2)
+unw_backtrace2 (void **buffer, int size, unw_context_t* uc2, int flag)
 {
   if (size == 0)
     return 0;
@@ -89,7 +89,7 @@ unw_backtrace2 (void **buffer, int size, unw_context_t* uc2)
   // need to copy, because the context will be modified by tdep_trace
   unw_context_t uc = *(unw_context_t*)uc2;
 
-  if (unlikely (unw_init_local2 (&cursor, &uc, UNW_INIT_SIGNAL_FRAME) < 0))
+  if (unlikely (unw_init_local2 (&cursor, &uc, flag) < 0))
     return 0;
 
   // get the first ip from the context
@@ -110,7 +110,7 @@ unw_backtrace2 (void **buffer, int size, unw_context_t* uc2)
   // and add 1 to it (the one we retrieved above)
   if (unlikely (tdep_trace (&cursor, buffer, &n) < 0))
     {
-      return slow_backtrace (buffer, remaining_size, &uc, UNW_INIT_SIGNAL_FRAME) + 1;
+      return slow_backtrace (buffer, remaining_size, &uc, flag) + 1;
     }
 
   return n + 1;
```

gleocadie: done

TannerFirl:
im not too familiar with the release policy, but any chance we could get this into v1.7-stable? seems like it should be relatively benign

djwatson:
Looks reasonable to me, thanks. 

Yea I think this is ok for 1.7, I'll push an rc2.  

Thanks for the test & man page!

src/mi/backtrace.c
```diff
@@ -69,12 +70,52 @@ unw_backtrace (void **buffer, int size)
...
+  buffer = buffer+1;
+  int remaining_size = size-1;
+
+  int n = remaining_size;
+
+  // returns the number of frames collected by tdep_trace or slow_backtrace
+  // and add 1 to it (the one we retrieved above)
+  if (unlikely (tdep_trace (&cursor, buffer, &n) < 0))
+    {
+      return slow_backtrace (buffer, remaining_size, &uc, flag) + 1;
```
TannerFirl, line 113: @gleocadie its hard for me to test this, but if `tdep_trace` modifies `uc` (as commented above) and then fails, it seems to me it would no longer be safe to pass `&uc` to slow_backtrace. Do you agree?

gleocadie: `tdep_trace` modifies `uc` but not the `unw_step` (at least I did not see any hint of modification).
The `slow_backtrace` relies only on `unw_step`.
From my tests (only on x86_64 arch) I had no crash.

TannerFirl: is it ok to use the modified `&uc` in slow_backtrace's `if (unlikely (unw_init_local2 (&cursor, uc, flag) < 0))` though?

gleocadie: oh I see your point. maybe a another copy should be done.

PR #445, opened 2022-12-16
fix byte_order_is_valid function logic

he7850:
byte_order_is_valid is an inline function extracted from unw_create_addr_space in libunwind 1.6, currently its logic mismatches the function intention.

Signed-off-by: hubin <hubin73@huawei.com>

### DIFF
--- a/include/libunwind_i.h
+++ b/include/libunwind_i.h
@@ -121,8 +121,8 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 static inline int
 byte_order_is_valid(int byte_order)
 {
-    return byte_order != UNW_BIG_ENDIAN
-        && byte_order != UNW_LITTLE_ENDIAN;
+    return byte_order == UNW_BIG_ENDIAN
+        || byte_order == UNW_LITTLE_ENDIAN;
 }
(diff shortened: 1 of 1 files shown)


he7850:
previous in libunwind 1.5:
```
unw_create_addr_space (unw_accessors_t *a, int byte_order)
{
...
  if (byte_order != 0 && byte_order != __LITTLE_ENDIAN
      && byte_order != __BIG_ENDIAN)
    return NULL;
...
}
```

current in libunwind 1.6:
```
unw_create_addr_space (unw_accessors_t *a, int byte_order)
{
...
  if (byte_order != 0 && byte_order_is_valid(byte_order) == 0)
    return NULL;
...
}
static inline int
byte_order_is_valid(int byte_order)
{
    return byte_order != UNW_BIG_ENDIAN
        && byte_order != UNW_LITTLE_ENDIAN;
}
```

kasperk81:
@bregma related to https://github.com/libunwind/libunwind/pull/188

bregma:
Confirmed. All of the unit tests pass 0 as the byte_order argument to `unw_create_addr_space()` so this function was never exercised. In fact, the **`UNW_*_ENDIAN`** macros are internal, which is another issue.

Issue #453, opened 2023-01-11
valgrind reports UMRs due to validate_mem writing uninitialzed memory to pipe.


peadar:
When running valgrind on a program that uses libunwind, it reports UMR errors. (Log attached)

This is caused by write_validate using the page-aligned address of the page it is attempting to validate as readable as an argument to syscall(SYS_write...). Apparently, syscall is used rather than the write() stub in libc because that circumvented the same warning with ASAN, but valgrind is clever enough to understand that SYS_write accesses memory pointed to by the arguments.
[log.txt](https://github.com/libunwind/libunwind/files/10390615/log.txt)

I have a patch that addresses this by having validate_mem use the original address that access_mem will try and access, rather than the start of that page - because we are going to read this anyway, it is, if things are functioning correctly, guaranteed to be initialized.

bregma:
Closed by #454

Issue #232, opened 2021-02-04
Possible bug in x86_64 validate_mem


godlygeek:
[The `validate_mem` function](https://github.com/libunwind/libunwind/blob/4407cf07d53de767c2a2b8297eaea00ce33ccd32/src/x86_64/Ginit.c#L287) wants to validate that the `unw_word_t` that it's been passed points to memory that is both mapped and readable. This word is apparently able to span across two different pages, per [this line](https://github.com/libunwind/libunwind/blob/4407cf07d53de767c2a2b8297eaea00ce33ccd32/src/x86_64/Ginit.c#L294).

Either `mincore` or `msync` is used to validate that the page or pages are mapped, and then `write_validate` is called to confirm that it's readable - but the call to `write_validate` only reads from the _first_ byte of the address, not the rest of the word.

Unless I'm misreading something, it seems like this would result in a segfault if the word spans across two pages, the first page is readable, and the second page is not.

bregma:
Fixed by PR #454

PR #454, opened 2023-01-11
Fix UMRs indicated by valgrind (x86_64)

peadar:
See issue #453 - The valgrind tool can be used to detect errors such as uninitialized memory reads, etc, at runtime. This is similar to what ASAN does, but does not require compile-time intervention (at the expense of a far greater runtime performance impact)

The `write_validate` function, in conjunction with one of the `mem_validate` functions, is used to ensure that a specific page is readable before actually accessing a specific address from that page (in order to avoid generating a segmentation fault). It does so by writing the first byte of the page to a pipe. If the page is mapped readable, the write succeeds, while if the page is not mapped, or not writeable, then the write will fail with EFAULT or similar.

It seems that there has previously already been issues with ASAN here - rather than using the `write` function in libc,  `syscall` is used in an attempt to circumvent the memory checks from ASAN.

This patch changes the `write_validate` function to take the original pointer, rather than the page-aligned version, and writes from that specific address. Given that we are about to read that value anyway, (probably to retrieve the contents of the stack in order to fulfill some location expression in the dwarf data), the data at that location should actually be initialized, while the data at the page-aligned offset may not be.  This removes UMR errors reported by valgrind, and possibly other tools that understand the semantics of individual numbered system calls invoked via syscall (We can probably actually remove the syscall() variant of the code, but I've left as-is to reduce the amount of churn for now)

While here, I've also fixed an issue where only the first page containing the data accessed by access_mem is validated as part of `write_validate` - `validate_mem`/`write_validate` now take a size argument, and we instead walk through the content of the value to be verified, validating acces every page-size forward of the first address, or, if that exceeds the extent of the object, the last byte in the object. This function is only used with sizeof (unw_word_t) as a size, so the pointers can only possibly straddle a single page boundary, but it's probably clearer to write it for the general case.  (See issue #232)

This addresses the issue for x86_64. Once done, I can do similar for other architectures using the `write_validate()` mechanism

### DIFF
--- a/src/x86_64/Ginit.c
+++ b/src/x86_64/Ginit.c
@@ -117,7 +117,7 @@ open_pipe (void)
 
 ALWAYS_INLINE
 static int
-write_validate (void *addr)
+write_validate (unw_word_t addr)
 {
   int ret = -1;
   ssize_t bytes = 0;
@@ -142,7 +142,7 @@ write_validate (void *addr)
        /* use syscall insteadof write() so that ASAN does not complain */
        ret = syscall (SYS_write, mem_validate_pipe[1], addr, 1);
 #else
-	  ret = write (mem_validate_pipe[1], addr, 1);
+	  ret = write (mem_validate_pipe[1], (void *)addr, 1);
 #endif
     }
   while ( errno == EINTR );
@@ -150,10 +150,10 @@ write_validate (void *addr)
   return ret;
 }
 
-static int (*mem_validate_func) (void *addr, size_t len);
-static int msync_validate (void *addr, size_t len)
+static int (*mem_validate_func) (unw_word_t addr, size_t len);
+static int msync_validate (unw_word_t addr, size_t len)
 {
-  if (msync (addr, len, MS_ASYNC) != 0)
+  if (msync ( (void *)uwn_page_start (addr), len, MS_ASYNC) != 0)
     {
       return -1;
     }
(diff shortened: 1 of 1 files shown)


src/x86_64/Ginit.c
```diff
@@ -142,18 +142,18 @@ write_validate (void *addr)
...
   return ret;
 }
 
-static int (*mem_validate_func) (void *addr, size_t len);
-static int msync_validate (void *addr, size_t len)
+static int (*mem_validate_func) (unw_word_t addr, size_t len);
+static int msync_validate (unw_word_t addr, size_t len)
 {
-  if (msync (addr, len, MS_ASYNC) != 0)
+  if (msync ( (void *)addr, len, MS_ASYNC) != 0)
```
djwatson, line 156: It looks like msync needs to be page aligned also?

man msync "EINVAL addr is not a multiple of PAGESIZE"

djwatson:
One comment on msync, but otherwise looks reasonable.  THanks

src/x86_64/Ginit.c
peadar: Ah, I'd even checked the manpage, but missed the reference in the ERRORS section

I have fixed the problem, and added an env var to force the validation function. This allowed me to test it manually at least. If you'd rather I removed env var toggle from the production code, let me know.

peadar:
Thanks for the fast feedback. Updated the patch. If everything else is ok, I can prepare same for other platforms where relevant

src/x86_64/Ginit.c
```diff
@@ -283,21 +310,30 @@ cache_valid_mem(unw_word_t addr)
...
+  if (len == 0)
     return 0;
 
-  if (mem_validate_func ((void *) addr, len) == -1)
+  if (uwn_page_start (addr) == 0)
     return -1;
 
-  cache_valid_mem(addr);
+  unw_word_t lastbyte = addr + (len - 1); // highest addressed byte of data to access
+  for (;;)
```
kasperk81, line 322:
```suggestion
  while (1)
```

peadar:
Is there anything else required from me here? I think the CI tests that failed are failing for other commits too, if I read them correctly.
Let me know if I can help.

bregma:
This all seems reasonable.  Bringing it to other targets would be nice.

peadar:
Thanks! I can spend some time covering other affected targets.

PR #457, opened 2023-02-03
Fix `single-bit-bitfield-constant-conversion` warning

am11:
```sh
# with clang-16 rc1
$ CC=clang-16 sh -c 'autoreconf -i && ./configure && make'
...
x86_64/Gstep.c:223:55: warning: implicit truncation from 'int' to a one-bit wide bit-field changes value from 1 to -1 [-Wsingle-bit-bitfield-constant-conversion]
                            c->frame_info.cfa_reg_rsp = 1;
                                                      ^ ~
```


bregma:
This is fine for master.

PR #471, opened 2023-02-25
Add introspection for march=armv8-a+sve

### LINKED ISSUE #470: aarch64 unit test failures on non-Linux-native builds
The Garm64-test-sve-signal.c and Larm64-test-sve-signal.c unit tests are always compiled unconditionally but fail to build with a non-Linux-native toolchain with the following error.
```
2023-02-24T21:31:26.6397435Z   CC       Larm64_test_sve_signal-Larm64-test-sve-signal.o
2023-02-24T21:31:26.6544281Z cc1: error: unknown value ‘native’ for ‘-march’
2023-02-24T21:31:26.6638100Z cc1: note: valid arguments are: armv8-a armv8.1-a armv8.2-a armv8.3-a armv8.4-a armv8.5-a armv8.6-a armv8-r
```
The `-march=native` switch is supported only on native Linux GCC. This needs to be detected at configure time or else something like `-march=armv8-a+sve` used instead.

am11:
Fix #470

```sh
checking if compiler supports -march=armv8-a+sve... configure: yes # or no, depending on the CI leg
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -389,6 +389,10 @@ if test x$GCC = xyes -a x$intel_compiler != xyes -a x$qcc_compiler != xyes -a x$
   LIBCRTS="-lgcc_s"
 fi
 
+AC_MSG_CHECKING([if compiler supports -march=native])
+AM_CONDITIONAL([COMPILER_SUPPORTS_MARCH_NATIVE], ["$CC" -march=native -S -o /dev/null -xc /dev/null > /dev/null 2>&1])
+AM_COND_IF([COMPILER_SUPPORTS_MARCH_NATIVE], [AC_MSG_NOTICE(yes)], [AC_MSG_NOTICE(no)])
+
 AC_MSG_CHECKING([for __builtin___clear_cache])
 AC_LINK_IFELSE(
   [AC_LANG_PROGRAM([[]], [[__builtin___clear_cache(0, 0)]])],
--- a/configure.ac
+++ b/configure.ac
@@ -390,11 +390,11 @@ if test x$GCC = xyes -a x$intel_compiler != xyes -a x$qcc_compiler != xyes -a x$
 fi
 
 OLD_CFLAGS="${CFLAGS}"
-CFLAGS="${CFLAGS} -march=native"
-AC_MSG_CHECKING([if compiler supports -march=native])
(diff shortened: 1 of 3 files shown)


bregma:
This doesn't support the (very important) case of cross compiling.

configure.ac
```diff
@@ -389,6 +389,10 @@ if test x$GCC = xyes -a x$intel_compiler != xyes -a x$qcc_compiler != xyes -a x$
   LIBCRTS="-lgcc_s"
 fi
 
+AC_MSG_CHECKING([if compiler supports -march=native])
+AM_CONDITIONAL([COMPILER_SUPPORTS_MARCH_NATIVE], ["$CC" -march=native -S -o /dev/null -xc /dev/null > /dev/null 2>&1])
```
bregma, line 393: This should use AC_COMPILE_IFELSE because there is often more to invoking the compiler than just $CC.

am11:
`march=native` (used in one test) can be deleted. It is not buying us anything.

bregma:
The test explicitly checks ARMv8 SVE-specific functionality. Sure, the test code is guarded internally by a macro that some toolchains define in the presence of SVE support for a particular CPU, but that macro is only defined when requested on the command line (eg. generically with `-march=armv8-a+sve` or, on Linux built self-hosted for certain vendor silicon, `-march=native`).

What the command-line option buys us is building and testing the SVE code path. The "fix" of just not enabling that code path is not really a fix.

am11:
Do you see any false positive / negative in the CI?

bregma:
Without enabling SVE support in the compiler, `Garm64-test-sve-signal.c` compiles to an (effectively) empty test case. That makes it a false positive in my eyes since I don't consider not testing as a way to avoid test failures to be a legitimate strategy.

am11:
OK, I was going by the CI, and it looks like CI does not exercise the SVE code path even with `AC_COMPILE_IFELSE` (which takes cross compilation into account).

bregma:
`-march=native` does not work for cross compiling. It only works for native self-hosted builds on Linux on certain silicon. It has to be `-march=armv8-a+sve` to work for cross compiling. The cross-compiled case is important.

bregma:
SVE test now builds (and fails, but that's an improvement). Some other test is hanging the ARMv7 and ARMv8 test runs, unrelated to this issue. Committing as correct.

PR #484, opened 2023-03-30
Provide AS-safe allocator to LZMA

### LINKED ISSUE #123: unw_get_proc_name can call an AS-unsafe function
The documentation claims the following:
> unw_get_proc_name() is thread-safe. If cursor cp is in the local address-space, this routine is also safe to use from a signal handler.

However, invoking this function on an x86_64 Linux machine on a local-address-space cursor might lead to the following call stack:
```
#0  lzma_stream_buffer_decode (memlimit=0x7fffffffca40, flags=0, allocator=0x0, in=0x7fffbd3144d0 "\375\067zXZ", in_pos=0x7fffffffca48, in_size=915412, out=0x7fffbc936000 "", 
    out_pos=0x7fffffffca50, out_size=4470304) at ../../../../src/liblzma/common/stream_buffer_decoder.c:21
#1  0x00007ffff7f8017b in _Uelf64_extract_minidebuginfo (ei=0x7fffffffcb90, mdi=0x7fffffffcb00) at elfxx.c:278
#2  0x00007ffff7f80261 in _Uelf64_get_proc_name_in_image (as=0x7ffff7f892a0 <local_addr_space>, ei=0x7fffffffcb90, segbase=93824993083392, mapoff=851968, ip=93824994457837, 
    buf=0x7fffffffd110 "foo", buf_len=42, offp=0x7fffffffd108) at elfxx.c:320
#3  0x00007ffff7f80393 in _Uelf64_get_proc_name (as=0x7ffff7f892a0 <local_addr_space>, pid=6674, ip=93824994457837, buf=0x7fffffffd110 "foo", buf_len=42, offp=0x7fffffffd108) at elfxx.c:353
#4  0x00007ffff7f75b68 in get_static_proc_name (as=0x7ffff7f892a0 <local_addr_space>, ip=93824994457837, buf=0x7fffffffd110 "foo", buf_len=42, offp=0x7fffffffd108, arg=0x7fffffffc520)
    at x86_64/Ginit.c:249
#5  0x00007ffff7f7415b in get_proc_name (as=0x7ffff7f892a0 <local_addr_space>, ip=93824994457837, buf=0x7fffffffd110 "foo", buf_len=42, offp=0x7fffffffd108, arg=0x7fffffffc520)
    at mi/Gget_proc_name.c:93
#6  0x00007ffff7f741e8 in _ULx86_64_get_proc_name (cursor=0x7fffffffcd10, buf=0x7fffffffd110 "foo", buf_len=42, offp=0x7fffffffd108) at mi/Gget_proc_name.c:111
```
The `lzma_stream_buffer_decode` function doesn't give any AS-safety guarantees; moreover, it claims to allocate memory with `malloc`/`free` if `allocator=NULL` (which is the case). Hence, `unw_get_proc_name` is unsafe to use from a signal handler.

bertwesarg:
Closes #123


configure.ac
```diff
@@ -305,7 +305,7 @@ AC_MSG_RESULT([$enable_msabi_support])
 LIBLZMA=
 AC_MSG_CHECKING([whether to support LZMA-compressed symbol tables])
 AC_ARG_ENABLE(minidebuginfo,
-AS_HELP_STRING([--enable-minidebuginfo], [Enables support for LZMA-compressed symbol tables]),, [enable_minidebuginfo=auto])
+AS_HELP_STRING([--enable-minidebuginfo], [Enables support for LZMA-compressed symbol tables]),, [enable_minidebuginfo=no])
```
bregma, line 308: Why? A lot of existing builds use the autodetection. They will break with this change. If your build target audetects support for LZMA-compressed symbol tables and you explicitly don't want them (an exceptional situation), it's better that you explicitly opt out instead of forcing the default case to explicitly opt in.

bregma (commented):
This PR has at least 3 unrelated changes.

1. The LZMA allocator mentioned in the title.
2. Adding `mi_init()` to the coredump and ptrace remotes
3. Renaming `uwn_page_start()` to `unw_page_start()`

One (documented) change per PR is preferred, since it makes bisection and fork maintenance easier.

configure.ac
bertwesarg: see the referenced issue #123

bertwesarg:
> * The LZMA allocator mentioned in the title.
> 
> * Adding `mi_init()` to the coredump and ptrace remotes

This is needed, so that `unw_page_size` gets initialized. Will make it its own commit, so that it is clear why this is now needed.

> 
> * Renaming `uwn_page_start()` to `unw_page_start()`

I would rather drop this commit entirely, instead of creating its own PR. Your call.

bertwesarg:
> makes bisection and fork maintenance easier

btw, this project does forward-merges for PR, thus no merge-commits. This is the ideal situation for bisection

> One (documented) change per PR is preferred

and without a merge commit there is also no back reference from master to the PR

So what remains are just the individual commits in master, without any notion of "PR"

bertwesarg:
We could avoid the TLS by moving the struct into `elf_w (extract_minidebuginfo)` and pass it as the value of `.opaque` in the `lzma_allocator` struct. Stack usage would be 257 currently. I did not made excessive testing, whether 16 allocations are too much.

I start to prefer the stack based implementation, as TLS might also need to call `malloc` inside glibc.

Any comments?

bertwesarg:
what needs to be done from my side, to get an approval? thanks

PR #486, opened 2023-03-30
Provide syscall wrappers for `mmap` and `munmap`

bertwesarg:
These functions are assumed to be direct wrappers to the syscalls, but in the past (see Open MPI) and also now (see UCX) runtimes do intercept these functions and do AS-unsafe stuff inside there runtimes. Hence provide internal syscall wrappers for these functions to improve AS-safety.

### DIFF
--- a/include/libunwind_i.h
+++ b/include/libunwind_i.h
@@ -217,6 +217,26 @@ do {                                            \
 
 #define SOS_MEMORY_SIZE 16384   /* see src/mi/mempool.c */
 
+/* Provide an internal syscall version of mmap to improve signal safety. */
+static ALWAYS_INLINE void *
+mi_mmap (void *addr, size_t len, int prot, int flags, int fd, off_t offset)
+{
+#ifdef SYS_mmap
+#ifdef __syscall // prefer over syscall on *BSD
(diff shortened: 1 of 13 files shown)


bregma:
We have to support targets other than GNU/Linux. Embedded GNU libc calls with Linux kernel bindings is not going to fly.

bertwesarg:
> We have to support targets other than GNU/Linux. Embedded GNU libc calls with Linux kernel bindings is not going to fly.

please point me how to make this portable then. thanks

include/libunwind_i.h
```diff
@@ -58,6 +58,8 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 #include <errno.h>
 #include <stdio.h>
 
+#include <sys/syscall.h>   /* For SYS_xxx definitions */
```
bregma, line 61: This non-portable header needs to be surrounded with an `#if defined(HAVE_SYS_SYSCALL_H)`.

bregma (changes requested):
Include file needs #ifdefing but otherwise looks good.

PR #490, opened 2023-04-19
Allow to use a custom `dl_iterate_phdr` implementation

### LINKED ISSUE #16: dl_iterate_phdr not async signal safe
The implementation of dl_iterate_phdr in glibc both takes locks, and allocates memory.  People hit this sporadically in production.   

Barring a fix in glibc, the best fix is probably an optionally linked library that will override dlopen, dlclose, and dl_iterate_phdr with a lock-free and non-mallocing cache.

bertwesarg:
Closes #16

Originally posted here:

https://lists.nongnu.org/archive/html/libunwind-devel/2016-02/msg00024.html

Already acked by @milianw (see ML) and thumbed-up by @djwatson (see issue)

### DIFF
--- a/include/libunwind-common.h.in
+++ b/include/libunwind-common.h.in
@@ -249,6 +249,10 @@ typedef struct unw_save_loc
   }
 unw_save_loc_t;
 
+struct dl_phdr_info;
+typedef int (*unw_iterate_phdr_callback_t) (struct dl_phdr_info *, size_t, void *);
+typedef int (*unw_iterate_phdr_func_t) (unw_iterate_phdr_callback_t, void *);
+
 /* These routines work both for local and remote unwinding.  */
 
 #define unw_local_addr_space		UNW_OBJ(local_addr_space)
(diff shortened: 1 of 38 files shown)


bregma (approved):
Looks good.

bregma:
I think this breaks the Windows cross builds in CI though.

bertwesarg:
I have new findings/questions, while thinking about solving the windows cross-build issues:

1. `dl_iterate_phdr` is Linux/Glibc only. Should the new `unw_set_iterate_phdr_function` API always be available or only on platforms which support it?
2. Does it make sense do have `unw_set_iterate_phdr_function` in the remote part at all?
3. There is already a `dl_iterate_phdr` in `src/dl-iterate-phdr.c`, but only for older Android. This could be used as the default.

bregma:
>     1. `dl_iterate_phdr` is Linux/Glibc only. Should the new `unw_set_iterate_phdr_function` API always be available or only on platforms which support it?

 It is not GNU/Linux only. The BSDs and QNX at the very least have it.

>     2. Does it make sense do have `unw_set_iterate_phdr_function` in the remote part at all?

It doesn't make sense for a remote unwinder to call `dl_iterate_phdr()` because that asks the dynamic loader to iterate over the PHDRs in the current (local) image. Having a `unw_set_iterate_phdr_function()` shouldn't hurt but initializing the default may be a challenge.

>     3. There is already a `dl_iterate_phdr` in `src/dl-iterate-phdr.c`, but only for older Android. This could be used as the default.

The problem with that one is it *is* Linux specific. I would think it's far better to autodetect if the target has `dl_iterate_phdr()` (there's already a check in configure.,ac and `HAVE_DL_ITERATE_PHDR` will be defined as appropriate) and fall back to a dummy function.

bertwesarg:
Thanks for the pointer, as this seams more like an optimization after we get the the general idea into libunwind, I don't wont pursue this for now. But the documentation has a glitch anyway, `struct link_map *dlf_link_map` is listed twice as a member of `struct dl_find_object`. One should be named `dlfo_link_map` and the second should probably be named `dlfo_eh_frame` with type `void*`.

bertwesarg:
I could guard it with `#ifndef UNW_REMOTE_ONLY`. Something like this is already done in IA64. There a member inside `struct unw_addr_space` is already removed, which I could do it too

Issue #489, opened 2023-04-15
Ltest-mem-validate and other test case fail on linux


ziyangc97:
hi, I am runing make check of libunwind 1.6.2 on openEuler(linux) arm system and there are 4 test cases failed.
[test-suite.log](https://github.com/libunwind/libunwind/files/11239083/test-suite.log)

My OS and libunwind info:
kernel:5.10.0-60.18.0.50.oe2203.aarch64
aarch64
libunwind:1.6.2
libtool:libtool-2.4.6-34

I try to debug but found out and stuck here:
file Ltest-mem-validate.c
```
if (childpid)
        {
          wait(&status);
          printf("iter i = %d, WIFEXITED(status) = %d\n", i, WIFEXITED(status));
          if (WIFEXITED(status))
              return WEXITSTATUS(status);
          else {
            printf("return -1\n");
```
the test-suite.log shows:
```
iter i = 3, WIFEXITED(status) = 0
return -1
FAIL Ltest-mem-validate (exit status: 255)
```

Please provide some hint why this is happening or some guidance how to singly run this test because I try to run it by gdb.

Appreciate anyone's help

bregma:
At the very least the return value of `wait()` needs to be checked for success or failure. As to why it would be failing (likely ECHILD, but possibly EINTR) will take more investigation.

bregma:
I suspect the real reason for failure is that the stack frame for 5 or more recursive calls of `consume_and_run()` may exceed 4k on aarch64. If you have the time, that's where I'd start investigating.

bregma:
Looking further, I think this test is bogus and highly reliant on (a) undefined behaviour (strict aliasing violation) and (2) ABI-specific behaviour (eg. stack growth direction, calling conventions). That it works on x86* is a fluke.

The test code contains the comment
>     We make the assumption that we are able to rewind far enough
>     (steps > 5) before touching the forbidden region in the stack,
>     at which point the unwinding should stop gracefully.
but if the unwinder hits the mprotected region (which it will, at least on ARM ABIs), the OS is supposed to raise a segfaukt, which the unwinder does not handle gracefully.

It's not clear what this test is attempting to validate (other than that the unwinder detects artificial restrictions on memory access, which it does not do and of course the test usually fails). If anyone could enlighten me on the purpose of this test I'm all ears otherwise I think it should just be removed.

he7850:
In my view, Ltest-mem-validate testcase intent is:

[...]

PR #494, opened 2023-05-03
aarch64: unw_step() validates address before calling dwarf_get

he7850:
fix failed testcase Ltest-mem-validate under aarch64 (https://github.com/libunwind/libunwind/issues/489)


bregma (approved):
Yes, this is clearly the functionality that Ltest-mem-validate is trying to exercise. Thatnks.

PR #495, opened 2023-05-03
Initial spike of implementing UNW_INFO_FORMAT_DYNAMIC for dwarf targets

KJTsanaktsidis:
This is a patch to start a discussion about UNW_INFO_FORMAT_DYNAMIC.

Currently, the documentation in libunwind-dynamic(3) says that this format is "the preferred dynamic unwind-info format" and that "it is generally the one used by full-blown runtime code generators". However, it seems that it's only actually implemented on Itanium, which rather substantially limits its usefulness!

This format is interesting for me because I want to try using it in a JIT compiler under the following constraints:
  * Code is generated in small chunks incrementally; thus, registering & unregistering unwind info needs to be fast.
  * Most blocks will use one of a few standard stack layouts. The ability to share unw_dyn_info_region_t instances between procedures is therefore attractive.
  * Code can be garbage-collected and inserted into freed pages; the IP of generated code is not eternally increasing. This makes the requirement that DWARF unwind tables be sorted in IP order problematic, because inserting code into the middle of the table requires much copying.
  * The project does not wish to take a dependency on a library for serializing DWARF info; the structs for UNW_INFO_FORMAT_DYNAMIC however can easily be assembled in the code generator code with nothing else but libunwind.
  * The JIT compiler does not emit very complex stack manipulations, so the full complexity of DWARF is not needed.

I have attempted to implement this format for x86_64 by filling in the "Not implemented yet" method `parse_dynamic`. This works inside the existing dwarf unwinder by walking through the UNW_INFO_FORMAT_DYNAMIC data and converting it to the existing register rule structure.

This seems to "work" - there's a test, which asserts generates both sp-relative and fp-relative unwind info for a procedure, and it seems to be able to successfully unwind from all points in the procedure.

Discussion points:

  * I implemented a few new DWARF_WHERE_ rules inside `apply_reg_state` to implement some of the behaviour. Strictly speaking, this is unnescessary; all of these rules could be replaced by use of DWARF_WHERE_EXPR, however I couldn't see any existing code to generate dwarf expressions, and it didn't seem like a good idea to serialise an expression just to immediately deserialise it again, so I did it this way.

  * I don't think the existing set of ops documented in libunwind-dynamic(3) is sufficient to express normal x86_64 stacks. In particular, I had to implement a couple of extra ops. :: UNW_DYN_RESTORE_REG is needed because x86_64 does not have a single-instruction way to tear down a stack frame (which is what I assume UNW_DYN_POP_FRAMES does on Itanium). So, `pop %rbp` needs this. :: UNW_DYN_SAVE_SP is needed because calling a function on x86_64 pushes the return address onto the stack, and the old %rbp is also pushed onto the stack, _before_ `mov %rsp, %rbp` happens; thus, the value of %rsp being saved into %rbp is _NOT_ "%rsp from the previous frame", but rather "%rsp from the previous frame + two words". So, I implemented these.

  * It wasn't clear what the behaviour of mixing UNW_DYN_ADD and UNW_DYN_SPILL_SP_REL should be; should the existing sp-relative spills be updated if the stack pointer is decremented? I decided that "yes" was the most useful answer, and that seemed to be what I deduced should happen from this old message: https://www.hpl.hp.com/hosted/linux/mail-archives/libunwind/2004-March/000217.html

  * It felt kind of wrong to implement this in src/dwarf, since it has ~nothing to do with dwarf. However, that's where the "implement me" stub was. Is this the right decision, or should the arch-specific code explicitly call e.g. `dynamic_step` _and_ `dwarf_step`?

  * I have no Itanium hardware, so I have absolutely no way to know how the existing implementation behaves other than squinting at code and guessing. I don't suppose there's an emulator or such the project uses to keep this working? I couldn't find anything.

  * Because this is currently only supported on Itanium, I have to assume the current set of users for it is ~zero. With that being the case, is implementing the existing interface something that we should do here, or should something else be designed specifically to meet the needs of CPU archs and JIT compilers from this millenium?

Anyway - interested in thoughts and feedback. Thanks for reading!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,6 +79,7 @@ tests/[GL]ia64-test-stack
 tests/ia64-test-dyn1
 tests/ia64-test-sig
 tests/[GL]x64-test-dwarf-expressions
+tests/x64-test-dyn1
 tests/x64-unwind-badjmp-signal-frame
 tests/*.log
 tests/*.trs
--- a/include/dwarf.h
+++ b/include/dwarf.h
@@ -232,6 +232,10 @@ typedef enum
     DWARF_WHERE_EXPR,           /* register saved */
     DWARF_WHERE_VAL_EXPR,       /* register has computed value */
     DWARF_WHERE_CFA,            /* register is set to the computed cfa value */
+    DWARF_WHERE_DYN_OFFSET,     /* register has the value in the previous frame, plus an offset */
+    DWARF_WHERE_DYN_FP_REL,     /* register is spilled to this fp-relative offset */
+    DWARF_WHERE_DYN_SP_REL,     /* register is spilled to this sp-relative offset */
+    DWARF_WHERE_DYN_SAVE_SP,    /* used to implement UNW_DYN_SAVE_SP */
   }
 dwarf_where_t;
 
--- a/include/libunwind-dynamic.h
+++ b/include/libunwind-dynamic.h
@@ -67,7 +67,13 @@ typedef enum
     UNW_DYN_POP_FRAMES,         /* drop one or more stack frames */
     UNW_DYN_LABEL_STATE,        /* name the current state */
     UNW_DYN_COPY_STATE,         /* set the region's entry-state */
-    UNW_DYN_ALIAS               /* get unwind info from an alias */
+    UNW_DYN_ALIAS,              /* get unwind info from an alias */
+    UNW_DYN_RESTORE_REG,        /* restores a register to the value it had in the previous frame */
+    UNW_DYN_SAVE_SP,            /* the value of SP in the previous frame is found as BP + displacement,
+                                   where displacement is the accumulated value from previous UNW_DYN_ADD
+                                   directives on SP. This displacement is needed on x86_64, because the
+                                   value of %rsp saved in %rbp is _not_ exactly the value from the
+                                   previous frame, but rather slightly lower. */
   }
 unw_dyn_operation_t;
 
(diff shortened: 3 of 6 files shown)


KJTsanaktsidis:
Hm. It's failing to build for some archs because they don't define a register for `UNW_TDEP_BP`. This is easy to do for some archs that have standard frame pointer registers (`x29` on aarch64, `ebp` on x86, etc), but I realised some archs don't define a standard frame pointer register at all?

I flicked through the Itanium architecture doc (this one: https://redirect.cs.umbc.edu/portal/help/architecture/24525601.pdf) and it seems to suggest that if the compiler makes use of a frame pointer, it's free to use any register as long as it's recorded in the unwind info as "psp" (previous stack pointer).

That made me wonder how `UNW_DYN_SPILL_FP_REL` could be working for ia64, and my _squinting_ suggests it works out what psp is by following ADDs/SPILLs for SP, and treating treating `UNW_DYN_SPILL_FP_REL` as a "psp-relative spill" instead of a "sp-relative spill".

Thus, _really_, what `UNW_DYN_SPILL_FP_REL` should represent perhaps is a CFA-relative spill, _not_ a SP-relative spill. This is equivalent on architectures where the frame-pointer register holds the previous frame's stack pointer value, but on x86_64 for example the frame pointer holds (previous frame stack pointer - 16); the return-address and previous rbp value are pushed onto the stack _before_ `mov %rbp, %rsp` is run.

I could change my implementation accordingly, but I wonder if introducing the CFA concept into UNW_INFO_FORMAT_DYNAMIC is wise; it kind of just becomes a bad implementation of DWARF CFI then.

PR #496, opened 2023-05-03
Add manpage for libunwind-coredump

### LINKED ISSUE #487: Missing manpage for libunwind-coredump
There is no matching manpage for the coredump remote.

bregma:
Closes #487


bertwesarg:
this reminds me that #490 also needs a new manpage. will get to it on the weekend

bregma:
This is missing the .man file. I forgot that the generated files get checked in in this instance. Maybe it would be better to just remove all the generated .man files, but that would not be backwards-compatible with any existing packagers.

PR #506, opened 2023-05-18
Make Debug() signal-safe

bregma:
`fprintf()` is not signal-safe. `write()` is signal-safe ands we assume `snprintf()` is signal-safe.


bregma:
I'm not concerned about compilers inlining the debug function. Production code shouldn't be shipping debug mode and if it was previously, it was shipping undefined behaviour (fprintf call from a signal handler). A debug build already produced very slow unwinding -- I had customer complain about unacceptable speed in a debug release I gave them for evaluation even though UNW_DEBUG_LEVEL was not set. If a compiler decides not to inline `_unw_debug()` (and they probably will) then it's the cost of using a debug build.

PR #508, opened 2023-05-20
Some fixes for FreeBSD

lemul:
Hi, could you please merge this three simple changes for FreeBSD.

### DIFF
--- a/include/libunwind-aarch64.h
+++ b/include/libunwind-aarch64.h
@@ -238,7 +238,7 @@ typedef ucontext_t unw_fpsimd_context_t;
 #include "libunwind-dynamic.h"
 
 #if defined(__FreeBSD__)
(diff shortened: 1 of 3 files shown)


lemul:
On Mon, May 22, 2023 at 12:03:55AM -0700, Bert Wesarg wrote:
> @bertwesarg commented on this pull request.
> 
> 
> 
> > -#ifdef __syscall // prefer over syscall on *BSD
> +#if defined(__FreeBSD__) // prefer over syscall on *BSD
> 
> so what about all the other *BSD now? the comment was pretty clear that this condition should match all of them
> 

well, first time it was by __BSD_VISIBLE, however NetBSD does not define it, 
OpenBSD defines __BSD_VISIBLE, but doesn’t have __syscall(). 
So, if NetBSD is supported we can add || defined _NETBSD_SOURCE

bertwesarg:
Your are right. Looks like my comment only covered FreeeBSD from the beginning. FreeBSD seems the only supported *BSD hence we can leave it. Fill try to open an PR to fix the comment soon

PR #511, opened 2023-05-23
Fix reversed logic in Debug()

bregma:
Logic was reversed when initializing the debug level, resulting in incorrectly formatted output.

Added comments of #if/#else/#endif to make following the code a little easier.


bertwesarg:
LGTM

PR #525, opened 2023-05-30
Fix test-init-local-signal for FreeBSD.

lemul:
FreeBSD libc thr_sigaction installs its own sighandler which adds two calls before our handler.

While here fix indentation of printf to match common style.


bregma:
s390x test-async-sig fails but I think that's am unrelated race condition that happens occasionally so not new.

PR #533, opened 2023-06-01
Fix test-ptrace on x86-linux target

bregma:
The unit test test-ptrace was failing on x86 targets. This was for 3 unrelated reasons.

1. `mi_mmap()` was bypassing the libc `mmap()` call and invoking `syscall()` instead. Unfortunately on x86 Linux the direct syscall is not supported because of the way parameters are passed to syscalls. Fall back to the libc `mmap()` for x86-linux.

2. The `ptrace(PTRACE_GETREGSET)` call just returns an error no x86-linux. Fall back to using `ptrace(PTRACE_GETREGS)` instead. That only works if the parameters are passed in the correct order (they were reversed). This might break on Solaris.

3. The CI test run as multilibbed tests (running 32-bit binaries on a 64-bit OS). The test itself runs a 64-bit binary (`ls`) and the elfxx.c code built into libunwind can not handle multilibbed use although it's supposed to be the fundamental architecture of the library. That's a different issue though, so in the mean time XFAIL the test in CI. When the test is run manually using a 32-bit x86 binary it passes.
```
$ tests/test-ptrace tests/.libs/test-strerror && printf "PASSED\n" || printf "FAILED\n"
PASSED
```

### DIFF
--- a/include/libunwind_i.h
+++ b/include/libunwind_i.h
@@ -221,18 +221,20 @@ do {                                            \
 static ALWAYS_INLINE void *
 mi_mmap (void *addr, size_t len, int prot, int flags, int fd, off_t offset)
 {
-#ifdef SYS_mmap
-#if defined(__FreeBSD__) // prefer over syscall on *BSD
+#if defined(SYS_mmap) && !defined(__i386__)
+  /* Where supported, bypass libc and invoke the syscall directly. */
+# if defined(__FreeBSD__) // prefer over syscall on *BSD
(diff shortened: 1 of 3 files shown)


lemul:
> * mi_mmap() was bypassing the libc mmap() call and invoking syscall() instead. Unfortunately on x86 Linux the direct syscall is not supported because of the way parameters are passed to syscalls. Fall back to the libc mmap() for x86-linux.
> *
also this code does not handle errno,
I didn’t check that errno might needed after mi_mmap

PR #536, opened 2023-06-06
Add new mi function unw_address_is_valid()

bregma:
Created a new machine-independent function `unw_address_is_valid()` to replace all the identical cut'n'paste implementations if `validate_mem()` for several architectures. Used it in several more architectures to make Ltest_mem_validate pass on more targets.

This should fix that test for all targets except for mips* where the failures are so broad that it was impossible to test and relevant changes. The mips* targets will have to be fixed in general first.


src/mi/Gaddress_validator.c
```diff
@@ -0,0 +1,349 @@
...
+bool
+unw_address_is_valid(UNUSED unw_word_t addr, UNUSED size_t len)
+{
+  Debug(1, "remote-only invoked\n");
+  return false;
+}
+
+#else /* !UNW_REMOTE_ONLY */
+
+static pthread_once_t _unw_address_validator_init_once = PTHREAD_ONCE_INIT;
```
bertwesarg, line 37: how AS-safe is this, or how can we ensure, that this is called before being in a signal context?

src/mi/Gaddress_validator.c
```diff
@@ -0,0 +1,349 @@
...
+  _unw_address_validator_initialized = ~0;
+}
+
+/* Cache of already validated addresses */
+enum { NLGA = 4 };
+
+#if defined(HAVE___CACHE_PER_THREAD) && HAVE___CACHE_PER_THREAD
+// thread-local variant
+static _Thread_local unw_word_t last_good_addr[NLGA];
+static _Thread_local int lga_victim;
```
bertwesarg, line 198: TLS variables may also trigger `malloc` in glibc, so how do we avoid to access this before being in a signal context in every thread?

src/mi/Gaddress_validator.c
bregma: Do you mean the initialization of the variable? It's done by the dynamic loader when the RW LOAD segment is loaded. I'm going to assume the dynamic loader implements dlopen() (or whatever internally implements loading an ELF file on your favourite OS) in a threadsafe manner that prevents any code from being executed before it's loaded.

If you're asking about `pthread_run_once()`, well, it's threadsafe. It doesn't matter if the thread is executing in a signal context or not.

src/mi/Gaddress_validator.c
bregma: The malloc in libc (GNU libc or otherwise) is when the DSO is loaded or the thread created. Threads do not receive signals before they're created and if a thread is doing a dlopen() directly from a signal context it's already solidly in undefined-behaviour territory. Only function-scoped thread-local storage triggers a memory allocation after load time or thread creation time, and these are file-scoped thread-locals.

Also, this code is copied and pasted from the existing x86_64, aarch64, s390x, and loongarch64 source files as a deduplication so it has presumably been proven in production for quite some time. We have unit tests that verify malloc is not called during unwinding.

src/mi/Gaddress_validator.c
bregma: The only time there would be a problem is if the first thread to run the `pthread_once()` function receives a signal while *in* that function and calls the `pthread_once()` function a second time from within the signal handler. That would potentially result in a deadlock -- in the case of GNU/Linux, it would spin forever on a futex.

Of course, the same problem already applies to `unw_init_local()` except it's far more likely to occur. My first idea was to add the initialization to `mi_init()` since that's protected by the mutex in `tdep_init()` but that quickly became a more invasive change for little benefit.

Other than just documenting the potential deadlock, can you suggest a better approach to initialization?

bertwesarg: > The only time there would be a problem is if the first thread to run the pthread_once() function receives a signal while in that function and calls the pthread_once() function

No, thats not what I have observed, when it comes to AS-unsafe functions. Most of the time, the thread does some libc stuff which in the end calls into the malloc part, and this takes a lock. If a signal is happening while holding the lock and the handler is calling into libunwind, it needs to avoid that it also calls something AS-unsafe which might also call into malloc which tries to take the same lock. Deadlock.

I don't know, if `pthread_once()` can call into `malloc()`, it does not look so but I cannot find something about AS-safe/unsafe for `pthread_once()` either.

The only thing I can comeup is, a documented way so, that all these things can be triggered voluntarily outside a signal handler. If this is just `unw_init_local()` its fine.

bregma: Latest change replaces the pthread_once() with a simple atomic flag check that should be both thread safe and signal safe and definitely does not make a call to malloc().

bertwesarg: fine with me. thanks

PR #541, opened 2023-06-12
Reduce compile-time warnings on FreeBSD/x86_64

lemul:
Unhide tdep_stash_frame from __linux__ due it’s os independent nature.


lemul:
Tested on FreeBSD -current, Ubuntu Jammy 22.04

PR #542, opened 2023-06-12
Make use of -lpthreads autodetected by configury

### LINKED ISSUE #540: libpthread not available on all targets
Although GNU/Linux and some BSDs requires `-lpthread` (or `-pthread`) to use POSIX threads, other targets do not have that library available.

Usage of that library should be detected at configuration time and only linked if available.

bregma:
Not all targets have a separate pthread library. Make it autodetected in configure.

Refactored a number of the optionally enabled things in configure.ac for the following reasons.

  1. Their dependencies are only checked when the feature is enabled
  2. reduce redundant checks
  3. Add defaults to the --help message
  4. Group related check and messages together
  5. Improve user feedback
  6. Make grammatic voice and style more consistent

There should be no functional changes to what gets built and when. Closes #540

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -26,31 +26,22 @@ LT_INIT
 AM_PROG_AS
 AM_PROG_CC_C_O
 
-dnl Checks for libraries.
-AC_CHECK_LIB(gcc_s, _Unwind_Resume)
-AC_CHECK_LIB(uca, __uc_get_grs)
-OLD_LIBS=${LIBS}
-AC_SEARCH_LIBS(dlopen, dl)
-LIBS=${OLD_LIBS}
-case "$ac_cv_search_dlopen" in
-  -l*) DLLIB=$ac_cv_search_dlopen;;
-  *) DLLIB="";;
-esac
-
-dnl Checks for header files.
-AC_CHECK_HEADERS(asm/ptrace_offsets.h asm/ptrace.h asm/vsyscall.h endian.h sys/endian.h \
-		sys/param.h execinfo.h ia64intrin.h sys/uc_access.h unistd.h signal.h \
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -110,57 +94,188 @@ SET_ARCH([$build_cpu],[build_arch])
...
+AM_CONDITIONAL([CONFIG_TESTS], [test x$enable_tests = xyes])
+AM_COND_IF([CONFIG_TESTS], [
+  old_LIBS="$LIBS"
+  AC_MSG_NOTICE([--- Checking for extra libraries linked to tests ---])
+  AC_SEARCH_LIBS([dlopen], [dl],
+                 [AS_IF([test "$ac_cv_search_dlopen" != "none required"],
+                        [AC_SUBST([DLLIB], ["$ac_cv_search_dlopen"])])])
+  AC_SEARCH_LIBS([pthread_key_create], [pthread],
+                 [AS_IF([test "$ac_cv_search_pthread_key_create" != "none required"],
+                        [AC_SUBST([PTHREADS_LIB], ["$ac_cv_search_pthread_key_create"])])])
```
bertwesarg, line 220: why not using https://www.gnu.org/software/autoconf-archive/ax_pthread.html?

bregma: ax_pthread.m4 is almost larger than configure.ac. This project currently has no infrastructure in place to handle third-party autoconf macro files (not that that is hard to set up) and the checks in ax_pthread.m4 seem like overkill for the purpose here, which is a few unit tests. Its primary purpose if to handle obscure edge cases on operating systems we don't even support, like AIX and Digital Unix. 

It's a good suggestion worth considering, but on the whole I think a simple test for whether `-lpthread` is necessary or not will improve portability while minimally impacting maintainability and build times.

PR #543, opened 2023-06-12
Some simple improvements to the x64-unwind-badjmp-signal-frame test for FreeBSD/Linux

lemul:
Please, merge. Thanks !


bregma:
Neither <dlfcn.h> nor `dladdr()` are portable. Are they necessary for the test? They don't seem to affect the result at all.

lemul:
> Neither <dlfcn.h> nor `dladdr()` are portable. Are they necessary for the test? They don't seem to affect the result at all.

yes, no functional changes here, it's only helps porting library for others.
are you ok if I guard this changes by #if 0 or even by #ifdef __FreeBSD__?

PR #547, opened 2023-06-13
Fix test-ptrace for FreeBSD

lemul:


lemul:
After the sleep, an idea came up on how best to implement this

PR #548, opened 2023-06-14
Improve ptrace test portability

lemul:


tests/test-ptrace.c
```diff
@@ -343,7 +346,7 @@ main (int argc, char **argv)
 	  if (!state)
 	    do_backtrace ();
 	  state ^= 1;
-#if HAVE_DECL_PTRACE_SYSCALL
+#if HAVE_DECL_PTRACE_SYSCALL && !HAVE_DECL_PT_SYSCALL
```
bregma, line 349: Would it not be better to just reverse the order if ifdefs?
```
#if defined(HAVE_DECL_PT_SYSCALL)
    ptrace (PT_SYSCALL, target_pid, (caddr_t)1, pending_sig);
#elif defined(HAVE_DECL_PTRACE_SYSCALL)
    ptrace (PTRACE_SYSCALL, target_pid, 0, pending_sig);
#else
...
```

PR #553, opened 2023-06-19
Unconditionally add pthread library on FreeBSD.

lemul:
On FreeBSD using AC_SEARCH_LIBS is not sufficient for finding the linker flags for linking with a pthread library.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -215,9 +215,15 @@ AM_COND_IF([CONFIG_TESTS], [
   AC_SEARCH_LIBS([dlopen], [dl],
                  [AS_IF([test "$ac_cv_search_dlopen" != "none required"],
                         [AC_SUBST([DLLIB], ["$ac_cv_search_dlopen"])])])
-  AC_SEARCH_LIBS([pthread_key_create], [pthread],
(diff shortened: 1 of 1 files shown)


bregma (changes requested):
I don't think this is going to do what you think it's going to do. There is no $OS_FREEBSD variable defined anywhere and there is no AC_SUBST() invocation.

Might I suggest this instead?
```
-  AC_SEARCH_LIBS([pthread_key_create], [pthread],
-                 [AS_IF([test "$ac_cv_search_pthread_key_create" != "none required"],
-                        [AC_SUBST([PTHREADS_LIB], ["$ac_cv_search_pthread_key_create"])])])
+  AS_IF([expr x$target_os : xfreebsd >/dev/null],
+        [ac_cv_search_pthread_key_create="-lpthread"],
+        [AC_SEARCH_LIBS([pthread_key_create], [pthread])])
+  AS_IF([test "$ac_cv_search_pthread_key_create" != "none required"],
+        [AC_SUBST([PTHREADS_LIB], ["$ac_cv_search_pthread_key_create"])])
```

PR #555, opened 2023-06-20
Added support for unwinding through PPC64 PLT entries.

kenche-linaro:
Added function to detect when specified IP is within a PLT entry. This is determined by reading surrounding instructions and matching them to known PPC64 PLT procedure instructions.

Added PPC64-only unittest for is_plt_entry. Mock instructions are set up for testing, using different offsets and non-PLT replacements.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,6 +78,8 @@ tests/[GL]ia64-test-readonly
 tests/[GL]ia64-test-stack
 tests/ia64-test-dyn1
 tests/ia64-test-sig
+tests/ppc64-test-altivec
+tests/ppc64-test-plt
 tests/[GL]x64-test-dwarf-expressions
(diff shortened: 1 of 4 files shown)


kenche-linaro:
> This is failing CI for ppc64 targets.
> 
> ```
>  /usr/lib/gcc-cross/powerpc64-linux-gnu/12/../../../../powerpc64-linux-gnu/bin/ld: ../src/.libs/libunwind-ppc64.a(Gget_accessors.o):(.opd+0x0): multiple definition of `_Uppc64_get_accessors_int'; ppc64-test-plt.o:(.opd+0x48): first defined here
> collect2: error: ld returned 1 exit status
> ```

Thanks. That should be resolved now.

bregma:
Why does the new `ppc64-test-plt` fail in CI?

PR #556, opened 2023-06-20
Cache synthetic eh_frame_hdr.

kenche-linaro:
Cache value of eh_frame_hdr to avoid having to open and find it within the same file again. The values of eh_frame_hdr are stored in a singly-linked list along with their respective filenames.


bregma:
This change is not threadsafe.

kenche-linaro:
> This change is not threadsafe.

Thanks. This should be resolved now.

PR #567, opened 2023-06-30
Introduce a new OS specific unw_step() function on x86_64.

lemul:
And move out OS specific code to unw_step().
For Linux its a vsyscall handling, for Solaris is no-op for now, and for FreeBSD its a handling of syscall calling sequence. Thats fixes FreeBSD unw_is_signal_frame() which is incorrectly returns true for syscall frame.

### DIFF
--- a/src/x86_64/Gos-freebsd.c
+++ b/src/x86_64/Gos-freebsd.c
@@ -38,7 +38,7 @@ unw_is_signal_frame (unw_cursor_t *cursor)
 {
   /* XXXKIB */
   struct cursor *c = (struct cursor *) cursor;
-  unw_word_t w0, w1, w2, b0, ip;
+  unw_word_t w0, w1, w2, ip;
   unw_addr_space_t as;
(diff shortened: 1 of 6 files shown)


bregma:
Fails to build from source on QNX.
```
86_64/.libs/Lstep.o: In function `_ULx86_64_step':
/home/swebb/projects/libunwind/pr-657/qnx-x86_64/src/../../src/x86_64/Gstep.c:125: undefined reference to `_ULx86_64_os_step'
/home/swebb/sdp_700/host/linux/x86_64/usr/bin/x86_64-pc-nto-qnx7.0.0-ld: .libs/libunwind.so.8: hidden symbol `_ULx86_64_os_step' isn't defined
/home/swebb/sdp_700/host/linux/x86_64/usr/bin/x86_64-pc-nto-qnx7.0.0-ld: final link failed: Bad value
```

bregma (approved):
OK.

PR #568, opened 2023-06-30
Remove tilegx as a target

### LINKED ISSUE #563: Remove support for tilegx
The tilegx architecture is no longer supported by toolchain vendors and we have no way to verify it works. We need to remove support for it to avoid bitrotting. Those who desperately still need libunwind for tilex can use release 1.7 or earlier.

bregma:
This target is no longer supported by common toolchains or operating systems. If support is required, it can still be found in libunwind version 1.7 and earlier.

Closes #563


bregma:
There's been no public comment on this and a sincerely doubt anyone will oppose it. Committing.

PR #570, opened 2023-07-01
Fix uc_addr() for FreeBSD/Aarch64.

lemul:
On FreeBSD the Neon registers are preserved in the mc_fpregs of the machine context structure.


bregma (approved):
Looks sane.

Issue #557, opened 2023-06-21
run-ptrace-mapper and run-ptrace-misc testcases fail on aarch64 and possible fix


ziyangc97:
The following make check ptrace-related tests fail on openEuler operating system aarch64.

**FAIL: run-ptrace-mapper
FAIL: run-ptrace-misc**

Below is tests/test-suite.log
```
FAIL: run-ptrace-mapper
=======================

unw_get_proc_info(ip=0xaaaae6120810) failed: ret=-10
unw_get_proc_info(ip=0xaaaae6120814) failed: ret=-10
unw_get_proc_info(ip=0xaaaae6120818) failed: ret=-10
unw_get_proc_info(ip=0xaaaae612081c) failed: ret=-10
child terminated by signal 14
FAILURE: detected 5 errors
FAIL run-ptrace-mapper (exit status: 255)

FAIL: run-ptrace-misc
=====================

unw_get_proc_info(ip=0xaaaac5080830) failed: ret=-10
unw_get_proc_info(ip=0xaaaac5080834) failed: ret=-10
unw_get_proc_info(ip=0xaaaac5080838) failed: ret=-10
unw_get_proc_info(ip=0xaaaac508083c) failed: ret=-10
unw_get_proc_info(ip=0xaaaac50807f0) failed: ret=-10
unw_get_proc_info(ip=0xaaaac50807f4) failed: ret=-10
unw_get_proc_info(ip=0xaaaac50807f8) failed: ret=-10
unw_get_proc_info(ip=0xaaaac50807fc) failed: ret=-10
FAILURE: detected 8 errors
FAIL run-ptrace-misc (exit status: 255)

```

From the test-suite.log, we found that unw_get_proc_info print error messge and return -10 which is UNW_ENOINFO(no unwind info found).
By comparing the implementation of unw_get_proc_info() between aarch64 and x86-64 architecture, we discover that in x86-64 the unw_get_proc_info() has following code which promise the test case won't fail
```
  if (dwarf_make_proc_info (&c->dwarf) < 0)
    {
      /* On x86-64, some key routines such as _start() and _dl_start()
         are missing DWARF unwind info.  We don't want to fail in that
         case, because those frames are uninteresting and just mark
         the end of the frame-chain anyhow.  */
      memset (pi, 0, sizeof (*pi));
      pi->start_ip = c->dwarf.ip;
      pi->end_ip = c->dwarf.ip + 1;
      return 0;
    }
  *pi = c->dwarf.pi;
[...]

PR #578, opened 2023-07-07
tests/mapper: avoid calling printf between SIGUSR1 and SIGUSR2

he7850:
fix https://github.com/libunwind/libunwind/issues/557 test-ptrace-mapper testcase fail.

glibc printf() needs using atomic instruction to acquire the stdout lock.
Under ARMv8.0 ISA, atomic instruction is realized using LL-SC routine, which will keep retrying if other cpu accesses the target memory during the atomic instruction.
In mapper testcase, parent process uses ptrace to single step mapper program, thus if calling printf between SIGUSR1 and SIGUSR2, printf will be single-step executed, and the atomic instruction will be stuck in deadloop.

### DIFF
--- a/tests/mapper.c
+++ b/tests/mapper.c
@@ -43,6 +43,15 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 # define MAP_NORESERVE 0
 #endif
 
+void __attribute__((noinline)) push_some_stacks(int n)
+{
+  if (n >= 1)
+  {
+    push_some_stacks(n - 1);
+    push_some_stacks(n - 1);
+  }
+}
+
 int
 main (void)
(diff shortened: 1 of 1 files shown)


bregma:
CI is failing on x86_64.
```
FAIL: run-ptrace-mapper
=======================

child terminated by signal 14
FAILURE: detected 1 errors
FAIL run-ptrace-mapper (exit status: 255)
```

he7850:
> CI is failing on x86_64.
> 
> ```
> FAIL: run-ptrace-mapper
> =======================
> 
> child terminated by signal 14
> FAILURE: detected 1 errors
> FAIL run-ptrace-mapper (exit status: 255)
> ```

strange, it passes in my pc machine, maybe need to remove some stacks for ci machine

Issue #569, opened 2023-06-30
test-async-sig fails on ubuntu20.04_x86 with -O2


bregma:
This test fails at `-O2` but passes at `-O0`, both GCC and clang.

PR #582, opened 2023-07-19
Bump to release 1.7.1

### LINKED ISSUE #535: 1.7.0 package name and version in pkg-config files are broken
First, thanks for the release, but the 1.7.0 package name is broken, it reads `libunwind-1.70.tar.gz`.

The reason is the missing `.` between `pkg_minor` and `pkg_suffix` in `mkvers`. This was changed in the 1.6 release branch 9a0de16192977b609691e824c900b628d623510b.

bregma:
Fixes format of release number string for proper semantic versioning.

Closes #535

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2,9 +2,9 @@ dnl Process this file with autoconf to produce a configure script.
 
 define(pkg_major, 1)
 define(pkg_minor, 7)
-define(pkg_extra, 0)
+define(pkg_extra, 1)
 define(pkg_maintainer, libunwind-devel@nongnu.org)
-define(mkvers, $1.$2$3)
+define(mkvers, $1.$2.$3)
 
 AC_INIT([libunwind],[mkvers(pkg_major, pkg_minor, pkg_extra)],[pkg_maintainer])
 AC_CONFIG_SRCDIR(src/mi/backtrace.c)


lemul:
+1, it would be nice to new release due to many fixes atleast for FreeBSD. 10x

lemul:
> > it would be nice to new release due to many fixes atleast for FreeBSD.
> 
> I'm aiming for a 1.8 release in the August timeframe to pick up the QNX and FreeBSD changes. It's not quite the yearly cadence of previous releases but there's been more activity lately.
> 
That's a great plan for me,
from the FreeBSD perspective all required changes are commited into the main branch

> The big thing that's still outstanding is getting the (Linux) CI green(er).
You mean those weird errors with -O3 -O0 on x86?

bregma:
> You mean those weird errors with -O3 -O0 on x86?

Exactly. #569 seems to be a problem with signal trampoline detection that only shows up in optimized code, which kind of smells like undefined behaviour somewhere, but nothing shows up with ubsan. It's a little disconcerting.

PPC, MIPS, and 32-bit ARM may just need to be demoted to second-line support targets since they've been broken for a long time unless someone steps up.

lemul:
> > You mean those weird errors with -O3 -O0 on x86?
> 
> Exactly. #569 seems to be a problem with signal trampoline detection that only shows up in optimized code, which kind of smells like undefined behaviour somewhere, but nothing shows up with ubsan. It's a little disconcerting.
> 

-1 from unw_step() is a UNW_UNSPEC? However, I dont see where on x86 Linux UNW_UNSPEC is returned, so it seems that -1 is result from some libc call.

> PPC, MIPS, and 32-bit ARM may just need to be demoted to second-line support targets since they've been broken for a long time unless someone steps up.

PR #598, opened 2023-08-10
Use pthread_sigmask instead of sigprocmask

kenche-linaro:
pthread_sigmask use in multithreaded programs is explicitly specified by POSIX.1.


bregma:
Passes check on QNX x86_6 and aarch64 and no new regression in CI. Good to go.

PR #599, opened 2023-08-21
Improved AArch64 is_plt_entry implementation

kenche-linaro:
Previous implementation could only handle cases where IP was at the first instruction in the PLT. This change implements detecting PLT entries regardless of where the current IP is within the entry.

Added AArch64-only unittest for is_plt_entry. Mock instructions are set up for testing, using different offsets and non-PLT replacements.


bregma:
Nice. Verified manually on QNX with no regression.

PR #602, opened 2023-08-24
Fallback to direct access instead of through /proc/[pid]/root/

happyCoder92:
In some namespace setups the binary might not even be present in the mnt_ns. This is the case if it's launched from an fd in a specially prepared empty mnt_ns via a execveat (sandboxing usecase).

Therefore it makes sense to fallback to the original path.

### DIFF
--- a/src/os-linux.c
+++ b/src/os-linux.c
@@ -84,6 +84,9 @@ tdep_get_elf_image (struct elf_image *ei, pid_t pid, unw_word_t ip,
   strcpy (full_path, root);
   strcat (full_path, mi.path);
 
+  if (stat(full_path, &st) || !S_ISREG(st.st_mode))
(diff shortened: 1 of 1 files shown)


bregma:
It seems to me this will *always* revert to the original path since `/proc/PID/root` is usually a symbolic link so `!S_ISREG(st.st_mode)` will always be true?

happyCoder92:
It's `stat` and not `lstat`, but yeah `/proc/PID/root` will likely be a dir and that's checked already above (line 73).
This is about the full path so `/proc/PID/root/path/to/my/executable/exec_file`, so expecting it to be a file is fine I believe.

bregma:
Ah, OK, I was misreading the preceding code. This is fine.

PR #605, opened 2023-08-25
Add support to look for debug libraries via the build-id

robertwhitton:
Candidate changes to add support for build-id in order to find the debug libraries. If if fails to find the debug libraries via build-id then it should fall-back to the legacy debuglink approach. Perhaps controversially I've renamed load_debuglink to load_debuglib in order to reflect that this function no longer just handles the debuglink mechanism.

### DIFF
--- a/src/dwarf/Gfind_proc_info-lsb.c
+++ b/src/dwarf/Gfind_proc_info-lsb.c
@@ -120,7 +120,7 @@ load_debug_frame (const char *file, char **buf, size_t *bufsize, int is_local,
   ei.image = NULL;
   *load_offset = 0;
 
-  ret = elf_w (load_debuglib) (file, &ei, is_local);
+  ret = elf_w (load_debuginfo) (file, &ei, is_local);
   if (ret != 0)
     return ret;
 
--- a/src/dwarf/Gfind_proc_info-lsb.c
+++ b/src/dwarf/Gfind_proc_info-lsb.c
@@ -120,7 +120,7 @@ load_debug_frame (const char *file, char **buf, size_t *bufsize, int is_local,
   ei.image = NULL;
   *load_offset = 0;
 
-  ret = elf_w (load_debuglink) (file, &ei, is_local);
+  ret = elf_w (load_debuglib) (file, &ei, is_local);
   if (ret != 0)
     return ret;
 
--- a/src/elfxx.c
+++ b/src/elfxx.c
@@ -35,7 +35,7 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 #endif /* HAVE_LZMA */
 
 static Elf_W (Shdr)*
-elf_w (section_table) (struct elf_image *ei)
+elf_w (section_table) (const struct elf_image *ei)
 {
   Elf_W (Ehdr) *ehdr = ei->image;
   Elf_W (Off) soff;
@@ -53,7 +53,7 @@ elf_w (section_table) (struct elf_image *ei)
 }
 
 static char*
-elf_w (string_table) (struct elf_image *ei, int section)
+elf_w (string_table) (const struct elf_image *ei, int section)
 {
   Elf_W (Ehdr) *ehdr = ei->image;
   Elf_W (Off) soff, str_soff;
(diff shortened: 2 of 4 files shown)


src/elfxx.c
```diff
@@ -445,6 +445,91 @@ elf_w (find_section) (struct elf_image *ei, const char* secname)
...
+
+      while(notes < notes_end)
+        {
+          const char prefix[] = "/usr/lib/debug/.build-id/";
+
+          /* See "man 5 elf" for notes about alignment in Nhdr */
+          const Elf_W(Nhdr *) nhdr = (const ElfW(Nhdr *)) notes;
+          const ElfW(Word) namesz = nhdr->n_namesz;
+          const ElfW(Word) descsz = nhdr->n_descsz;
+          const ElfW(Word) nameasz = UNW_ALIGN(namesz, 4); /* Aligned size */
```
bregma, line 496: This needs to be
```
     const Elf_W(Nhdr)* nhdr = (const Elf_W(Nhdr)*) notes;
     const Elf_W(Word) namesz = nhdr->n_namesz;
     const Elf_W(Word) descsz = nhdr->n_descsz;
     const Elf_W(Word) nameasz = UNW_ALIGN(namesz, 4); /* Aligned size */
```

src/elfxx.c
```diff
@@ -445,6 +445,91 @@ elf_w (find_section) (struct elf_image *ei, const char* secname)
...
+          const char *name = (const char *) (nhdr + 1);
+          const uint8_t *desc = (const uint8_t *) name + nameasz;
+          unsigned j;
+
+          notes += sizeof(*nhdr) + nameasz + UNW_ALIGN(descsz, 4);
+
+          if ((namesz != sizeof(ELF_NOTE_GNU)) ||  /* Spec says must be "GNU" with a NULL */
```
bregma, line 503: ELF_NOTE_GNU is not defined non-GNU platforms. All of this code needs to be wrapped in `#if defined(ELF_NOTE_GNU)`.

src/elfxx.c
```diff
@@ -445,6 +445,91 @@ elf_w (find_section) (struct elf_image *ei, const char* secname)
...
+  unsigned i;
+
+  if (!elf_w (valid_object) (ei))
+    return -1;
+
+  shdr = elf_w (section_table) (ei);
+  if (!shdr)
+    return -1;
+
+  for (i = 0; i < ehdr->e_shnum; ++i, shdr = (const Elf_W (Shdr) *) (((const uint8_t *) shdr) + ehdr->e_shentsize))
```
bregma, line 476: Section headers could be completely stripped out (think core dumps or embedded images). You're better off looking for a PT_NOTE segment through the phdrs since that's guaranteed to be present.

bregma (changes requested):
Renaming `load_debuglink` to something else is good since it's no longer just looking for the `.gnu_debuglink` section but since it's not a library (GNU binutils refers to it as "separate debug info") I would prefer `load_debuginfo`. That covers all the cases of unstripped executables as wells as separate debug files for libraries and executables.

robertwhitton:
I'll take a look at these comments. They all look reasonable to me.

src/elfxx.c
robertwhitton: I wonder when ELF_NOTE_GNU and NT_GNU_BUILD_ID were added to glibc? A bit of digging suggests that NT_GNU_BUILD_ID didn't appear until 2.7 (ELF_NOTE_GNU is earlier). Is there a minimum version of glibc header files that libunwind supports? I can fix this another way if necessary.

bregma: It should be enough to just `#ifdef` directly on those macros since they could be supplied by implementations other than GNU for compatibility.

src/dwarf/Gfind_proc_info-lsb.c
```diff
@@ -120,7 +120,7 @@ load_debug_frame (const char *file, char **buf, size_t *bufsize, int is_local,
   ei.image = NULL;
   *load_offset = 0;
 
-  ret = elf_w (load_debuglink) (file, &ei, is_local);
+  ret = elf_w (load_debuglib) (file, &ei, is_local);
```
bregma, line 123: Missed this change?

bregma (approved):
I still get
```
../../src/elfxx.h:42:25: warning: '_Uelf64_add_hex_byte' defined but not used [-Wunused-function]
   42 | # define elf_w(x)       _Uelf64_##x
```
building for QNX but I'm inclined to let that go for now.

Issue #603, opened 2023-08-24
libunwind QNX aarch64 lose some stack frame compared with QNX gdb


iDings:
QNX's bt_get_backtrace api is totally broken on aarch64, really glad to see that libunwind support QNX platform, thanks :+1: 

I encountered an issue on QNX aarch64 libunwind:
Problem: 
   when use libunwind nto api([example](https://gist.github.com/iDings/ab547a17cf4d153f58e7183115a74eaf)) to do remote unwind, the output seems success, but when compared with QNX's gdb output, found that an stack frame is losed. 
   90118 is syslogd's pid, 1 is tid of syslogd 
Case 1:
   libunwind's output:
   ```
    # pidin ar | grep syslogd
      90118 syslogd -f /system/etc/syslog.conf 

    # ./libunwind_example_remote 90118 1 
    0x2864a67cd0: (SignalWaitinfoMask+0x8)
    0x2864aa121c: (pselect+0x164)
    0x2864aa14e8: (select+0x48)
    0x1fded855b0: (main+0x2e0)
  ```
  use QNX's gdb remote attach and backtrace, output:
```
(gdb) target qnx 172.31.1.110:8000
Remote debugging using 172.31.1.110:8000
MsgNak received - resending
Remote target is little-endian
(gdb) attach 90118
Attaching to pid 90118 tid 1
[New pid 90118 tid 1]
[New pid 90118 tid 2]
[New pid 90118 tid 3]
[New pid 90118 tid 4]
[New pid 90118 tid 5]
SignalWaitinfoMask () at /builds/workspace/710-SDP/build_aarch64/lib/c/kercalls/aarch64/SignalWaitinfoMask.S:8
8       /builds/workspace/710-SDP/build_aarch64/lib/c/kercalls/aarch64/SignalWaitinfoMask.S: No such file or directory.
(gdb) bt
#0  SignalWaitinfoMask () at /builds/workspace/710-SDP/build_aarch64/lib/c/kercalls/aarch64/SignalWaitinfoMask.S:8
[...]

PR #607, opened 2023-08-31
Fix lose caller frame of kernel call on QNX aarch64

iDings:
QNX aarch64 kernel call without frame pointer and cfi directives, this will fallback to use frame pointer unwinder, but the frame pointer is caller's frame pointer, causing caller frame be skipped.

Try to sniff kernel call pattern, get procedure's ip range by symtab and do the matchup from the beginning of procedure. If matched, then fallback to use link register unwinder.

This patch try to fix https://github.com/libunwind/libunwind/issues/603#issue-1865190934.

### DIFF
--- a/src/elfxx.c
+++ b/src/elfxx.c
@@ -34,6 +34,34 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 #include <lzma.h>
 #endif /* HAVE_LZMA */
 
+struct symbol_info
+{
+  const char *strtab;
+  const Elf_W (Sym) *sym;
+  Elf_W (Addr) start_ip;
+};
+
+struct symbol_lookup_context
+{
+  unw_addr_space_t as;
+  unw_word_t ip;
+  struct elf_image *ei;
+  Elf_W (Addr) load_offset;
+  Elf_W (Addr) *min_dist;
+};
+
+struct symbol_callback_data
+{
+  char *buf;
+  size_t buf_len;
+};
+
+struct ip_range_callback_data
+{
+  Elf_W (Addr) *start_ip;
+  Elf_W (Addr) *end_ip;
+};
+
 static Elf_W (Shdr)*
 elf_w (section_table) (const struct elf_image *ei)
 {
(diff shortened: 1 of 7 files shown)


src/elfxx.c
```diff
@@ -154,9 +155,104 @@ elf_w (lookup_symbol) (unw_addr_space_t as UNUSED,
...
+                         Elf_W (Addr) real_start_ip)
+{
+  int ret = -UNW_ENOINFO;
+
+  struct {
+    unw_word_t ip;
+    char *buf;
+    size_t buf_len;
+    Elf_W (Addr) *min_dist;
+  } *arg = opaque;
```
bregma, line 174: I'm not sure I understand why this is passed as an opaque pointer with the struct redefined in several places. Since these are all static functions would it not be better to declare the args struct once and pass it explicitly?

src/elfxx.c
```diff
@@ -401,6 +497,66 @@ elf_w (get_proc_name) (unw_addr_space_t as, pid_t pid, unw_word_t ip,
...
+  unsigned long segbase, mapoff;
+  struct elf_image ei;
+  int ret;
+  char file[PATH_MAX];
+
+  ret = tdep_get_elf_image (&ei, pid, ip, &segbase, &mapoff, file, PATH_MAX);
+  if (ret < 0)
+    return ret;
+
+  ret = elf_w (load_debuglink) (file, &ei, 1);
```
bregma, line 548: This function name just changed (#605).

bregma (commented):
This is good -- some of the new functions could be re-used elsewhere for consistency (and improved readability) but it's not required and can be done later.

iDings:
Updated, look forward to receive your suggestions.

And whether do we need an unified GNU code style description, then can use clang-format to make the code style coincide?
I see some different code style in the source. 
https://github.com/libunwind/libunwind/blob/c1c88dacbb246d17324d8fcc1afaa2a5536a8f33/src/arm/Gtrace.c#L148-L152
https://github.com/libunwind/libunwind/blob/c1c88dacbb246d17324d8fcc1afaa2a5536a8f33/src/elfxx.c#L282-L287

bregma:
This looks great.

One thing I've learned over more than 40 years as a professional software developer is that style really doesn't matter as long as the code is readable. I try to make any new code use the GNU style and and changes to existing code follow the existing style but I don't think it's worth it to bring existing code into line or even to insist on it from contributors.

PR #608, opened 2023-08-31
Fix unaligned memory accesses in */Ginit.c

### LINKED ISSUE #340: Apparent UB in `access_mem` and maybe others
We just tried running `UndefinedBehaviorSanitizer` in our build with `tcmalloc` using `libunwind` as the unwinder, which results in this report:

```
[j0] src/third_party/unwind/dist/src/x86_64/Ginit.c:334:14: runtime error: load of misaligned address 0x7fabeb9d70ca for type 'unw_word_t' (aka 'unsigned long'), which requires 8 byte alignment
--
[j0] 0x7fabeb9d70ca: note: pointer points here
[j0]  66 f5  00 00 48 8d 15 cf f8 00  00 4c 89 ec 41 ff e4 66  0f 1f 84 00 00 00 00 00  83 47 04 01 c3 90
[j0]               ^
[j0]     #0 0x7fabe8807bc1 in access_mem /home/katwu/mongodb/mongo/src/third_party/unwind/dist/src/x86_64/Ginit.c:334:14
[j0]     #1 0x7fabe8809ada in is_plt_entry /home/katwu/mongodb/mongo/src/third_party/unwind/dist/src/x86_64/Gstep.c:43:14
[j0]     #2 0x7fabe8809ada in _ULx86_64_step /home/katwu/mongodb/mongo/src/third_party/unwind/dist/src/x86_64/Gstep.c:126
[j0]     #3 0x7fabeb6c958c in GetStackTrace_libunwind(void**, int, int) /home/katwu/mongodb/mongo/src/third_party/gperftools/dist/src/stacktrace_libunwind-inl.h:138:9
[j0]     #4 0x7fabeb6c9819 in maybe_convert_libunwind_to_generic_fp() /home/katwu/mongodb/mongo/src/third_party/gperftools/dist/src/stacktrace.cc:353:12
[j0]     #5 0x7fabeb6cada6 in init_default_stack_impl() /home/katwu/mongodb/mongo/src/third_party/gperftools/dist/src/stacktrace.cc:385:3
[j0]     #6 0x7fabeb6b41fb in (anonymous namespace)::google_init_module_stacktrace_init_default_stack_impl() /home/katwu/mongodb/mongo/src/third_party/gperftools/dist/src/stacktrace.cc:396:1
[j0]     #7 0x7fabeb6b41fb in GoogleInitializer::GoogleInitializer(char const*, void (*)(), void (*)()) /home/katwu/mongodb/mongo/src/third_party/gperftools/dist/src/base/googleinit.h:46
[j0]     #8 0x7fabeb6b41fb in __cxx_global_var_init /home/katwu/mongodb/mongo/src/third_party/gperftools/dist/src/stacktrace.cc:396
[j0]     #9 0x7fabeb6b41fb in _GLOBAL__sub_I_stacktrace.cc /home/katwu/mongodb/mongo/src/third_party/gperftools/dist/src/stacktrace.cc
[j0] Waiting to connect to mongod on port 20000.
[j0]     #10 0x7fabeb9e6732 in call_init /build/glibc-OTsEL5/glibc-2.27/elf/dl-init.c:72
[j0]     #11 0x7fabeb9e6732 in _dl_init /build/glibc-OTsEL5/glibc-2.27/elf/dl-init.c:119
[j0]     #12 0x7fabeb9d70c9  (/lib64/ld-linux-x86-64.so.2+0x10c9)
```

At least on the surface, this appears to be a valid because nothing seems to ensure that `addr` is aligned to be dereferenced as an `unsigned long`. For example:
* https://github.com/10gen/mongo/blob/0c818a98dc5634a1536818d44f23573f19e3ffbf/src/third_party/unwind/dist/src/x86_64/Ginit.c#L334
* https://github.com/10gen/mongo/blob/0c818a98dc5634a1536818d44f23573f19e3ffbf/src/third_party/unwind/dist/src/x86_64/Ginit.c#L355
* https://github.com/10gen/mongo/blob/0c818a98dc5634a1536818d44f23573f19e3ffbf/src/third_party/unwind/dist/src/x86_64/Ginit.c#L360
* https://github.com/10gen/mongo/blob/0c818a98dc5634a1536818d44f23573f19e3ffbf/src/third_party/unwind/dist/src/x86_64/Ginit.c#L387
* https://github.com/10gen/mongo/blob/0c818a98dc5634a1536818d44f23573f19e3ffbf/src/third_party/unwind/dist/src/x86_64/Ginit.c#L391

We are unsure how to address this failure because we don't know whether there is reason to assume these should result in aligned memory accesses. In any case, it clearly does not always, so we assume that whatever precondition was assumed has been violated somehow. What would be the best course of action for correcting this report?

### LINKED ISSUE #530: x86_64 possible misalignment issue
UBSan raises a possible misalignment issue on x86_64-linux targets.

Session log of unit test **Gtest-bt** follows.
```
./../src/x86_64/Ginit.c:358:14: runtime error: load of misaligned address 0x555555559687 for type 'unw_word_t', which requires 8 byte alignment
0x555555559687: note: pointer points here
 00 00 ff d0 83  3d 3e 48 00 00 00 0f 85  ac 01 00 00 bf 00 00 10  00 e8 c2 dc ff ff 48 89  04 24 48
             ^ 

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7ebac50 in ?? () from /lib/x86_64-linux-gnu/libgcc_s.so.1
(gdb) bt
#0  0x00007ffff7ebac50 in ?? () from /lib/x86_64-linux-gnu/libgcc_s.so.1
#1  0x00007ffff7ebc77b in _Unwind_Backtrace () from /lib/x86_64-linux-gnu/libgcc_s.so.1
#2  0x00007ffff75686f8 in __sanitizer::BufferedStackTrace::UnwindSlow (this=0x7fffffffc170, pc=140737353575790, max_depth=<optimized out>)
    at ../../../../src/libsanitizer/sanitizer_common/sanitizer_unwind_linux_libcdep.cpp:130
#3  0x00007ffff7545480 in ubsan_GetStackTrace (stack=stack@entry=0x7fffffffc170, max_depth=max_depth@entry=256, pc=140737353575790, bp=140737488342288, context=context@entry=0x0, 
    fast=<optimized out>) at ../../../../src/libsanitizer/ubsan/ubsan_diag.cpp:41
#4  0x00007ffff7545519 in MaybePrintStackTrace (pc=<optimized out>, bp=<optimized out>) at ../../../../src/libsanitizer/sanitizer_common/sanitizer_flags.h:38
#5  0x00007ffff75462f2 in __ubsan::ScopedReport::~ScopedReport (this=0x7fffffffca80, __in_chrg=<optimized out>) at ../../../../src/libsanitizer/ubsan/ubsan_diag.cpp:389
#6  0x00007ffff7549a21 in handleTypeMismatchImpl (Data=Data@entry=0x7ffff7fa6840, Pointer=Pointer@entry=93824992253575, Opts=...) at ../../../../src/libsanitizer/ubsan/ubsan_handlers.cpp:92
#7  0x00007ffff754ad45 in __ubsan::__ubsan_handle_type_mismatch_v1 (Data=Data@entry=0x7ffff7fa6840, Pointer=Pointer@entry=93824992253575) at ../../../../src/libsanitizer/ubsan/ubsan_handlers.cpp:99
#8  0x00007ffff7f76d6e in access_mem (as=<optimized out>, addr=93824992253575, val=0x7fffffffcdc0, write=<optimized out>, arg=<optimized out>) at ../../src/x86_64/Ginit.c:348
#9  0x00007ffff7f7ae42 in dwarf_get (c=0x55555555d700 <cursor>, loc=..., val=0x7fffffffcdc0) at ../../include/tdep-x86_64/libunwind_i.h:205
#10 0x00007ffff7f7bfc6 in _Ux86_64_step (cursor=0x55555555d700 <cursor>) at ../../src/x86_64/Gstep.c:218
#11 0x0000555555557629 in do_backtrace () at ../../tests/Gtest-bt.c:109
#12 0x00005555555579e9 in segv_handler (signal=<optimized out>, siginfo=<optimized out>, context=<optimized out>) at ../../tests/Gtest-bt.c:176
#13 <signal handler called>
#14 0x0000000000000123 in ?? ()
#15 0x0000555555559687 in main (argc=<optimized out>, argv=<optimized out>) at ../../tests/Gtest-bt.c:262
```

sfc-gh-sgiesecke:
This was detected by UBSan:
```
/worker/build/66bf5b2131e0cc44/root/external/libunwind/src/aarch64/Ginit.c:362:14: runtime error: load of misaligned address 0xffff7ffd376c for type 'unw_word_t', which requires 8 byte alignment
```

Partial stack trace in v1.7.2:
```
#5  0x000000000044ad14 in access_mem (as=0x570638 <local_addr_space>, addr=281474809427820, val=0xffffffffcf38, write=0, arg=0xffffffffd420) at /worker/build/66bf5b2131e0cc44/root/external/libunwind/src/aarch64/Ginit.c:362
#6  0x000000000044b7fc in _Uaarch64_is_signal_frame (cursor=0xffffffffd420) at /worker/build/66bf5b2131e0cc44/root/external/libunwind/src/aarch64/Gis_signal_frame.c:51
#7  0x00000000004535e0 in _Uaarch64_step (cursor=0xffffffffd420) at /worker/build/66bf5b2131e0cc44/root/external/libunwind/src/aarch64/Gstep.c:204
```

The same issue applies to other architectures as well.

Fixes #340 
Fixes #530

### DIFF
--- a/src/aarch64/Ginit.c
+++ b/src/aarch64/Ginit.c
@@ -132,7 +132,7 @@ access_mem (unw_addr_space_t as, unw_word_t addr, unw_word_t *val, int write,
   if (unlikely (write))
     {
       Debug (16, "mem[%lx] <- %lx\n", addr, *val);
-      memcpy (addr, val, sizeof(unw_word_t));
+      memcpy ((void *) addr, val, sizeof(unw_word_t));
     }
   else
     {
@@ -143,7 +143,7 @@ access_mem (unw_addr_space_t as, unw_word_t addr, unw_word_t *val, int write,
         Debug (16, "mem[%016lx] -> invalid\n", addr);
         return -1;
       }
-      memcpy (val, addr, sizeof(unw_word_t));
+      memcpy (val, (void *) addr, sizeof(unw_word_t));
       Debug (16, "mem[%lx] -> %lx\n", addr, *val);
     }
   return 0;
@@ -164,12 +164,12 @@ access_reg (unw_addr_space_t as, unw_regnum_t reg, unw_word_t *val, int write,
 
   if (write)
     {
-      memcpy (addr, val, sizeof(unw_word_t));
+      memcpy ((void *) addr, val, sizeof(unw_word_t));
       Debug (12, "%s <- %lx\n", unw_regname (reg), *val);
     }
   else
     {
-      memcpy (val, addr, sizeof(unw_word_t));
+      memcpy (val, (void *) addr, sizeof(unw_word_t));
       Debug (12, "%s -> %lx\n", unw_regname (reg), *val);
     }
   return 0;
--- a/src/aarch64/Ginit.c
+++ b/src/aarch64/Ginit.c
@@ -132,7 +132,7 @@ access_mem (unw_addr_space_t as, unw_word_t addr, unw_word_t *val, int write,
   if (unlikely (write))
     {
       Debug (16, "mem[%lx] <- %lx\n", addr, *val);
-      *(unw_word_t *) addr = *val;
+      memcpy (addr, val, sizeof(unw_word_t));
     }
   else
     {
@@ -143,7 +143,7 @@ access_mem (unw_addr_space_t as, unw_word_t addr, unw_word_t *val, int write,
         Debug (16, "mem[%016lx] -> invalid\n", addr);
         return -1;
       }
-      *val = *(unw_word_t *) addr;
+      memcpy (val, addr, sizeof(unw_word_t));
       Debug (16, "mem[%lx] -> %lx\n", addr, *val);
     }
   return 0;
@@ -164,12 +164,12 @@ access_reg (unw_addr_space_t as, unw_regnum_t reg, unw_word_t *val, int write,
 
   if (write)
     {
-      *(unw_word_t *) addr = *val;
+      memcpy (addr, val, sizeof(unw_word_t));
       Debug (12, "%s <- %lx\n", unw_regname (reg), *val);
     }
   else
     {
-      *val = *(unw_word_t *) addr;
+      memcpy (val, addr, sizeof(unw_word_t));
       Debug (12, "%s -> %lx\n", unw_regname (reg), *val);
     }
   return 0;
--- a/src/arm/Ginit.c
+++ b/src/arm/Ginit.c
@@ -83,7 +83,7 @@ access_mem (unw_addr_space_t as, unw_word_t addr, unw_word_t *val, int write,
   if (write)
     {
       Debug (16, "mem[%x] <- %x\n", addr, *val);
-      *(unw_word_t *) addr = *val;
+      memcpy (addr, val, sizeof(unw_word_t));
     }
   else
     {
@@ -93,7 +93,7 @@ access_mem (unw_addr_space_t as, unw_word_t addr, unw_word_t *val, int write,
           Debug (16, "mem[%#010lx] -> invalid\n", (long)addr);
           return -1;
         }
-      *val = *(unw_word_t *) addr;
+      memcpy (val, addr, sizeof(unw_word_t));
       Debug (16, "mem[%#010lx] -> %#010lx\n", (long)addr, (long)*val);
     }
   return 0;
@@ -114,12 +114,12 @@ access_reg (unw_addr_space_t as, unw_regnum_t reg, unw_word_t *val, int write,
 
   if (write)
     {
-      *(unw_word_t *) addr = *val;
+      memcpy (addr, val, sizeof(unw_word_t));
       Debug (12, "%s <- %x\n", unw_regname (reg), *val);
     }
   else
     {
-      *val = *(unw_word_t *) addr;
+      memcpy (val, addr, sizeof(unw_word_t));
       Debug (12, "%s -> %x\n", unw_regname (reg), *val);
     }
   return 0;
--- a/src/arm/Ginit.c
+++ b/src/arm/Ginit.c
@@ -83,7 +83,7 @@ access_mem (unw_addr_space_t as, unw_word_t addr, unw_word_t *val, int write,
   if (write)
     {
       Debug (16, "mem[%x] <- %x\n", addr, *val);
-      memcpy (addr, val, sizeof(unw_word_t));
+      memcpy ((void *) addr, val, sizeof(unw_word_t));
     }
   else
     {
@@ -93,7 +93,7 @@ access_mem (unw_addr_space_t as, unw_word_t addr, unw_word_t *val, int write,
           Debug (16, "mem[%#010lx] -> invalid\n", (long)addr);
           return -1;
         }
-      memcpy (val, addr, sizeof(unw_word_t));
+      memcpy (val, (void *) addr, sizeof(unw_word_t));
       Debug (16, "mem[%#010lx] -> %#010lx\n", (long)addr, (long)*val);
     }
   return 0;
@@ -114,12 +114,12 @@ access_reg (unw_addr_space_t as, unw_regnum_t reg, unw_word_t *val, int write,
 
   if (write)
     {
-      memcpy (addr, val, sizeof(unw_word_t));
+      memcpy ((void *) addr, val, sizeof(unw_word_t));
       Debug (12, "%s <- %x\n", unw_regname (reg), *val);
     }
   else
     {
-      memcpy (val, addr, sizeof(unw_word_t));
+      memcpy (val, (void *) addr, sizeof(unw_word_t));
       Debug (12, "%s -> %x\n", unw_regname (reg), *val);
     }
   return 0;
--- a/src/hppa/Ginit.c
+++ b/src/hppa/Ginit.c
@@ -91,11 +91,11 @@ access_mem (unw_addr_space_t as, unw_word_t addr, unw_word_t *val, int write,
   if (write)
     {
       Debug (12, "mem[%x] <- %x\n", addr, *val);
-      memcpy (addr, val, sizeof(unw_word_t));
+      memcpy ((void *) addr, val, sizeof(unw_word_t));
     }
   else
     {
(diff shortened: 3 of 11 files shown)


lemul:
Misaligned ip? How did you accomplish that? Maybe we need to fix the cause of it?

sfc-gh-sgiesecke:
Hm, the calling code just looks like this:
```
  unw_cursor_t cursor;
  unw_context_t context;

  unw_getcontext(&context);
  unw_init_local(&cursor, &context);
  while (unw_step(&cursor) > 0) {
    // ...
  }
```

lemul:
ah, ip = 0xffff7ffd376c, it's ok

sfc-gh-sgiesecke:
> This was reported in issues #340 and #530. Would you be able to expand the fix to all architectures?

Ah, yes I think I can do that.

sfc-gh-sgiesecke:
> > This was reported in issues #340 and #530. Would you be able to expand the fix to all architectures?
> 
> Ah, yes I think I can do that.

@bregma I updated all other architectures' Ginit.c files as well now.

bregma:
Not sure why the aarch64 CI tests are failing with this change. Doesn't make any sense.

sfc-gh-sgiesecke:
> Not sure why the aarch64 CI tests are failing with this change. Doesn't make any sense.

Almost all architectures have failing tests. Are they green on main?

bregma:
The aarch64 tests are green on trunk (https://github.com/libunwind/libunwind/wiki/Current-CI-Status).

I have re-run the CI job to see if it was a fluke -- sometimes github actions can be a little flaky like that -- but it seems consistent. I will investigate in the next few days when I'm back at the office.  As I say it doesn't make sense since this change should be functionally identical to the existing code.

sfc-gh-sgiesecke:
Locally, I have trouble building the HEAD of libunwind. I run into
```
make[2]: *** No rule to make target 'aarch64/offsets.h', needed by 'all-am'.  Stop.
```
The file is not in the repository, not sure if it's supposed to be generated?

bregma:
OK, there were two problems (and you may blush at the second one).

First off, there's some compilation warnings in `access_mem()` that, when fixed, allows some tests to pass at `-O2` that were failing. That's a hint the compiler is pointing out undefined behaviour and the optimizer is eliminating code as a result. In this case, it because an integer is not a pointer and so violate the strict aliasing rule.

Second, well, the arguments to `memcpy()` are reversed in `access_reg()`.
```
diff --git a/src/aarch64/Ginit.c b/src/aarch64/Ginit.c
index 788f0448..821e45d2 100644
--- a/src/aarch64/Ginit.c
+++ b/src/aarch64/Ginit.c
@@ -132,7 +132,7 @@ access_mem (unw_addr_space_t as, unw_word_t addr, unw_word_t *val, int write,
   if (unlikely (write))
     {
       Debug (16, "mem[%lx] <- %lx\n", addr, *val);
-      memcpy (addr, val, sizeof(unw_word_t));
+      memcpy ((void *)addr, val, sizeof(unw_word_t));
     }
   else
     {
@@ -143,7 +143,7 @@ access_mem (unw_addr_space_t as, unw_word_t addr, unw_word_t *val, int write,
         Debug (16, "mem[%016lx] -> invalid\n", addr);
         return -1;
       }
-      memcpy (val, addr, sizeof(unw_word_t));
+      memcpy (val, (void *)addr, sizeof(unw_word_t));
       Debug (16, "mem[%lx] -> %lx\n", addr, *val);
     }
   return 0;
@@ -169,7 +169,7 @@ access_reg (unw_addr_space_t as, unw_regnum_t reg, unw_word_t *val, int write,
     }
   else
     {
-      memcpy (addr, val, sizeof(unw_word_t));
+      memcpy (val, addr, sizeof(unw_word_t));
       Debug (12, "%s -> %lx\n", unw_regname (reg), *val);
     }
   return 0;

```

With these changes I get 30 PASS 3 SKIP out of 33 tests on aarch64-pc-linux-gnu locally.

You might want to double-check the other architectures for this problem too.

sfc-gh-sgiesecke:
> Second, well, the arguments to memcpy() are reversed in access_reg().

Oops...

bregma:
OK, this is good to go. Thank you for this work.

sfc-gh-sgiesecke:
Re the build issues: v1.7.2 builds fine for me. I tried to bisect this, but apparently there are multiple overlapping build issues since v1.7.2.  All revisions between 08bc8c0da934fefec90bc8b96dcd582054d28934^..634206d8fe6cd0cf8f53aca8dcdf4110aef40904 suffer from `aarch64/Gresume.c:42:7: error: 'asm' operand has impossible constraints` and somewhere in that range the issue mentioned above was introduced.

Apparently c7df8e3e0d25b1584827d7b8a43199c0c402e7b7 broke this, but then the error was
```
make[2]: *** No rule to make target 'x86_64/offsets.h', needed by 'all-am'.  Stop.
```
(which is even stranger, since that's for a different architecture) and after 1b480c9936dd158bbcec49cbf3d1bb985207aa37 this changed to
```
make[2]: *** No rule to make target 'aarch64/offsets.h', needed by 'all-am'.  Stop.
```

It's not really clear to me what part of the Makefile requires this in the aarch64 build.

bregma:
That sounds like you need to rerun `autoreconf` and `./configure` to regenerate `Makefile.in` and `Makefile`.

sfc-gh-sgiesecke:
Ok, it seems I was missing libtool. But on HEAD, the `asm` issue remains. This was probably introduced by 08bc8c0da934fefec90bc8b96dcd582054d28934 (which mentions something about gcc 12; while I am using gcc 10), and later moved to a different file.

sfc-gh-sgiesecke:
Hm, clang has a different error:
```
error: inline assembly requires more registers than available
   65 |         "ldr x0,  %[x0]\n\t"
      |         ^
```

PR #612, opened 2023-09-06
Get filename and offset from ip

iDings:
For the case that can't get symbol info on board, 
it's a benefit if can get filename and offset, then use addr2line to decode on host.

This patch mainly export the result of ```tdep_get_elf_image``` to  public.


bregma:
This is wonderful (it's been way down on my to-do list, so far down I haven't even opened an issue for it).

It should be extended to other address spaces (local, ptrace, coredump), the new API functions should have man pages, and it would be great if there were tests added to the test suite. If you'd rather just open issues for these it would be fine and they can be addressed in separate PRs.

iDings:
Glad that you like this api,  i will finish the missing part of this PR

- [ ] extended to  (local, ptrace, coredump) address spaces
- [ ] add man pages
- [ ] add tests to the test suite

iDings:
@bregma an question need your help: unw_get_elf_filename will use ```tdep_get_elf_image``` internally, the result of this api will start with "/proc/$PID/root/" such as ``` [/proc/3232552/root/usr/lib/x86_64-linux-gnu/libc-2.31.so+0x2408a]```
   on linux. This is an problem or just fine?
https://github.com/libunwind/libunwind/blob/6abbc122d1740741d66c83f556eaea4cd7010980/src/os-linux.c#L44

iDings:
@bregma this patch is almost done, help to review this changes, thanks.
1.  reuse `tdep_get_elf_image`, if ei is NULL, don't map elf image, just return elf filename. And this can fix previous concern on linux platform: https://github.com/libunwind/libunwind/pull/612#issuecomment-1715774310
2. manual is heavily based  on unw_get_proc_name.tex,  hope i did not make some mistake.

bregma:
This is great. Good work, thanks.

PR #616, opened 2023-09-15
Fix maps leak if caller's pathlen is too small

iDings:
`maps_close` should be called when passed path buffer is too small


bertwesarg (approved):
Thanks, LGTM

PR #617, opened 2023-09-15
Import config.h before dwarf-config.h.

giomasce:
Header tdep-x86_64/dwarf-config.h depends on config.h because it uses CONFIG_MSABI_SUPPORT.

I discovered this problem because some compilation units actually import config.h before dwarf-config.h, and some others don't. So the `struct dwarf_cursor` definition was different in different compilation units, which caused segfaults.

I'm not sure if the proposed fix is the best one, or if dwarf-config.h should rather include config.h directly. Feel free to adjust accordingly.


bregma:
This is fine as-is. The include hierarchy, especially for `config.h`, needs some grooming and that can be done as a separate issue.

bregma (approved):

PR #620, opened 2023-09-19
Fix clang x86_64 unit test failure

bregma:
For whatever reason the clang runtime started raising a SIGILL instead of a SIGSEGV when dereferencing NULL. Fixed the test case to trap both signals.


bregma:
It seems this is not happening any more (clang 12, Ubuntu 20.04 x86_64). Mysterious, but whatever.

PR #638, opened 2023-10-10
Further improvements to unw_step fallback method on AArch64.

kenche-linaro:
Frame pointer (FP) cannot always be relied upon if DWARF unwinding fails. Depending on position in the function prologue/epilogue, the frame record may be pointed to by the FP, SP, or at an offset to SP. Detect which case it is by inspecting the function disassembly and matching to known frame record store/load instructions.

Unittest added to verify different frame record store/load methods are detected when given real function examples from a python2.7 binary.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,5 +84,6 @@ tests/[GL]x64-test-dwarf-expressions
 tests/x64-unwind-badjmp-signal-frame
 tests/[GL]arm64-test-sve-signal
 tests/aarch64-test-plt
+tests/aarch64-test-frame-record
 tests/*.log
 tests/*.trs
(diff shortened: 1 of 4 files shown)


bregma:
This change is consistently failing 3 CI tests for aarch64.

kenche-linaro:
> This change is consistently failing 3 CI tests for aarch64.

Thanks. I will take a look at this.

kenche-linaro:
> This change is consistently failing 3 CI tests for aarch64.

This should be resolved now.

I'm not sure why `build-clang-x86_64-O0` and `build-cross-s390x` are failing. The tests pass when I run through the CI steps manually.

bregma:
Still get 2 failures on aarch64-nto-qnx8.0.0 (`Gtest-trace` and `Ltest-trace`) but neither are new, just the failure mode has changed, and I'm not sure it's not just the test code that's a little off.

I think this is fine as-is so it's going in.

kenche-linaro:
Thanks!

PR #653, opened 2023-10-29
Use pkg-config to find dependencies

blue42u:
Currently the dependencies (zlib and liblzma) are found only through compiler checks. This means users wanting to customize these dependencies need to fiddle with `CFLAGS` and `LDFLAGS` to point to the one they want. This is inconvenient and nonstandard, and isn't the main purpose for these very generic variables.

This PR converts the dependency handling to use the standard `pkg-config` method instead. This allows users to customize the dependencies via `PKG_CONFIG_PATH`, no `*FLAGS` needed.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -413,38 +413,32 @@ if test x$enable_msabi_support = xyes; then
 fi
 AC_MSG_RESULT([$enable_msabi_support])
 
-LIBLZMA=
 AC_MSG_CHECKING([whether to support LZMA-compressed symbol tables])
 AC_ARG_ENABLE(minidebuginfo,
 AS_HELP_STRING([--enable-minidebuginfo], [Enables support for LZMA-compressed symbol tables]),, [enable_minidebuginfo=auto])
 AC_MSG_RESULT([$enable_minidebuginfo])
-if test x$enable_minidebuginfo != xno; then
-   AC_CHECK_LIB([lzma], [lzma_mf_is_supported],
-   [LIBLZMA=-llzma
-    AC_DEFINE([HAVE_LZMA], [1], [Define if you have liblzma])
+AS_IF([test x$enable_minidebuginfo != xno], [
+   PKG_CHECK_MODULES([LIBLZMA], [liblzma],
+   [AC_DEFINE([HAVE_LZMA], [1], [Define if you have liblzma])
     enable_minidebuginfo=yes],
    [if test x$enable_minidebuginfo = xyes; then
       AC_MSG_FAILURE([liblzma not found])
     fi])
-fi
-AC_SUBST([LIBLZMA])
+])
 AM_CONDITIONAL(HAVE_LZMA, test x$enable_minidebuginfo = xyes)
 
-LIBZ=
 AC_MSG_CHECKING([whether to support ZLIB-compressed symbol tables])
 AC_ARG_ENABLE(zlibdebuginfo,
 AS_HELP_STRING([--enable-zlibdebuginfo], [Enables support for ZLIB-compressed symbol tables]),, [enable_zlibdebuginfo=auto])
 AC_MSG_RESULT([$enable_zlibdebuginfo])
-if test x$enable_zlibdebuginfo != xno; then
-   AC_CHECK_LIB([z], [uncompress],
-   [LIBZ=-lz
-    AC_DEFINE([HAVE_ZLIB], [1], [Define if you have libz])
+AS_IF([test x$enable_zlibdebuginfo != xno], [
+   PKG_CHECK_MODULES([LIBZ], [zlib],
+   [AC_DEFINE([HAVE_ZLIB], [1], [Define if you have libz])
     enable_zlibdebuginfo=yes],
    [if test x$enable_zlibdebuginfo = xyes; then
       AC_MSG_FAILURE([libz not found])
     fi])
-fi
-AC_SUBST([LIBZ])
+])
 AM_CONDITIONAL(HAVE_ZLIB, test x$enable_zlibdebuginfo = xyes)
 
 AC_MSG_CHECKING([for Intel compiler])
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -151,12 +151,14 @@ libunwind_coredump_la_SOURCES =                \
 	coredump/_UPT_get_dyn_info_list_addr.c \
 	coredump/_UPT_put_unwind_info.c        \
 	coredump/_UPT_resume.c
+libunwind_coredump_la_CFLAGS =                 \
+	$(LIBLZMA_CFLAGS) $(LIBZ_CFLAGS)
 libunwind_coredump_la_LDFLAGS =                \
 	$(COMMON_SO_LDFLAGS)                   \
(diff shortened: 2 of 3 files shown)


bregma:
Using `CPPFLAGS` and `LDFLAGS` _is_ the well-documented standard way of specifying non-standard installation locations for depdendent software using the GNU autotools-based configure scripts. `configure --help` calls that out specifically.

Using `pkd-config` only works on some desktop systems (eg. self-hosted Linux), which is only a subset of target systems libunwind gets built for. In addition, using `pkg-config` for cross-compiling (another very common usage scenario for libunwind) involves a lot of special manual setup steps for those targets where it's even supported.

Making the general and common cases more difficult so that edge cases that already work can become easier seems like the wrong call. I'm inclined to reject this PR but I'll leave it open for a while if anyone else want to comment in favour.

bertwesarg:
I prefer to provide the user dependency dependent `*FLAG` variables. So in this case `ZLIB_{CFLAGS,LDFLAGS,LIBS}` and `LZMA_{CFLAGS,LDFLAGS,LIBS}`. If the configure check works with these variables, then they are added to `{CFLAGS,LDFLAGS,LIBS}`. The user can then populate this prior configure with `ZLIB_CFLAGS=$(pkg-config --cflags)`/…

bregma:
It already works if you just use `configure CPPFLAGS=$(pkg-config --cflags zlib) LDFLAGS=$(pkg-config --libs-only-L zlib)` (for example) and it doesn't require any non-portable or unconventional hand-coded deviations from documented and established autoconf practice.

blue42u:
> Using `CPPFLAGS` and `LDFLAGS` is the well-documented standard way of specifying non-standard installation locations for depdendent software using the GNU autotools-based configure scripts. `configure --help` calls that out specifically. ...

The recommended way to handle dependencies for autotools is to use `pkg-config`, cite the [Autotools Mythbuster](https://autotools.info/pkgconfig/index.html). Outside of autotools, adding dependencies to `CPPFLAGS` and `LDFLAGS` is **heavily discouraged** by all other build systems, if they don't have a custom idea (e.g. `cmake`) they often rely on `pkg-config` instead of raw compiler tests.

Coming in from the Linux software community at large, not supporting `pkg-config` for dependency resolution is surprising and causes very reasonable headaches when trying to build libunwind the first time. IMO it should "just work" without needing extra configuration. Hence the PR.

> I prefer to provide the user dependency dependent `*FLAG` variables. ...

The `PKG_CHECK_MODULES` macro actually adds variables as you describe (with the exception that `LDFLAGS` and `LIBS` are one variable named `LIBS`):
```console
$ ./configure --help
...
  LIBLZMA_CFLAGS
              C compiler flags for LIBLZMA, overriding pkg-config
  LIBLZMA_LIBS
              linker flags for LIBLZMA, overriding pkg-config
  LIBZ_CFLAGS C compiler flags for LIBZ, overriding pkg-config
  LIBZ_LIBS   linker flags for LIBZ, overriding pkg-config
...
```
This also means `pkg-config` is optional, it only needs to be installed during `autoreconf` to provide the macro.

___

Updated logic, I think this resolves the issues mentioned:

1. If `*_CFLAGS` and `*_LIBS` are set for a dependency (i.e. `LIBLZMA` or `LIBZ`), use those compiler flags.
2. If `pkg-config` is available and is able to find a dependency (`liblzma` or `zlib`), use those compiler flags.
3. Search for the dependency using `AC_SEARCH_LIBS` (`-llzma` or `-lz`), and if found use the identified linker flag.
4. Otherwise consider the dependency not found.

bregma:
I have no objection to using separate `*_CFLAGS` and `*_LIBS` variables in the interest of consistency with established norms.

I object to breaking cross compilation since cross compilation of libunwind is a very *very* common thing. As it is, this change always uses pkg-config to check the build machine for the libraries and not the target for non-Linux-desktop targets and even for cross-compiling Linux desktops targets it requires installing extra packages or custom wrappers just to get functionality we already have that just works. If it requires extraordinary workarounds for the general case just to make a special case of using a custom pkg-config configuration easier it's not going in.

blue42u:
I don't want to be confrontational, but using a custom `pkg-config` configuration is very common on Linux desktop/server, it's *the* way you build software without root access. For example, using a software stack built by [Spack](https://spack.io/):

```console
$ spack load liblzma zlib-api pkgconf  # Sets PKG_CONFIG_PATH etc.
$ ./configure --prefix=`realpath ./install`
$ make
$ make install
```

This pattern "just works" for all CMake-based, Meson-based and IME a majority of autoconf-based software. In libunwind's case, at best this fails in `configure`, at worst it links against a different `zlib` or `liblzma` than desired (only to be discovered much later when things crash). Being different here and enforcing manual configuration only causes surprises and hurts the user experience.

> As it is, this change always uses pkg-config to check the build machine for the libraries and not the target for non-Linux-desktop targets and even for cross-compiling Linux desktops targets it requires installing extra packages or custom wrappers just to get functionality we already have that just works. If it requires extraordinary workarounds for the general case just to make a special case of using a custom pkg-config configuration easier it's not going in.

The process to cross-compile with `pkg-config` is not all that difficult: you set `PKG_CONFIG` to a short wrapper script that sets the `PKG_CONFIG_(PATH|LIBDIR|SYSROOT_DIR)` environment variables, as noted in the man page and better described in the [Autotools Mythbuster](https://autotools.info/pkgconfig/cross-compiling.html). I assume this is the "extraordinary workaround" you're concerned about. `pkg-config`  has widespread prevalence on Linux desktop/server, IMHO any developers that are cross-compiling on such systems likely already have such a wrapper script in hand (or a package installed that provides it). In summary, this workaround is really quite ordinary on such systems.

There are other workarounds available. You can set the specific `(LIBZ|LIBLZMA)_(CFLAGS|LIBS)` variables and override the `pkg-config` checks completely. Or, you can mask out `pkg-config` and fall back to `CFLAGS` and `LDFLAGS`, e.g.:

```console
$ ./configure PKG_CONFIG= CC=x-gcc --host=x CFLAGS=-I/zlib-x/include LDFLAGS=-L/zlib-x/lib ...
```

Or both at once, the workarounds are not mutually exclusive. Both of these workarounds are minimal tweaks to what a user would do currently, cross-compiling or otherwise.

With this PR, the build "just works" in a range of very common cases that currently require (on Linux desktop/server, uncommon) manual configuration, and current cross-compilation cases require minimal (and on Linux desktop/server, common) changes to `./configure` arguments/environment. I'll respect your decision, but IMHO this PR is overall a change for the better and an improvement to the user experience.

rossburton:
>I object to breaking cross compilation since cross compilation of libunwind is a very very common thing.

FWIW, as someone whose day job is maintaining a cross-compiled distribution, pkg-config is awesome.  We've carried a patch to make gnupg use pkg-config for not long off a decade despite upstream refusing to accept it specifically because it makes cross-compiling easier.

In a cross-build pkg-config needs to be configured once and everything just works.  The alternative is knowing exactly what variables need to be set of what packages, which is painful.

mattst88:
FWIW, I agree that pkg-config is the right way to go (when cross-compiling or otherwise).

PR #660, opened 2023-11-12
Fix can't get signal restorer proc name on aarch64-linux

iDings:
aarch64-linux's default signal restore function is provided by kernel's vDSO. 
This patch need based on https://github.com/libunwind/libunwind/pull/658 to make libunwind search symbol on vDSO. 

This patch fix following two issue:
1. dump vdso from proc maps, the symbol type of __kernel_rt_sigreturn is NOTYPE. 
```shell
eu-readelf -s .dynsym libvdso-aarch64.so 
...
3: 0000000000000790      8 NOTYPE  GLOBAL DEFAULT        7 __kernel_rt_sigreturn@@LINUX_2.6.39
...
```
2. If current cusor is signal_frame, the ip not need to decrease one, otherwise can't get the proc name by ip.

On x86-64 with ubuntu20.04 , use Ltest-bt test,  still can't decode signal restore procedure name,  from following log info,
the ip is still be decreased one, i didn't see any signal frame pattern code on x86_64's [unw_is_signal_frame](https://github.com/libunwind/libunwind/blob/master/src/x86_64/Gos-linux.c#L85), how x86_64 check current frame is signal?
Need some special check on x86-64 linux.
```
_ULx86_64_get_proc_name ip:0x7f70d995608f
load_debuginfo:/proc/4154962/root/usr/lib/x86_64-linux-gnu/libc-2.31.so
load_debuginfo:/usr/lib/debug/.build-id/18/78e6b475720c7c51969e69ab2d276fae6d1dee.debug
00007f70d9956090                                  (sp=00007ffefb7c9000)
        proc=0x7f70d995608f-0x7f70d9956099
        handler=0x0 lsda=0x0 gp=0x7f70d9aff000
        [/usr/lib/x86_64-linux-gnu/libc-2.31.so+0x43090]
```


bregma:
> how x86_64 check current frame is signal?

On x86_64 signal trampolines are indicated by the presence of an `S` in the CIE's augmentation data.  See `dwarf/Gfde.c` if that's any help.

iDings:
This patch https://github.com/libunwind/libunwind/pull/660/commits/fbf434bbe80b69ee0d584ebe4998045283953d78 broken some CIs, need some better way to fix `can't decode signal restore procedure name on x86_64 linux`.

The main problem is if current frame is signal frame pc should not decrease one to get procedure name, 
but for x86_64-linux , the `c->sigcontext_format` is still `X86_64_SCF_NONE`,  and [unw_is_signal_frame](https://github.com/libunwind/libunwind/blob/master/src/x86_64/Gos-linux.c#L92) return false, only after use unw_get_proc_info or unw_step, the sigcontext_format will be updated to X86_64_SCF_LINUX_RT_SIGFRAME.

iDings:
> > how x86_64 check current frame is signal?
> 
> On x86_64 signal trampolines are indicated by the presence of an `S` in the CIE's augmentation data. See `dwarf/Gfde.c` if that's any help.

@bregma  it helps, i need do some more learns about dwarf.  And for x86_64-linux, maybe need prefetch `S` in the CIE's augmentation data to update sigcontext_format when do unw_step.

bregma:
How does the cursor get in to the signal trampoline without going through `unw_step()`?

iDings:
Take Ltest-bt as an example, add following log info

```diff
diff --git a/tests/Gtest-bt.c b/tests/Gtest-bt.c
index 31179f3e..0b0bf70a 100644
--- a/tests/Gtest-bt.c
+++ b/tests/Gtest-bt.c
@@ -77,6 +77,7 @@ do_backtrace (void)
       unw_get_reg (&cursor, UNW_REG_IP, &ip);
       unw_get_reg (&cursor, UNW_REG_SP, &sp);
       buf[0] = '\0';
+      printf ("before get_proc_name: is_signal_frame:%d\n", unw_is_signal_frame(&cursor));
       if (unw_get_proc_name (&cursor, name, sizeof (name), &off) == 0)
 	{
 	  if (off)
@@ -84,12 +85,13 @@ do_backtrace (void)
 	  else
 	    snprintf (buf, sizeof (buf), "<%s>", name);
 	}
-      if (verbose)
+      if (0)
 	{
 	  printf ("%016lx %-32s (sp=%016lx)\n", (long) ip, buf, (long) sp);
 
 	  if (unw_get_proc_info (&cursor, &pi) == 0)
 	    {
+          printf ("after get_proc_info: is_signal_frame:%d\n", unw_is_signal_frame(&cursor));
 	      printf ("\tproc=0x%lx-0x%lx\n\thandler=0x%lx lsda=0x%lx gp=0x%lx",
 		  (long) pi.start_ip, (long) pi.end_ip,
 		  (long) pi.handler, (long) pi.lsda, (long) pi.gp);
@@ -109,7 +111,10 @@ do_backtrace (void)
           if (unw_get_elf_filename (&cursor, name, sizeof (name), &off) == UNW_ESUCCESS)
                 printf ("\t[%s+0x%lx]\n", name, (long) off);
 	}
+    else
+	  printf ("%016lx %-32s (sp=%016lx)\n", (long) ip, buf, (long) sp);
 
+      printf ("call unw_step\n");
       ret = unw_step (&cursor);
       if (ret < 0)
 	{
```

The output log of example:
```
Backtrace across SIGSEGV handler:                                         
segv_handler: got signal 11
        explicit backtrace:                                               
before get_proc_name: is_signal_frame:0                                   
0000000000402781 <do_backtrace+0x21>              (sp=00007ffd2bd70850)

call unw_step                                                             
before get_proc_name: is_signal_frame:0                                   
0000000000402993 <segv_handler+0x13>              (sp=00007ffd2bd70930)

call unw_step                                                             
before get_proc_name: is_signal_frame:0                                   
00007f428e152bb0                                  (sp=00007ffd2bd70940)

call unw_step                                                             
before get_proc_name: is_signal_frame:1                                
00000000004025e7 <main+0x1e7>                     (sp=00007ffd2bd71630)

call unw_step                                                             
before get_proc_name: is_signal_frame:0                                
00007f428e13cb8a <__libc_start_call_main+0x7a>    (sp=00007ffd2bd71700)

call unw_step
before get_proc_name: is_signal_frame:0
00007f428e13cc4b <__libc_start_main+0x8b>         (sp=00007ffd2bd717a0)

call unw_step  
before get_proc_name: is_signal_frame:0
0000000000402695 <_start+0x25>                    (sp=00007ffd2bd71800)

call unw_step  
                                                                          
        via backtrace():                                                  
[0] ip=0x402887
[1] ip=0x402993                                                           
[2] ip=0x7f428e152bb0                                                     
[3] ip=0x4025e7
[4] ip=0x7f428e13cb8a
[5] ip=0x7f428e13cc4b
[6] ip=0x402695
```

I think on IP:00007f428e152bb0, the `unw_is_signal_frame` should be true, but need do an next `unw_step` or call `unw_get_proc_info` , then `unw_is_signal_frame` becomes true.

bregma:
Ah, yes, I see. It doesn't process the CIE until it's trying to unwind the current frame. That's not a problem when unwinding but it's a problem for metadata calls (like procinfo, procname). Needs some serious changes.

iDings:
It's really weird why build-gcc-i686-O0 is not passed because the `test-async-sig` failed. And i revert the patch, and test the upstream libunwind on my local ubuntu20.04.6,  can also reproduce the crash of  `test-async-sig` probabilistically, it seems not related to this patch.

On local ubbuntu, use following cammand to build, and run `make check`,  
```
./configure CFLAGS="-O0 -m32 -g" CXXFLAGS="-O0 -m32 -g" --build=x86_64-pc-linux-gnu --host=i686-pc-linux-gnu --disable-shared --enable-debug
make
make check
```

If disable `UNW_CACHE_GLOBAL` and `UNW_CACHE_PER_THREAD` , and test 100 times, all passed.
https://github.com/libunwind/libunwind/blob/a9a110f7685202fe2c7c603128e780eb3bb3ea80/tests/test-async-sig.c#L143-L146

iDings:
yeah, for https://github.com/libunwind/libunwind/issues/569, `-O0` seems only decrease the probability of the crash of `test-async-sig`.  I am trying to dig more about the cache problem.

And for the  `can't get signal restorer proc name on x86_64`, add signal_frame sniffer in x86_64 `unw_is_signal_frame`, and check whether current frame is signal frame or not at the end of `unw_step`, if current is signal frame, set `use_prev_instr` to 0,
this can also prevent `fetch_proc_info` to decrease ip by one in next `unw_step`.  What do think about this approach? 
I try to use signal frame flag in dwarf directly, but failed.  
https://github.com/libunwind/libunwind/blob/a9a110f7685202fe2c7c603128e780eb3bb3ea80/src/dwarf/Gparser.c#L436-L456

PR #662, opened 2023-11-17
Fix edge condition in unw_address_is_valid

### LINKED ISSUE #661: unw_address_is_valid failed to validate incorrect address
Hi,

We had a coredump when unwinding, with the backtrace looks like

```
#0  access_mem (as=<optimized out>, addr=140196322476030, val=<optimized out>, write=<optimized out>, arg=<optimized out>) at x86_64/Ginit.c:94
#1  0x00007f836b32e3ef in _ULx86_64_step (cursor=cursor@entry=0x7ffd435c0a20) at x86_64/Gstep.c:240
#2  0x00007f836b32edce in trace_init_addr (rsp=<optimized out>, rbp=<optimized out>, rip=<optimized out>, cfa=<optimized out>, cursor=0x7ffd435c0a20,
    f=0x7f809e68e9d0) at x86_64/Gtrace.c:249
#3  trace_lookup (rsp=<optimized out>, rbp=<optimized out>, rip=<optimized out>, cfa=<optimized out>, cache=0x7f836b352fe0, cursor=0x7ffd435c0a20)
    at x86_64/Gtrace.c:331
#4  _ULx86_64_tdep_trace (cursor=cursor@entry=0x7ffd435c0a20, buffer=buffer@entry=0x7f832aa316b8, size=size@entry=0x7ffd435c0644) at x86_64/Gtrace.c:449
#5  0x00007f836b32c4ec in unw_backtrace (buffer=0x7f832aa316b8, size=size@entry=128) at mi/backtrace.c:70
#6  0x00005583e78fbcee in prof_backtrace_impl (vec=<optimized out>, len=0x7ffd435c1288, max_len=<optimized out>) at src/prof_sys.c:60
....
```

Specifically, we have:

```
(gdb) x/8b addr
0x7f81fffffffe: ""
0x7f81ffffffff: ""
0x7f8200000000: <error: Cannot access memory at address 0x7f8200000000>
0x7f8200000000: <error: Cannot access memory at address 0x7f8200000000>
0x7f8200000000: <error: Cannot access memory at address 0x7f8200000000>
0x7f8200000000: <error: Cannot access memory at address 0x7f8200000000>
0x7f8200000000: <error: Cannot access memory at address 0x7f8200000000>
0x7f8200000000: <error: Cannot access memory at address 0x7f8200000000>
```

Looking specifically at the `unw_address_is_valid` function, i'm not sure that it's checking the address correctly, specifically this part

```
// If we're still on the same page, we're done.
size_t stride = len-1 < (size_t) unw_page_size ? len-1 : (size_t) unw_page_size;
len -= stride;
addr += stride;
if (unw_page_start (addr) == unw_page_start (lastbyte))
        break;
```

if for any reason, `addr` and `lastbyte` are on 2 different consecutive pages, then we skip the check for `lastbyte` page, which is incorrect. I think the check should be

```
// If we're still on the same page, we're done.
if (unw_page_start (addr) == unw_page_start (lastbyte))
        break;
size_t stride = len-1 < (size_t) unw_page_size ? len-1 : (size_t) unw_page_size;
len -= stride;
addr += stride;
```

bregma:
Logic was skipping the last page in a multi-page check.

Closes #661

### DIFF
--- a/src/mi/Gaddress_validator.c
+++ b/src/mi/Gaddress_validator.c
@@ -42,10 +42,10 @@ static atomic_flag _unw_address_validator_initialized = ATOMIC_FLAG_INIT;
 static int _mem_validate_pipe[2] = {-1, -1};
 
 #ifdef HAVE_PIPE2
-static void
+static int
 _do_pipe2 (int pipefd[2])
 {
-  int result UNUSED = pipe2 (pipefd, O_CLOEXEC | O_NONBLOCK);
+  return pipe2 (pipefd, O_CLOEXEC | O_NONBLOCK);
 }
 #else
 static void
@@ -61,17 +61,20 @@ _set_pipe_flags (int fd)
   fcntl (fd, F_SETFL, status_flags);
 }
 
-static void
+static int
 _do_pipe2 (int pipefd[2])
 {
-  pipe (pipefd);
+  if (pipe (pipefd) < 0)
+    {
+      return -1;
+    }
   _set_pipe_flags(pipefd[0]);
   _set_pipe_flags(pipefd[1]);
 }
 #endif
 
 
-static void
+static int
 _open_pipe (void)
 {
   if (_mem_validate_pipe[0] != -1)
@@ -79,7 +82,7 @@ _open_pipe (void)
   if (_mem_validate_pipe[1] != -1)
     close (_mem_validate_pipe[1]);
 
-  _do_pipe2 (_mem_validate_pipe);
+  return _do_pipe2 (_mem_validate_pipe);
 }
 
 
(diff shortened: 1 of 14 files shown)


src/mi/Gaddress_validator.c
```diff
@@ -239,32 +236,37 @@ unw_address_is_valid(unw_word_t addr, size_t len)
...
+
   if (unlikely (!atomic_flag_test_and_set(&_unw_address_validator_initialized)))
     {
       _open_pipe ();
     }
 
-  unw_word_t lastbyte = addr + (len - 1); // highest addressed byte of data to access
-  while (1)
+  for (unw_word_t page_addr = start_page_addr;
+  	   page_addr < end_page_addr;
```
dqminh, line 258: if `end_page_addr` wrap around, then this will fail because we will not verify the page_addr. For example, page_addr = 2^64.

dqminh: @bregma unfortunately we are dealing with some weird bug with a JIT engine that output addresses like that, so running this uncover such edge cases.

bregma:
@dqminh Does this latest change satisfy your case?

bregma:
This is going in because it fixes the issue and addresses the pre-existing edge cases.

PR #675, opened 2023-12-02
configure.ac: fix bashism in LDFLAGS append

thesamesam:
'+=' is not required for POSIX shells and may not work with e.g. /bin/sh provided by dash. Just expand it instead.


bregma:
This is fine. Could you also make a PR for the `v1.8_stable` branch?

thesamesam:
> This is fine. Could you also make a PR for the `v1.8_stable` branch?

Done in https://github.com/libunwind/libunwind/pull/676.

bertwesarg:
why not using `AS_VAR_APPEND` directly?

PR #684, opened 2023-12-15
Add use of libabigail for ABI checks

bregma:
Added `make -C src check-abi` target to leverage libabigail 2.0 tools to check for ABI changes.

Add ABI baseline files for aarch64, i686, riscv, s390x, and x86_64 Linux targets and aarch64 and x86_64 QNX SDP 7.1 targets.

Used the check-abi target in the CI-unix github workflow.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -1251,9 +1251,45 @@ libunwind_la_LIBADD += $(LIBLZMA) $(LIBZ)
 
 noinst_HEADERS += unwind/unwind-internal.h
 
-EXTRA_DIST = $(libunwind_la_EXTRAS_ia64)
+EXTRA_DIST = $(libunwind_la_EXTRAS_ia64) abi
 
 MAINTAINERCLEANFILES = Makefile.in
+
+#
+# ABI checks
+#
+# Running the ABI checks require configuring with CFLAGS="-g -Og".
+# The result is output in a file with the `.abidiff` extension which is also
+# printed to stdout if there is an ABI mismatch detected.
+#
+# This also requires libabigail tools of at least version 2.0 to be installed.
+#
+.la.abi:
+	$(AM_V_GEN)soname=$$(sed -n -e "s#dlname='\(.*\)\.[0-9]*'#\1#p" $<); \
+	if test -n "$$soname"; then \
+		$(ABICHECK) .libs/$${soname} \
+		            --no-show-locs \
+		            --no-comp-dir-path \
+		            --no-corpus-path \
+		            --drop-undefined-syms \
+		            --drop-private-types \
+		            --out-file $@; \
+	else \
+		printf "(no SONAME found in %s)\n" "$<"; \
+	fi
+
+abi_srcdir = "$(top_srcdir)/src/abi"
+abi_targetdir = "$(abi_srcdir)/$(arch)/$(target_os)"
+.abi.abidiff:
+	$(AM_V_GEN)if test -f "$<"; then \
+		$(ABIDIFF) --suppressions "$(abi_srcdir)/libunwind.supp" \
+		           -l "$(abi_targetdir)/$<" "$<" >$@ \
+			|| cat $@; \
+	fi
+
+abi-check: $(lib_LTLIBRARIES:.la=.abidiff) 
+
+.PHONY: abi-check
 
 # The -version-info flag accepts an argument of the form
 # `current[:revision[:age]]'. So, passing `-version-info 3:12:1' sets
--- a/.github/workflows/CI-unix.yml
+++ b/.github/workflows/CI-unix.yml
@@ -42,23 +42,30 @@ jobs:
         if: ${{ matrix.target.arch }} = 'i686'
         run: |
           sudo apt update
-          sudo apt install -y g++-12-multilib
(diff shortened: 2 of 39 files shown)


src/Makefile.am
```diff
@@ -1248,10 +1248,44 @@ AM_CPPFLAGS = -I$(top_srcdir)/include -I$(top_srcdir)/include/tdep-$(arch) -I.
...
+               --drop-private-types
+
+abi-generate: $(lib_LTLIBRARIES)
+	@printf 'Generating ABI files\n'; \
+	mkdir -p abi; \
+	for lib in $(lib_LTLIBRARIES:.la=); do \
+		sofile="$$(find . -name $${lib}.so)"; \
+		abifile="abi/$${lib}-${arch}-${target_os}.abi"; \
+		abidw "$${sofile}" $(abidw_common) --out-file "$${abifile}"; \
+	done
```
bertwesarg, line 1276: honor silient rules:

```suggestion
	@mkdir -p abi; \
	for lib in $(lib_LTLIBRARIES:.la=); do \
		sofile="$$(find . -name $${lib}.so)"; \
		abifile="abi/$${lib}-${arch}-${target_os}.abi"; \
		if $(AM_V_P); then echo abidw "$${sofile}" $(abidw_common) --out-file "$${abifile}"; \
		              else echo "  GEN ABI  $${abifile}"; \
		fi; \
		abidw "$${sofile}" $(abidw_common) --out-file "$${abifile}"; \
	done
```

bregma:
Found a better, more elegant way to generate and process the ABI files by using **make** rules instead of shell scripting.

src/Makefile.am
bregma: See latest revision.

bertwesarg:
> Found a better, more elegant way to generate and process the ABI files by using **make** rules instead of shell scripting.

nice. just in case you want a custom prefix instead of the generic `GEN`. put something like this into your Makefile.am:

```
UNW_V_ABIDW = $(unw__v_ABIDW_@AM_V@)
unw__v_ABIDW_ = $(unw__v_ABIDW_@AM_DEFAULT_V@)
unw__v_ABIDW_0 = @echo "  ABIDW    $@";
unw__v_ABIDW_1 =
UNW_V_ABIDIFF = $(unw__v_ABIDIFF_@AM_V@)
unw__v_ABIDIFF_ = $(unw__v_ABIDIFF_@AM_DEFAULT_V@)
unw__v_ABIDIFF_0 = @echo "  ABIDIFF  $@";
unw__v_ABIDIFF_1 =
```

and use `$(UNW_V_ABIDW)` and `$(UNW_V_ABIDIFF)`

src/Makefile.am
```diff
@@ -1248,10 +1248,39 @@ AM_CPPFLAGS = -I$(top_srcdir)/include -I$(top_srcdir)/include/tdep-$(arch) -I.
...
 MAINTAINERCLEANFILES = Makefile.in
 
+#
+# ABI checks
+#
+# Running the ABI checks require configuring with CFLAGS="-g -Og" and running
+# the targets explicitly with `make -C src abi-check`. The result is output in a
+# file with the `.abidiff` extension.
+#
+# This also requires libabigail tools of at least version 2.0 to be installed.
```
bertwesarg, line 1262: Ubunt 20.04 TLS  still includes 1.6. I think that justifies a configure check, which then also includes the `--enable-shared` check

src/Makefile.am
```diff
@@ -1248,10 +1248,39 @@ AM_CPPFLAGS = -I$(top_srcdir)/include -I$(top_srcdir)/include/tdep-$(arch) -I.
...
+#
+.la.abi:
+	$(AM_V_GEN) soname=$$(sed -n -e "s#dlname='\(.*\)\.[0-9]*'#\1#p" $<); \
+	if test -n "$$soname"; then \
```
bertwesarg, line 1266: test for the existens of `test -f .libs/$${soname}`?

bregma:
The archived ABI files need to be generated against a build configured with `--enable-debug` because that configuration flag **changes the ABI** and it's used in CI. Maybe there should be two sets of archived ABI files, one for with `--enable-debug` and one for without?

src/Makefile.am
```diff
@@ -1248,10 +1248,39 @@ AM_CPPFLAGS = -I$(top_srcdir)/include -I$(top_srcdir)/include/tdep-$(arch) -I.
 AM_CCASFLAGS = $(AM_CPPFLAGS)
 noinst_HEADERS += unwind/unwind-internal.h
 
-EXTRA_DIST = $(libunwind_la_EXTRAS_ia64)
+EXTRA_DIST = $(libunwind_la_EXTRAS_ia64) abi
 
 MAINTAINERCLEANFILES = Makefile.in
 
+#
```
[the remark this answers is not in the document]
bregma: Now it'll just work (and do nothing) with `--disable-shared`.

src/Makefile.am
bregma: Done.

src/Makefile.am
bregma: Done.

bertwesarg:
LGTM

Issue #164, opened 2020-03-19
Failing tests Alpine Linux (musl libc) v1.3.1


TBK:
```sh
===========================================
   libunwind 1.3.1: tests/test-suite.log
===========================================

# TOTAL: 35
# PASS:  24
# SKIP:  0
# XFAIL: 0
# FAIL:  11
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: Gtest-resume-sig
======================

Segmentation fault (core dumped)
FAIL Gtest-resume-sig (exit status: 139)

FAIL: Ltest-resume-sig
======================

Segmentation fault (core dumped)
FAIL Ltest-resume-sig (exit status: 139)

FAIL: Gtest-resume-sig-rt
=========================

Segmentation fault (core dumped)
FAIL Gtest-resume-sig-rt (exit status: 139)

FAIL: Ltest-resume-sig-rt
=========================

Segmentation fault (core dumped)
FAIL Ltest-resume-sig-rt (exit status: 139)

FAIL: Gtest-trace
=================

FAILURE: unw_step() loop and unw_backtrace() depths differ: 3 vs. 3
FAILURE: detected 2 errors
FAILURE: unw_step() loop and backtrace() depths differ: 3 vs. 3
FAIL Gtest-trace (exit status: 255)

FAIL: Ltest-trace
=================

FAILURE: unw_step() loop and unw_backtrace() depths differ: 3 vs. 3
FAILURE: detected 2 errors
FAILURE: unw_step() loop and backtrace() depths differ: 3 vs. 3
FAIL Ltest-trace (exit status: 255)

FAIL: Ltest-init-local-signal
=============================

unw_getcontext steps was 1, ucontext steps was 2, should be -1
FAIL Ltest-init-local-signal (exit status: 255)

FAIL: test-async-sig
====================

FAILURE: unw_step() looping over 102 iterations
FAILURE: unw_step() looping over 102 iterations
FAILURE: unw_step() looping over 102 iterations
FAILURE: unw_step() looping over 102 iterations
FAILURE: unw_step() looping over 102 iterations
FAILURE: unw_step() looping over 102 iterations
FAILURE: unw_step() looping over 102 iterations
FAILURE: unw_step() looping over 102 iterations
FAILURE: unw_step() looping over 102 iterations
FAILURE: unw_step() looping over 102 iterations
FAILURE: unw_step() looping over 102 iterations
FAILURE: unw_step() looping over 102 iterations
[...]

PR #691, opened 2023-12-30
Add explicit support for linux-musl targets

bregma:
- added musl target CI builds to `CI-unix.yml`
- added config-time check for the libucontext (for `getcontext()` support)
- added better detection of `NULL` IP in x86_64 `unw_step()` because musl does not follow the *de facto* x86_64 ABI 
  conventions for main program entry

Partially addresses #164 but signal trampoline detection still needs to be addressed for x86_64 and there are other issues for other architectures.

### DIFF
--- a/.github/workflows/CI-unix.yml
+++ b/.github/workflows/CI-unix.yml
@@ -19,9 +19,9 @@ on:
       - master
 
 jobs:
-  build-native:
+  build-gnu-native:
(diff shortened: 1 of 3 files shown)


bregma:
I changed the name of the CI jobs so the required checks need to be manually renamed -- after this gets committed.

PR #699, opened 2024-01-15
Add missing return in _do_pipe2

am11:
Found this on linux-x64 when updating dotnet-runtime to v1.8.0.
```
 /runtime/src/native/external/libunwind/src/mi/Gaddress_validator.c:73:1: error: non-void function does not return a value in all control paths [-Werror,-Wreturn-type]
     73 | }
        | ^
  1 error generated.
  make[2]: *** [/runtime/artifacts/obj/external/libunwind/CMakeFiles/libunwind.dir/build.make:872: /runtime/artifacts/obj/external/libunwind/CMakeFiles/libunwind.dir/runtime/src/native/external/libunwind/src/mi/Gaddress_validator.c.o] Error 1
  make[2]: *** Waiting for unfinished jobs....
  [ 41%] Building C object /runtime/artifacts/obj/external/libunwind/CMakeFiles/libunwind.dir/runtime/src/native/external/libunwind/src/x86_64/Ginit
```

(There are other issues in other OS/archs, will post the patches as I figure out the minimal fixes for those..)

### DIFF
--- a/src/mi/Gaddress_validator.c
+++ b/src/mi/Gaddress_validator.c
@@ -70,6 +70,7 @@ _do_pipe2 (int pipefd[2])
     }
   _set_pipe_flags(pipefd[0]);
   _set_pipe_flags(pipefd[1]);
+  return 0;
 }
 #endif
 


bregma:
I'm concerned something so obvious is not picked up in the CI builds. Does this require `/W4` in the CFLAGS in `CMakeLists.txt`?

am11:
This was on linux-x64, so I think it appears with `Wall`: https://github.com/dotnet/runtime/blob/d7604b6d259c3beb9a5213f7995e4bf6cc0b3cf9/eng/native/configurecompiler.cmake#L26 (+/- some warning omissions).

bregma:
GNU/Linux has `pipe2()` so the bad code should never be exercised when targeting that OS. Unless your CMake configure does not autodetect support for that function, which is not the case for libunwind Linux CI builds. Note that Musl/Linux, FreeBSD, and QNX all also have `pipe2()`.

The code is obviously wrong and needs to be fixed. I just want to be able to detect and fix such problems in the libunwind CI precommit builds, and the existing Win64 precommit CI builds do not do this.

am11:
> Unless your CMake configure does not autodetect support for that function, which is not the case for libunwind Linux CI builds.

This was the problem. Adding the missing introspection and definition fixed the build without this patch.

bregma:
OK, found QNX OS 7.0 does not have `pipe2()` so I can exercise that code path in regression testing.

PR #706, opened 2024-01-15
tests/Garm64-test-sve-signal: check that SVE is present before running tests

rossburton:
If the compiler supports `-march=armv8-a+sve` then those options are used to build this test, but all that needs is a sufficiently new compiler.

This then results in the `__ARM_FEATURE_SVE` check always passing, because SVE is explicitly enabled.

However it's perfectly possible for the compiler to support +sve but the machine running the code to not, which results with the test crashing with "Illegal instruction".

Handle this case by checking HWCAP for SVE support, and skipping the test if it is not present.


bregma:
`<sys/auxv.h>` and `getauxval()` are Linux-specific and will fail to compile on FreeBSD and QNX (and any other OS like Solaris).  You'll need to add a configuration-time check for their presence

rossburton:
> `<sys/auxv.h>` and `getauxval()` are Linux-specific and will fail to compile on FreeBSD and QNX (and any other OS like Solaris). You'll need to add a configuration-time check for their presence

Good point.  Easily done with some `#if`ery.

rossburton:
The Linuxisms are now guarded and other platforms easily added.

tests/Garm64-test-sve-signal.c
```diff
@@ -96,8 +101,22 @@ int64_t square(svint64_t z0)
   return res;
 }
 
+bool has_sve(void) {
+#if defined(__linux__)
+  return (getauxval(AT_HWCAP) & HWCAP_SVE) ? true : false;
+#else
+  printf("Cannot determine if SVE is present, assuming it is\n");
+  return true;
```
bregma, line 109: It would be better to assume  the OS does *not* support SVE. That way the test will SKIP instead of crash and FAIL and if a port wants it to PASS they need to fix the test. SKIP means "not supported" and "FAIL" means "supported but broken".

bregma (changes requested):

tests/Garm64-test-sve-signal.c
rossburton: Done

bregma (approved):

Issue #693, opened 2024-01-02
Failed to build from source, aarch64, got /usr/bin/ld: ../src/.libs/libunwind.so: undefined reference to `__aarch64_cas8_acq_rel'


jiridanek:
Full compilation log: https://download.copr.fedorainfracloud.org/results/packit/skupperproject-skupper-router-1014/fedora-rawhide-aarch64/06847035-skupper-router/builder-live.log.gz

```
Making install in tests
make[1]: Entering directory '/builddir/build/BUILD/libunwind-1.8.0-rc1/tests'
  CC       forker.o
  CCLD     forker
  CC       Gperf-simple.o
  CCLD     Gperf-simple
/usr/bin/ld: ../src/.libs/libunwind.so: undefined reference to `__aarch64_cas8_acq_rel'
/usr/bin/ld: ../src/.libs/libunwind.so: undefined reference to `__aarch64_swp1_acq_rel'
/usr/bin/ld: ../src/.libs/libunwind.so: undefined reference to `__aarch64_ldadd8_acq_rel'
/usr/bin/ld: ../src/.libs/libunwind.so: undefined reference to `__aarch64_ldadd4_acq_rel'
collect2: error: ld returned 1 exit status
make[1]: *** [Makefile:1204: Gperf-simple] Error 1
make[1]: Leaving directory '/builddir/build/BUILD/libunwind-1.8.0-rc1/tests'
make: *** [Makefile:618: install-recursive] Error 1
```

jiridanek:
Tried to do

```
+ ./configure --prefix=/builddir/build/BUILD/libunwind-install CFLAGS=-latomic LDFLAGS=-latomic
```

and then I got

```
In file included from aarch64/Los-linux.c:4:
aarch64/Gos-linux.c: In function â€˜_ULaarch64_local_resumeâ€™:
aarch64/Gos-linux.c:41:7: error: â€˜asmâ€™ operand has impossible constraints
   41 |       __asm__ __volatile__ (
      |       ^~~~~~~
make[2]: *** [Makefile:4381: aarch64/Los-linux.lo] Error 1
make[2]: Leaving directory '/builddir/build/BUILD/libunwind-1.8.0-rc1/src'
make[1]: *** [Makefile:4527: install] Error 2
make[1]: Leaving directory '/builddir/build/BUILD/libunwind-1.8.0-rc1/src'
make: *** [Makefile:618: install-recursive] Error 1
[...]

PR #710, opened 2024-01-17
Handle musl on PPC32

### LINKED ISSUE #709: Fails to build for ppc32 on musl with mcontext_t errors
If I build for 32-bit ppc on musl, the build fails:

```
libunwind-1.8.0/src/ppc32/ucontext_i.h:48:62: error: 'mcontext_t' has no member named 'uc_regs'; did you mean 'gregs'?
```

This is a long-standing difference between glibc and musl that both projects refuse to budge on. For example, see https://www.openwall.com/lists/musl/2018/02/22/1.

rossburton:
On Linux, glibc and musl disagree over the layout of the ucontext_t structure.  For more details, see the musl mailing list:

https://www.openwall.com/lists/musl/2018/02/22/1

Add conditionals to handle both the glibc and musl layout of the structures.

Closes #709.

### DIFF
--- a/src/ppc32/Ginit.c
+++ b/src/ppc32/Ginit.c
@@ -42,6 +42,13 @@ static struct unw_addr_space local_addr_space;
 
 unw_addr_space_t unw_local_addr_space = &local_addr_space;
 
+/* glibc and musl disagree over the layout of this struct */
+#ifdef __GLIBC__
+#define _UC_MCONTEXT_REGS(x) x->uc_mcontext.uc_regs
+#else
+#define _UC_MCONTEXT_REGS(x) x->uc_regs
+#endif
+
 static void *
 uc_addr (ucontext_t *uc, int reg)
 {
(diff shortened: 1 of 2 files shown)


bregma (approved):
I have no way of verifying the changes on powerpc-linux-musl targets, but it doesn't break the powerpc-linux-gnu target so I'm OK with getting this in.

rossburton:
Interestingly on ppc32/musl I then need to link to libatomic_ops:

```
../src/.libs/libunwind-ppc32.so: undefined reference to `__atomic_store_8'
../src/.libs/libunwind-ppc32.so: undefined reference to `__atomic_load_8'
../src/.libs/libunwind-ppc32.so: undefined reference to `__atomic_compare_exchange_8'
```

I wonder if the new code is introducing a copy.

bregma:
Sounds like a variant of #693

rossburton:
I noticed a mistake in the precedence order, new iteration posted.

bregma:
OK.

PR #712, opened 2024-01-18
configure: don't disable documentation if latex2man isn't present

rossburton:
latex2man is only needed to generate the manpages from the LaTeX sources, but as the git repository has pregenerated manpages in (and so do the tarballs) this tool is only needed if the sources have been changed.

Instead of turning off documentation,  just warn that the manpages cannot be rebuilt.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -297,8 +297,7 @@ AC_ARG_ENABLE([documentation],
 AC_MSG_RESULT([$enable_documentation])
 AC_PATH_PROG([LATEX2MAN],[latex2man])
 AS_IF([test "x$LATEX2MAN" = "x" && test "x$enable_documentation" != "xno"], [
-  AC_MSG_WARN([latex2man not found. Install latex2man. Disabling docs.])
-  enable_documentation="no";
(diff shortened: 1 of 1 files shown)


bregma (changes requested):
To do this right `doc/Makefile.am` should use $(LATEX2MAN) instead of L2M so if that tool isn't present, the build will not fail if a `.tex` source file has been touched, it just won't do anything. LATEX2MAN is already set by `configure` and substituted into the Makefile, and after all the user is now being warned that documentation will not be built.

It's good that this fixes that docs won't even be installed in the tool is not present. It's bad that builds could break if that tool is not present.

rossburton:
Sorry for failing to get back to this for a year...

How will the build not do anything if the tools are not present? If `latex2man` was not found then `LATEX2MAN` is unset, so the command line won't be valid and will error out if the rules are called:
```
$ make
  GEN      unw_strerror.man
/bin/bash: line 1: -t: command not found
```

The makefiles should definitely be using the variables that `AC_PATH_PROG` sets. I wonder if the Makefile should also have stub rules for when documentation is disabled that simply warns that it can't do anything.

PR #732, opened 2024-03-18
configure.ac: detect if -latomic is required

### LINKED ISSUE #693: Failed to build from source, aarch64, got /usr/bin/ld: ../src/.libs/libunwind.so: undefined reference to `__aarch64_cas8_acq_rel'

bregma:
Some OS runtimes require libatomic be linked in separately to get standard atomic operations to work. Try to detect that at configure time.

Fixes #693

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -94,6 +94,18 @@ dnl Checks for library functions.
 AC_CHECK_FUNCS(dl_iterate_phdr dl_phdr_removals_counter dlmodinfo getunwind \
 		ttrace mincore pipe2 sigaltstack execvpe)
 
+AC_MSG_CHECKING([if -latomic is required])
+  AC_LINK_IFELSE([AC_LANG_PROGRAM([[
+                        #include <stdatomic.h>
+                        atomic_bool ab_ = 0;
+                    ]],[[
+                        atomic_load(&ab_);
+                    ]])],
+                 [use_libatomic=no],
+                 [use_libatomic=yes])
+AC_MSG_RESULT([$use_libatomic])
+AS_IF([test "$use_libatomic" = "yes"],[LIBS="-latomic $LIBS"])
+
 AC_MSG_CHECKING([if building with AltiVec])
 AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
 #ifndef __ALTIVEC__


bertwesarg:
No hidden `.` to be found. LGTM

rossburton:
This doesn't work for my testing on ppc32 with musl:

```
checking if -latomic is required... no
...
ld: ../src/.libs/libunwind-ppc32.so: undefined reference to `__atomic_store_8'
```
Explicitly adding `-latomic` fixes the build.

rossburton:
Now I'm also carrying a patch for ppc32/musl so that it builds (https://github.com/libunwind/libunwind/pull/710) , but with that patch I need to add a fetch from an `atomic_llong` to get it to realise it needs libatomic and successfully link.

Now I don't need to make this change for ppc32/glibc, so it's an open question as to whether this is a quirk of musl vs glibc, or a bug in the musl support.

kraj:
> Now I'm also carrying a patch for ppc32/musl so that it builds (#710) , but with that patch I need to add a fetch from an `atomic_llong` to get it to realise it needs libatomic and successfully link.
> 
> Now I don't need to make this change for ppc32/glibc, so it's an open question as to whether this is a quirk of musl vs glibc, or a bug in the musl support.

I think the issue perhaps is that on glibc libgcc_s.so is a linker stub

```
❯ file /usr/aarch64-linux-gnu/lib64/libgcc_s.so
/usr/aarch64-linux-gnu/lib64/libgcc_s.so: ASCII text

❯ cat /usr/aarch64-linux-gnu/lib64/libgcc_s.so
/* GNU ld script
   Use the shared library, but some functions are only in
   the static library.  */
GROUP ( libgcc_s.so.1 -lgcc )
```
so it can pull in static libgcc during link underneath, I am not sure if we have same for musl, if not thats the reason why you see it on musl and not on glibc.

rossburton:
To make it clear, this change to the patch means we no longer need to add explicit `-latomic` flags:

```
+AC_MSG_CHECKING([if -latomic is required])
+  AC_LINK_IFELSE([AC_LANG_PROGRAM([[
+                        #include <stdatomic.h>
+                        atomic_bool ab_ = 0; atomic_llong al_ = 0;
+                    ]],[[
+                        atomic_load(&ab_); atomic_load(&al_);
+                    ]])],
+                 [use_libatomic=no],
+                 [use_libatomic=yes])
```

PR #745, opened 2024-04-18
configure.ac: stop adding -D__EXTENSIONS__

### LINKED ISSUE #743: configure.ac: incorrect use of IS_SOLARIS causes -D__EXTENSION__ to always be used when compiling
The `configure.ac` file has an incorrect `if $OS_SOLARIS` construct, which causes `-D__EXTENSIONS__` to always appear on every compiler command line. This may have unintended consequences.

The construct should use the `AM_COND_IF()` macro instead.

bregma:
An incorrect use of an automake conditional in a shell `if` construct forced -D__EXTENSIONS__ to be added to all compilation command lines. The AM_COND_IF automake macro was used instead.

Tested against x86_64-linux-gnu and the -D__EXTENSIONS__ no longer appeared on the compile command line.

Closes #743

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -484,12 +484,12 @@ fi
 AC_MSG_RESULT([$intel_compiler])
 
 AC_MSG_CHECKING([if building on Solaris then define __EXTENSIONS__ macro])
-if $OS_SOLARIS; then
+AM_COND_IF([OS_SOLARIS],[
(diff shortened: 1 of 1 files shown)


bertwesarg:
LGTM

bregma:
Replaced by #747 which fixed the problem in a different way.

Issue #742, opened 2024-04-18
[v1.8] aarch64: error: inline assembly requires more registers than available


bregma:
Compile errors on aarch64-linux-* with clang and GCC (when -O2 not specified)
```
  CC       aarch64/Los-linux.lo
In file included from ../../src/aarch64/Los-linux.c:4:
../../src/aarch64/Gos-linux.c: In function ‘_ULaarch64_local_resume’:
../../src/aarch64/Gos-linux.c:41:7: error: ‘asm’ operand has impossible constraints
   41 |       __asm__ __volatile__ (
      |       ^~~~~~~
make[2]: *** [Makefile:4795: aarch64/Los-linux.lo] Error 1
```

This is a clone of #702 for the 1.8 branch.

sdebnath-ebay:
@bregma this is the same issue as discussed in https://github.com/libunwind/libunwind/issues/702, would be great to have this in 1.8.2. Issue #702 is tagged for 1.9.0 - patch fix would be better in 1.8 series. Thanks.

bregma:
This issue is for backporting the patch from the master branch (with a 1.9.0 milestone) to the 1.8 branch (with a 1.8.2 milestone). Github does not allow you to target issues to multiple branches. Nothing goes into a stable release branch without (a) an issue to close and (2) going on to the master branch first.

sdebnath:
Ah, thanks @bregma. Looking forward to it.

TannerFirl:
I see PR #770 was opened to resolve this issue. Are we waiting on something, or could #770 be merged to resolve this issue?

Lnk2past:
Just ran into this too :(

bregma:
Closed by #770

PR #748, opened 2024-04-22
Rework inline aarch64 asm for setcontext

### LINKED ISSUE #702: aarch64: error: inline assembly requires more registers than available
Hm, clang has a different error:
```
error: inline assembly requires more registers than available
   65 |         "ldr x0,  %[x0]\n\t"
      |         ^
```

_Originally posted by @sfc-gh-sgiesecke in https://github.com/libunwind/libunwind/issues/608#issuecomment-1708458429_

---

Hitting the same issue in dotnet-runtime, where arm64 legs are cross-compiled using clang.

bregma:
Modern GC and clang were barfing on the inline asm constraints for the aarch64-linux setcontext() replacement. Reformulated the asm code to reduce the required constraints.

Fixed #702

### DIFF
--- a/src/aarch64/Gos-linux.c
+++ b/src/aarch64/Gos-linux.c
@@ -2,6 +2,7 @@
    Copyright (C) 2008 CodeSourcery
    Copyright (C) 2011-2013 Linaro Limited
    Copyright (C) 2012 Tommi Rantala <tt.rantala@gmail.com>
+   Copyright 2024 Stephen M. Webb  <swebb@blackberry.com>
 
 This file is part of libunwind.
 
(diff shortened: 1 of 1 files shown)


ararslan:
Hi @bregma, is there anything preventing this PR from moving forward? I confirmed that this change (applied to v1.8.1) fixes the build in my case.

Somewhat as an aside, it seems Xamarin has their [own fix for this](https://github.com/xamarin/xamarin-android/commit/a7b576885624d09c4cc9d65412da2a7558fd3b36#diff-2ee3c8cf4e808eb189d302113a4f51a812ef05ff875bb832449b4d9194c58619). I'd be curious to know how their approach differs from yours as implemented in this PR, if at all.

Thanks for working on this!

bregma:
I think the only blocker for this is an attention deficit on my part. I'll merge to main and start on a backport for #742

PR #753, opened 2024-05-05
Fixed miscompilation of unw_getcontext() on ARM

ian-veego:
`unw_getcontext()` followed by `unw_init_local()` produced invalid code due to r1 not being specified as output register in the inline assembly. Compiler assumed that this register already contained the right value of `unw_context_t*` to be passed to `unw_init_local()`.

### DIFF
--- a/include/libunwind-arm.h
+++ b/include/libunwind-arm.h
@@ -288,7 +288,7 @@ unw_tdep_context_t;
     "mov r0, #0\n\t"                                                                            \
     "stmia %[base]!, {r0-r15}\n\t"                                                              \
     VSTMIA                                                                                      \
-    : [r0] "=r" (r0) : [base] "r" (unw_base) : "memory");                                       \
+    : [r0] "=r" (r0), [base] "+r" (unw_base) : : "memory");                                     \
   (int)r0; })
 #else /* __thumb__ */
 #define unw_tdep_getcontext(uc) ({                                        \


bregma:
While the change looks correct, I'm having trouble seeing it make any difference in the generated assembly code. The test suite files all generate identical assembly (using `CFLAGS="-marm -save-temps"` when building) and the test results are the same.

I'm going to commit this because as far as I can see it does no harm and may indeed fix broken code somewhere. I just can;t verify it.

ian-veego:
> While the change looks correct, I'm having trouble seeing it make any difference in the generated assembly code. The test suite files all generate identical assembly (using `CFLAGS="-marm -save-temps"` when building) and the test results are the same.
> 
> I'm going to commit this because as far as I can see it does no harm and may indeed fix broken code somewhere. I just can;t verify it.

I stumbled across this issue when debugging our project on ARM. And I'm quite surprised nobody found it before. And actually the following Thumb section does exactly the same as this patch for the respective mode.

The generated code was as follows:
1. First lines of `unw_getcontext()` load `unw_context_t* uc` into r1
2. `stmia r1! {r0-r15}` increases r1 by 64
3. `unw_init_local()` takes the same uc as its second argument which (as compiler assumes) is already conveniently loaded into r1 that is exactly the register used to pass the second argument per ARM ABI. So the compiler doesn't bother doing anything extra to r1 at this point leading to invalid value being passed to the function.
After the patch the compiler inserts an additional instruction to reload r1 with the correct value just before the call.

PR #771, opened 2024-06-17
Fix signal handling for aarch64-nto-qnx*

### LINKED ISSUE #767: aarch64-nto-qnx* fails to unwind past signal trampoline
**Environment**
- libunwind version: 1.8
- architecture: aarch64
- OS: QNX (7.0, 7.1, 8.0)

**Description**
The unwinder is failing to restore the correct context through the signal trampoline on aarch64-nto-qnx*.

Output of `Gtest-bt -v` looks like this.
```
Backtrace across signal handler:
 @ 0x2b604122d8sighandler: got signal 15, sp=39db798774
        explicit backtrace:
00000016559784ec <do_backtrace+0x8c>              (sp=00000039db7985d0)
        proc=0x1655978460-0x1655978794
        handler=0x0 lsda=0x0 gp=0x12e78
        [/system/xbin/libunwind/Gtest-bt+0x14ec]
00000016559787d0 <sighandler+0x38>                (sp=00000039db798750)
        proc=0x1655978798-0x165597882c
        handler=0x0 lsda=0x0 gp=0x12e78
        [/system/xbin/libunwind/Gtest-bt+0x17d0]
0000002b603e5224 <__signalstub+0x18>              (sp=00000039db798780)
        [/proc/boot/libc.so.5+0x28224]
00000039db7968a4                                  (sp=0000002b604122d0)
0000002b6047e650                                  (sp=0000002b604122d0)
        [/proc/boot/libc.so.5+0xb1650]
```

Note that there is no unit test to verify the correct expected (and fundamental) behaviour. This particular unit test passes because the addresses returned from `backtrace()` match the addresses found through an explicit backtrace, but both are incorrect because the one is implemented in terms of the other.

**Reproduction Steps**
1. Run `Gtest-bt -v`
2. Examine the output expecting to see `SignalKill` and `main` preceding `__signalstub` in the backtrace.
3. See error

bregma:
QNX does not put the signal context at a fixed offset in the signal stack. Need to get the pointer to the context passed by the kernel to the trampoline instead.

Added unit tests to detect this situation (`[GL]test-sig-context`).

Fixes #767

### DIFF
--- a/src/aarch64/Gstep.c
+++ b/src/aarch64/Gstep.c
@@ -2,7 +2,7 @@
    Copyright (C) 2008 CodeSourcery
    Copyright (C) 2011-2013 Linaro Limited
    Copyright (C) 2012 Tommi Rantala <tt.rantala@gmail.com>
-   Copyright 2022 Blackberry Limited.
+   Copyright 2022,2024 Blackberry Limited.
 
 This file is part of libunwind.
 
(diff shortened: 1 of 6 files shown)


bregma:
s390x-linux-gnu fails the new [GL]test-sig-context unit tests. Turns out many of the other tests are false PASSes as well. Removed that target from the mandatory CI pass list until it's addressed and will open an issue.

PR #775, opened 2024-06-19
RFC: Do not fallback to slow_backtrace when tdep_trace stops unwinding

milianw:
On an aarch64 system tdep_trace was returning `-UNW_ESTOPUNWIND` a lot while profiling a Qt 6 / QML application with heaptrack. I have not understood why exactly, but note that QML injects JIT frames and more, which potentially lead to such failures?

Without the patch here, unwinding became excessively slow because the `slow_backtrace` fallback continuously called `tdep_get_elf_image` which is extremely slow for applications with many entries in their `/proc/<pid>/maps` file.

With the patch here applied, I could use heaptrack again with the expected bearable overhead. The backtraces still look fine, so for me this patch here looks like a good workaround. I'm unsure about the implications though, thus this RFC.


bregma (approved):
No, this makes sense: no point in doing a slow backtrace if the fast backtrace definitely knows it can't go any further.

Seems to fail the aarch64-linux-gnu `[GL]test-trace` unit tests in CI. Can you please investigate?

Manually verified on aarch64-nto-qnx7.1.0 and it passes all the backtrace tests.

milianw:
Are the test failures related? I cannot submit with them failing. Can you override this if they are unrelated?

Thanks!

bregma:
The failure is related.

(1) the aarch64 `tdep_trace()` function was not returning the correct count of addresses it had stored when it returns with a failure code. This is a latent bug revealed by your code change.
(2) `unw_backtrace2()` was not exactly doing the right things to prime the `tdep_trace()` function, so the latter always failed and the slow path was always taken regardless, at least until your code change (see (1)).

Here's a patch to your branch that passes all the tests on both *aarch64-linux-gnu* and *x86_64-linux-gnu* targets as well as *aarch64-nti-qnx7.1.0* target. It also has a big change to the Gtest-trace test program to make it less cryptic, both codewise and in its verbose output.

[gh-775.patch.txt](https://github.com/user-attachments/files/16254541/gh-775.patch.txt)

Give that a try and see if it makes it past CI.

bregma:
OK, I pushed the fix for #779 that caused the CI failures with your patch. I you would rebase on the latest HEAD it should be good to go.

milianw:
> OK, I pushed the fix for #779 that caused the CI failures with your patch. I you would rebase on the latest HEAD it should be good to go.

Thanks, did that now. Let's see if the CI agrees :crossed_fingers:

milianw:
damn, it seems there are still quite a lot of failures @bregma. I don't have the time to investigate that right now, maybe in a few weeks after the summer vacation time at the earliest.

milianw:
@bregma ping? can you help me out with the remaining failures?

milianw:
Great, thanks a lot @gleocadie!

@bregma it would be really great if you could chime in again to unblock these patches here

mothacehe:
That one breaks both arm and armhf tests which were just fixed. I would suggest a revert, until this is understood.

PR #787, opened 2024-07-30
add Fix of ELFW in find_build_id_path

aokblast:
There is ELFW in find_build_id_path but I did't find any definition of ElfW, thus I think it is a legacy naming convention.

I have test on Ubuntu 22.04, and
```c
#if defined(ELF_NOTE_GNU) && defined(NT_GNU_BILD_ID)
```
will return false, which cause the gcc emits no error. I think GCC disable generating build-id in default.


bregma:
OK, so your target is arm-none-newlib?

aokblast:
Yes

bregma:
You're going to run into a lot of trouble. libunwind relies on a POSIX OS interface for some of its implementation (eg. `mmap()`) and neither newlib nor an OS of "none" provide a POSIX API.

Anyways, this fix is still correct. In it goes.

PR #791, opened 2024-08-01
Fix crash when reading ELF files with invalid program segment pointers

elkx1:
In some cases, an ELF file could contain program headers with empty segments but non-null offsets (they have p_filesz = 0 but p_offset != 0).

The case that triggered this change is libcap.so.2.48 on Amazon Linux 2023,

```
Name         : libcap
Version      : 2.48
Release      : 2.amzn2023.0.3
Architecture : aarch64
Size         : 1.4 M
Source       : libcap-2.48-2.amzn2023.0.3.src.rpm
```

In this binary, there is a `.gnu_debugdata` section. We can extract that with objcopy and then run `readelf -l` on it to see the program headers:

```
cp /usr/lib64/libcap.so.2.48 .
objcopy --dump-section .gnu_debugdata=minidebuginfo_objcopy.xz
libcap.so.2.48
xz -d minidebuginfo_objcopy.xz
readelf -l minidebuginfo_objcopy
```

Note that the size of the file `minidebuginfo_objcopy` is 5608 bytes.

```
Elf file type is DYN (Shared object file)
Entry point 0x0
There are 11 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
               0x0000000000000380 0x0000000000001fc8  R      0x10000
LOAD           0x0000000000000000 0x0000000000010000 0x0000000000010000
               0x0000000000000000 0x0000000000003d8c  R E    0x10000
LOAD           0x0000000000000000 0x0000000000020000 0x0000000000020000
               0x0000000000000000 0x0000000000001414  R      0x10000
LOAD           0x000000000000fa58 0x000000000003fa58 0x000000000003fa58
               0x0000000000000000 0x00000000000005e0  RW     0x10000
DYNAMIC        0x000000000000fa58 0x000000000003fbc8 0x000000000003fbc8
               0x0000000000000000 0x0000000000000210  RW     0x8
NOTE           0x00000000000002a8 0x00000000000002a8 0x00000000000002a8
               0x0000000000000020 0x0000000000000020  R      0x8
NOTE           0x00000000000002c8 0x00000000000002c8 0x00000000000002c8
               0x00000000000000b8 0x00000000000000b8  R      0x4
GNU_PROPERTY   0x00000000000002a8 0x00000000000002a8 0x00000000000002a8
               0x0000000000000020 0x0000000000000020  R      0x8
GNU_EH_FRAME   0x0000000000010000 0x0000000000020380 0x0000000000020380
               0x0000000000000000 0x000000000000027c  R      0x4
GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
               0x0000000000000000 0x0000000000000000  RW     0x10
GNU_RELRO      0x000000000000fa58 0x000000000003fa58 0x000000000003fa58
               0x0000000000000000 0x00000000000005a8  R      0x1
```

Here we see that some segments are listed as starting at an offset greater than the file size, e.g. the `DYNAMIC` segment, whose offset is `0x000000000000fa58 == 64088`. Note, however, that these segments have a size of 0.

Note further that this is not the same offset as the `DYNAMIC` segment in the outer library `libcap`:

```
DYNAMIC        0x000000000002fbc8 0x000000000003fbc8 0x000000000003fbc8
               0x0000000000000210 0x0000000000000210  RW     0x8
```

so where this number comes from is unknown. Either way, the segment size in the file is 0, so we should never try to read from it.

---

This change introduces a new function `get_program_segment` that can be used to obtain a pointer to the segment within the image for a given program header. If the program header specifies a zero size segment, NULL is returned. Using this function to dereference program segments instead of doing so directly avoids crashes in such cases.

There is also a slight change in behaviour in the `find_build_id_path` function - it was using `p_memsz` field of the program header to find the end of the notes section, but I also believe this to be incorrect and it should be using the `p_filesz` field. Probably benign in most cases but in general the memory size could be larger than the file size, so this error could lead to reading past the end of the image buffer.

### DIFF
--- a/src/dwarf/Gfind_unwind_table.c
+++ b/src/dwarf/Gfind_unwind_table.c
@@ -114,13 +114,13 @@ dwarf_find_unwind_table (struct elf_dyn_info *edi,
 
   if (peh_hdr)
     {
-      if (pdyn)
+      Elf_W(Dyn) *dyn = (Elf_W(Dyn) *)elf_w (get_program_segment) (&edi->ei, pdyn, NULL);
+      if (dyn)
         {
           /* For dynamically linked executables and shared libraries,
              DT_PLTGOT is the value that data-relative addresses are
              relative to for that object.  We call this the "gp".  */
-                Elf_W(Dyn) *dyn = (Elf_W(Dyn) *)(pdyn->p_offset
-                                                 + (char *) edi->ei.image);
+
           for (; dyn->d_tag != DT_NULL; ++dyn)
             if (dyn->d_tag == DT_PLTGOT)
               {
@@ -136,8 +136,13 @@ dwarf_find_unwind_table (struct elf_dyn_info *edi,
            absolute.  */
         edi->di_cache.gp = 0;
 
-      hdr = (struct dwarf_eh_frame_hdr *) (peh_hdr->p_offset
-                                           + (char *) edi->ei.image);
+      hdr = (struct dwarf_eh_frame_hdr *) elf_w (get_program_segment) (&edi->ei, peh_hdr, NULL);
+      if (!hdr)
+        {
+          Debug (1, "table `%s' missing\n", path);
+          return -UNW_ENOINFO;
+        }
+
       if (hdr->version != DW_EH_VERSION)
         {
           Debug (1, "table `%s' has unexpected version %d\n",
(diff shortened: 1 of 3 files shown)


bregma:
Good stuff.

PR #793, opened 2024-08-06
Support `-mbranch-protection=pac-ret` on aarch64

### LINKED ISSUE #647: -mbranch-protection=pac-ret support aarch64
Using `-mbranch-protection=standard` or just `-mbranch-protection=pac-ret` causes `make check` test failures. The additional tests which fail are: `Gtest-trace`, `Ltest-trace`, `Lrs-race`, `Gtest-exc`, `Ltest-exc`, 

Using `-mbranch-protection=none` and `-mbranch-protection=bti` does not add test failures (https://github.com/libunwind/libunwind/issues/512).

`-mbranch-protection=standard` is an arm64 default in the Ubuntu Archive, Debian, and Fedora, and I believe this has caused dovecot FTBFS issues https://bugs.launchpad.net/ubuntu/+source/dovecot/+bug/2036268

https://gcc.gnu.org/onlinedocs/gcc/AArch64-Options.html

elkx1:
- Support for pac-ret on aarch64 existed partially before inside the DWARF parser but this needed applying elsewhere in the code
- Move PAC mask stripping functions to be accessible from more places and rename to `tdep_ptrauth_insn_mask` for consistency with the corresponding accessor function
- Apply PAC stripping in appropriate places
- Add ptrace accessor function implementation for `ptrauth_insn_mask`
- make check is now passing on aarch64 when compiled with `CFLAGS="-mbranch-protection=standard"`
- There is an extra function in the ptrace library ABI now (`_UPT_ptrauth_insn_mask`) but there are also some other ABI changes reported by libabigail / `make abi-check` that are unrelated to this change, so I am not sure whether to proceed with updating the .abi files or not

Closes #647

### DIFF
--- a/src/ptrace/_UPT_accessors.c
+++ b/src/ptrace/_UPT_accessors.c
@@ -35,5 +35,6 @@ unw_accessors_t _UPT_accessors =
     .access_fpreg               = _UPT_access_fpreg,
     .resume                     = _UPT_resume,
     .get_proc_name              = _UPT_get_proc_name,
-    .get_elf_filename           = _UPT_get_elf_filename
+    .get_elf_filename           = _UPT_get_elf_filename,
+    .ptrauth_insn_mask          = _UPT_ptrauth_insn_mask
   };
--- a/include/libunwind-ptrace.h
+++ b/include/libunwind-ptrace.h
@@ -56,6 +56,7 @@ extern int _UPT_get_proc_name (unw_addr_space_t, unw_word_t, char *, size_t,
 extern int _UPT_get_elf_filename (unw_addr_space_t, unw_word_t, char *, size_t,
                                   unw_word_t *, void *);
 extern int _UPT_resume (unw_addr_space_t, unw_cursor_t *, void *);
+extern unw_word_t _UPT_ptrauth_insn_mask (unw_addr_space_t, void *);
 extern unw_accessors_t _UPT_accessors;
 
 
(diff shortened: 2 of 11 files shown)


src/ptrace/_UPT_accessors.c
```diff
@@ -35,5 +35,6 @@ unw_accessors_t _UPT_accessors =
     .access_fpreg               = _UPT_access_fpreg,
     .resume                     = _UPT_resume,
     .get_proc_name              = _UPT_get_proc_name,
-    .get_elf_filename           = _UPT_get_elf_filename
+    .get_elf_filename           = _UPT_get_elf_filename,
+    .ptrauth_insn_mask          = _UPT_ptrauth_insn_mask
```
elkx1, line 39: Should this also be wrapped in `#ifdef UNW_TARGET_AARCH64`? I wasn't sure

a74nh:
.NET uses libunwind and is looking at supporting PAC-RET on AArch64. So this PR is currently a blocker. Discussion is [here](https://github.com/dotnet/runtime/pull/108561). 

Any progress on this PR would be appreciated.
Thanks!

bregma:
OK, I have finally confirmed this fixes PAC-related problems on aarch64-unknown-nto-qnx8.0.0 so it's good to go.

Sorry for the delay.

elkx1:
Thanks very much for your effort on this one!

PR #795, opened 2024-08-16
Fix UMR indicated by valgrind in `unw_address_is_valid`

bjosv:
Validate the start of the _initialized_ address range instead of the first address in that first page.
This fixes the uninitialized memory access/read errors indicated by tools like `valgrind`.

This was first fixed via #454, but after #662 the warnings resurfaced.
See #454 description for more background information and reasons.


bjosv:
The failing CI steps in this PR are similar on master, but just let me know if there are anything I should change or help with.

PR #799, opened 2024-08-27
[Debugging] Do not fallback to slow_backtrace when tdep_trace stops unwinding

gleocadie:
On an aarch64 system tdep_trace was returning `-UNW_ESTOPUNWIND` a lot while profiling a Qt 6 / QML application with heaptrack. I have not understood why exactly, but note that QML injects JIT frames and more, which potentially lead to such failures?

Without the patch here, unwinding became excessively slow because the `slow_backtrace` fallback continuously called `tdep_get_elf_image` which is extremely slow for applications with many entries in their `/proc/<pid>/maps` file.

With the patch here applied, I could use heaptrack again with the expected bearable overhead. The backtraces still look fine, so for me this patch here looks like a good workaround.

### DIFF
--- a/src/mi/backtrace.c
+++ b/src/mi/backtrace.c
@@ -67,7 +67,8 @@ unw_backtrace (void **buffer, int size)
   if (unlikely (unw_init_local (&cursor, &uc) < 0))
     return 0;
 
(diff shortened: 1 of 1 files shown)


gleocadie:
Sorry wrong place :/ (wanted to debug it in my fork)

PR #801, opened 2024-08-28
[RFC] Don't hold reg state cache lock while calling into dynamic loader

artempyanykh:
**Context**: AFAIU this dependency between libunwind and the dynamic loader (DL) is a known problem. I did some digging and testing and it seems that dropping the cache lock before calling into DL and re-acquiring it after is safe. With that said, my understanding of libunwind is limited, so I might be missing something obvious. Would appreciate feedback/review!

**Summary**:
Gparser/`find_reg_state` wants to acquire and hold 2 locks:
1. The `cache.lock`
2. (when the value is not in cache) `dl_load_write_lock` transitively via a call to
   `dl_iterate_phdr`.

But since libunwind doesn't own the 2nd lock and `dl_iterate_phdr` is a public API
this is a recipe for lock ordering inversion.

Looking at the code, it *seems* that libunwind doesn't necessarily *need* to hold the
cache lock while executing the _slow path_ and calling into DL. We can release
the lock before and re-acquire it after calling into DL.

This means several things:
1. (+) No deadlocks between libunwind and DL
2. (-) Not holding the lock while executing the _slow path_ means that multiple
threads can do the same work before the cache is populated 
    a. writing the same value into the cache multiple times if slightly wasteful doesn't seem to be problematic
3. (+) -||- multiple threads can do _different_ work on the _slow path_ in parallel (modulo having to synchronize on `dl_iterate_phdr`),
4. (+) the cache can serve other threads without having to wait for the _slow path_ to finish

**Test plan**:
* Existing suite of tests pass (`make check` on x86_64 Linux)
* Newly added test fails before and passes after the change.

**Appendix**:
The chain of calls is
1. find_reg_state ->
2. lock(cache.lock) // via get_rs_cache
3. fetch_proc_info ->
4. dwarf_find_proc_info ->
5. dl_iterate_phdr ->
6. lock(dl_load_write_lock)


src/dwarf/Gparser.c
```diff
@@ -1022,6 +1022,9 @@ find_reg_state (struct dwarf_cursor *c, dwarf_state_record_t *sr)
     }
   else
     {
+      if (cache)
+        put_rs_cache (c->as, cache, &saved_mask);
```
interwq, line 1026: Hi @djwatson, we encountered a tricky issue and can certainly use your help here. Holding the cache lock while calling into the dynamic loader below could cause deadlocks (since DL may trigger unwind as well). Is dropping and reacquiring the cache the right fix here?

artempyanykh:
@djwatson or @bregma a gentle ping. If you have a bit of time, your input would be very valuable here. The code change is 3 lines in `src/dwarf/Gparser.c`, the rest of this PR are tests.

PR #808, opened 2024-10-05
Fix build on FreeBSD (32-bit arm)

sunpoet:
Reference:	https://cgit.freebsd.org/ports/commit/?id=663a5a08daebb7563af762e032c9776d4b2846aa


sunpoet:
Thanks!

Issue #492, opened 2023-04-28
Issue unwinding with .debug_frame for ARM Yocto 4 build


tennesseedaved:
For Yocto 4 build of libunwind 1.6.2, evidenced by https://github.com/libunwind/libunwind/blob/master/src/dwarf/Gfind_proc_info-lsb.c#L963 and table match failure, the IP segment offset used for lookup appears to be offset by an extra component of segbase.

The patch below appears to resolve lookup, but to date, I have not seen any other reports of encountering this issue which seems so fundamental to arm dwarf unwind behavior. 

The problematic formula is something like this for UNW_INFO_FORMAT_TABLE case:
segment rel ip = actual ip - segbase - (segbase - code vaddr)
For "code vaddr" of 0:
segment rel ip = actual ip - 2*segbase

```
--- a/src/dwarf/Gfind_proc_info-lsb.c
+++ b/src/dwarf/Gfind_proc_info-lsb.c
@@ -911,7 +911,7 @@ dwarf_search_unwind_table (unw_addr_space_t as, unw_word_t ip,
                            int need_unwind_info, void *arg)
 {
   const struct table_entry *e = NULL, *table;
-  unw_word_t ip_base = 0, segbase = 0, last_ip, fde_addr;
+  unw_word_t ip_base = 0, segbase = 0, map_rel_offset = 0, last_ip, fde_addr;
   unw_accessors_t *a;
 #ifndef UNW_LOCAL_ONLY
   struct table_entry ent;
@@ -954,20 +954,27 @@ dwarf_search_unwind_table (unw_addr_space_t as, unw_word_t ip,
   a = unw_get_accessors_int (as);
 
   segbase = di->u.rti.segbase;
+  map_rel_offset = di->load_offset;
   if (di->format == UNW_INFO_FORMAT_IP_OFFSET) {
     ip_base = di->start_ip;
   } else {
     ip_base = segbase;
+    // di->load_offset for UNW_INFO_FORMAT_TABLE is calculated as
+    // segbase (.so load addr) - CODE load virtaddr
+    // already relative to actual address
+    // Not wrong, but results in 2 actual addr relative
+    // transforms if used directly
+    map_rel_offset -= segbase;
   }
 
-  Debug (6, "lookup IP 0x%lx\n", (long) (ip - ip_base - di->load_offset));
+  Debug (6, "lookup IP 0x%lx\n", (long) (ip - ip_base - map_rel_offset));
 
[...]

PR #826, opened 2025-01-25
Fix .debug_frame DWARF unwinding on arm

mothacehe:
Hello,

This fixes two issues that I encountered while trying to have .debug_frame based unwinding working on arm:

- The .debug_frame DWARF based unwinding is currently completely broken on x86_64, arm (and most likely all other architectures). That [commit](https://github.com/libunwind/libunwind/commit/cee5505a99d42a423059a1add8b87b75aeb45daa) is causing the segbase to be substracted twice. Reverting it fixes .debug_frame based unwinding on x86_64 and arm architectures.

- The PC value extracted from signal frames, was decremented by one on ARM, which is wrong. That's because the `use_prev_instr` boolean was not set as for other architectures. I did put that into evidence by missing an FDE that was right on PC.

Thanks,

Mathieu

### DIFF
--- a/src/arm/Gos-linux.c
+++ b/src/arm/Gos-linux.c
@@ -121,6 +121,7 @@ arm_handle_signal_frame (unw_cursor_t *cursor)
   dwarf_get (&c->dwarf, c->dwarf.loc[UNW_ARM_R15], &c->dwarf.ip);
 
   c->dwarf.pi_valid = 0;
+  c->dwarf.use_prev_instr = 0;
 
   return 1;
 }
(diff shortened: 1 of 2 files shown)


mothacehe:
> Would this fix #492?

Yes, I have been using libunwind on Yocto to test that one. I can confirm that it fixes .debug_frame support on ARM with Yocto.

gleocadie:
@mothacehe is it possible to add a test in libunwind test suite ?

mothacehe:
> @mothacehe is it possible to add a test in libunwind test suite ?

Yes, I can add a test regarding .debug_frame based unwinding support. Regarding the PC issue on ARMv7, it will be more tricky to reproduce that error consistently.

mothacehe:
@gleocadie, #836 is fixing the Gtest-bt test that is currently failing on ARM. As the Gtest-bt test is based on the default EXIDX unwinder, it also introduces a new Garm-test-debug-frame-bt test that is running the same test with the DWARF, debug_frame based unwinder.

As exepected #826 is fixing that new test.

bregma:
Could you please rebase to current HEAD on `main` to pick up the fix for CI corss-build failures?

PR #836, opened 2025-02-24
Fix the backtrace test on ARM

mothacehe:
Hello,

The `Gtest-bt` test is currently failing on ARM. That's because the last `unw_step` returns cantunwind when using the default EXIDX unwinder. Ignore that specific error when using that unwinder on ARM.

That fixes the backtrace test on ARM, but then it is still a very weak test because it doesn't check for the number of frames in the backtrace. It can very be that there is only one single frame, when we are expecting at least three frames for the various backtraces that are created.

Improve the test so that it checks that the backtraces that are produced have at least three frames.

Finally, there is nothing testing the DWARF, .debug_frame based unwinder on ARM that can be used by setting the `UNW_ARM_UNWIND_METHOD=1` environment variable. Introduce a new test, `Garm-test-debug-frame-bt.c` that is running the `Gtest-bt` test with that environment variable set.

The test is currently failing, but is fixed by #826 

Thanks,

Mathieu

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,5 +87,6 @@ tests/x64-unwind-badjmp-signal-frame
 tests/[GL]arm64-test-sve-signal
 tests/aarch64-test-plt
 tests/aarch64-test-frame-record
+tests/[GL]arm-test-debug-frame-bt
 tests/*.log
 tests/*.trs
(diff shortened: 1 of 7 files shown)


bregma:
You're going to need to unhide `unwi_unwind_method()` in `src/arm/Gglobal.c`.

mothacehe:
> You're going to need to unhide `unwi_unwind_method()` in `src/arm/Gglobal.c`.

Done, thanks. I missed that one because I was using tests linked to the static library.

bregma:
Now that you've added a new symbol to the ABI it's failing the ABI check. That will also need updating.

mothacehe:
> Now that you've added a new symbol to the ABI it's failing the ABI check. That will also need updating.

Fixed as well.

PR #843, opened 2025-03-27
Try unwinding when RIP is null

gleocadie:


src/x86_64/Gstep.c
```diff
@@ -300,6 +309,8 @@ unw_step (unw_cursor_t *cursor)
         }
       if (!DWARF_IS_NULL_LOC (c->dwarf.loc[RIP]))
         {
+          // GUESSED or STANDARD
+          c->frame_info.frame_type = UNW_X86_64_FRAME_GUESSED;
```
gleocadie, line 313: Used GUESSED to force address validation

Issue #857, opened 2025-04-28
Support for FreeBSD 15 riscv64


rdunkle34:
Are there plans for porting for FreeBSD 15 riscv64 ?
I am running a VisionFive2.

rdunkle34:
I am a sysadmin.  I am happy to test anything.

bregma:
When you tried building it, what error s did you get?

rdunkle34:
There are no riscv repositories at this time.  You have to do this yourself.  I am just pull from main.  Then I have my own repository.
<root@riscv>:/usr/ports# git show
commit 1ee887feb01190ee493c2373e2fb0d6b03fda39f (HEAD -> main, origin/main, origin/HEAD)
Author: Hiroki Tagato <tagattie@FreeBSD.org>
Date:   Sat May 24 15:09:02 2025 +0900

FreeBSD riscv.smallcatbrain.com 15.0-CURRENT FreeBSD 15.0-CURRENT #0 main-n277212-79e70e12f5c9: Tue May 13 08:20:21 EEST 2025

rdunkle34:
I need 15 to get the nics to work.

rdunkle34:
one warning:
  CC       ptrace/_UPT_get_dyn_info_list_addr.lo
ptrace/_UPT_get_dyn_info_list_addr.c:85:3: warning: Implement get_list_addr(), please. [-W#warnings]
   85 | # warning Implement get_list_addr(), please.
------------------
Making check in src
make  check-am
Making check in tests
make  test-proc-info test-static-link  test-strerror Gtest-bt Ltest-bt  Gtest-init Ltest-init  Gtest-concurrent  Ltest-concurrent  Gtest-sig-context  Ltest-sig-context  Gtest-trace Ltest-trace  test-async-sig  test-flush-cache  test-init-remote test-mem  test-reg-state  Ltest-varargs  Ltest-nomalloc  Ltest-nocalloc Lrs-race  Ltest-mem-validate  Ltest-init-local-signal Gtest-exc  Ltest-exc  Gtest-resume-sig  Ltest-resume-sig  Gtest-resume-sig-rt  Ltest-resume-sig-rt test-ptrace  test-setjmp         check-namespace.sh run-ptrace-mapper run-ptrace-misc
`test-proc-info' is up to date.
`test-static-link' is up to date.
`test-strerror' is up to date.
`Gtest-bt' is up to date.
`Ltest-bt' is up to date.
`Gtest-init' is up to date.
[...]

PR #866, opened 2025-05-23
add initial freebsd riscv support

kasperk81:
contributes to https://github.com/libunwind/libunwind/issues/857

### DIFF
--- a/include/tdep-riscv/jmpbuf.h
+++ b/include/tdep-riscv/jmpbuf.h
@@ -23,8 +23,6 @@ LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 
-#if defined __linux__
-
 /* https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/riscv/setjmp.S;h=0b92016b311b11aa9eeb62b38c670a262f1924c9;hb=HEAD */
 #define JB_SP           13
 #define JB_RP           0
@@ -45,5 +43,3 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 #else
 # error "Add offsets here"
 #endif /* __riscv_xlen */
-
-#endif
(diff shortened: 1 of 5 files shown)


rdunkle34:
make (couple of warnings)   
install appears normal
now compiling programs that will use libunwind
[make.log](https://github.com/user-attachments/files/20421567/make.log)
[install.log](https://github.com/user-attachments/files/20421568/install.log)

include/tdep-riscv/jmpbuf.h
```diff
@@ -23,8 +23,6 @@ LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
 OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 
-#if defined __linux__
-
 /* https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/riscv/setjmp.S;h=0b92016b311b11aa9eeb62b38c670a262f1924c9;hb=HEAD */
 #define JB_SP           13
 #define JB_RP           0
```
zhaofengli, line 28: This is specifically for Linux with glibc (should have added a `__GLIBC__` ifdef). FreeBSD has the following:

- https://github.com/freebsd/freebsd-src/blob/bada3811884cb22d088c8b8e4dbc39227682ce48/sys/riscv/include/setjmp.h#L40
- https://github.com/freebsd/freebsd-src/blob/bada3811884cb22d088c8b8e4dbc39227682ce48/lib/libc/riscv/gen/_setjmp.S#L42
- `man 3 setjmp`

src/riscv/offsets.h
```diff
@@ -1,13 +1,7 @@
-#ifdef __linux__
-
 /* Linux-specific definitions: */
 
 /* The RISC-V ucontext has the following structure:
 
    https://github.com/torvalds/linux/blob/44db63d1ad8d71c6932cbe007eb41f31c434d140/arch/riscv/include/uapi/asm/ucontext.h
 */
 #define UC_MCONTEXT_REGS_OFF 176
```
zhaofengli, line 7: Same applies here, and you will actually need to do the correct calculation in `src/riscv/Gstep.c`. At the very least, instead of just removing the `#ifdef __linux__`, add a `__FreeBSD__` check which emits nothing (currently `riscv_handle_signal_frame` just returns `-UNW_EUNSPEC` on non-Linux).

PR #870, opened 2025-05-24
Fix freebsd arm64 build

kasperk81:


giordano:
Probably integrating https://github.com/libunwind/libunwind/issues/873#issuecomment-2912205211 (that patch as is breaks non-FreeBSD systems, do it needs some conditionals like those introduced here) would be the best solution?

bregma:
Superseded by #887

PR #871, opened 2025-05-24
Implement Gresume for freebsd riscv64

kasperk81:
contributes to https://github.com/libunwind/libunwind/issues/857


zhaofengli:
You shouldn't fix a warning with code that's incorrect. I think we should have a riscv-freebsd branch and iterate until it's ready (like #267). PRs like #866 and this shouldn't be merged to main - If we cut a release with this code, people may expect it to work then encounter weird corruptions and crashes because it doesn't actually work.

PR #881, opened 2025-06-02
Fix all arm and armhf tests.

mothacehe:
This fixes all the broken tests on both arm and armhf architectures by:
- Removing pre-linking support, which is deprecated and breaks arm support
- Disabling test-async-sig on arm and armhf because it generates random backtraces that can fail when involving glibc functions without .EXIDX directives.


bregma:
Confirmed that prelink doesn't even work on my x86_64-linux desktop because it can't handle the VDSO.

PR #882, opened 2025-06-03
Revert "Do not fallback to slow_backtrace when tdep_trace stops"

mothacehe:
This reverts commit 5b195ffd5020de5dac19d550c44b58201c69e5b4 because it breaks both arm and armhf test suites.

### DIFF
--- a/src/mi/backtrace.c
+++ b/src/mi/backtrace.c
@@ -67,8 +67,7 @@ unw_backtrace (void **buffer, int size)
   if (unlikely (unw_init_local (&cursor, &uc) < 0))
     return 0;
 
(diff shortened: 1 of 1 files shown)


gleocadie:
Good catch.

mothacehe:
> Good catch. But what's weird is that the test suites were already failing before this PR was merged .

There were some failures before that one but it made it worse by breaking `G/LTest-trace`, `Ltest-varargs` and `Lrs-race`. Since 62dfb323390c0b81dc1b9777258fc4f4080cb254 and 7718ecc1ddce054f84a5741ff664c241cdbe6da7 all arm and armhf tests should be passing.

Issue #873, opened 2025-05-27
Compilation broken on aarch64-freebsd in v1.8.2


giordano:
**Environment**
- libunwind version: v1.8.2
- architecture: aarch64
- OS: FreeBSD (both 14.1 and 15.0)

**Description**

Version v1.8.2 broke compilation on aarch64-freebsd:

```
  CC       ptrace/_UPT_ptrauth_insn_mask.lo
ptrace/_UPT_ptrauth_insn_mask.c:32:16: error: variable has incomplete type 'struct iovec'
   32 |   struct iovec iovec;
      |                ^
ptrace/_UPT_ptrauth_insn_mask.c:32:10: note: forward declaration of 'struct iovec'
   32 |   struct iovec iovec;
      |          ^
ptrace/_UPT_ptrauth_insn_mask.c:38:17: error: use of undeclared identifier 'PTRACE_GETREGSET'
   38 |   ret = ptrace (PTRACE_GETREGSET, pid, NT_ARM_PAC_MASK, &iovec);
      |                 ^
ptrace/_UPT_ptrauth_insn_mask.c:38:40: error: use of undeclared identifier 'NT_ARM_PAC_MASK'
   38 |   ret = ptrace (PTRACE_GETREGSET, pid, NT_ARM_PAC_MASK, &iovec);
      |                                        ^
3 errors generated.
```

**Reproduction Steps**

```sh
./configure
gmake
```

giordano:
Oh, this may be fixed by #870, too.

PR #887, opened 2025-06-09
Avoid NT_ARM_PAC_MASK on older Linux kernels

phongn:
Older kernels still in wide use (e.g. Enterprise Linux 8) do not have `NT_ARM_PAC_MASK` support or the associated structs. Compilation fails:

```
ptrace/_UPT_ptrauth_insn_mask.c: In function ‘_UPT_ptrauth_insn_mask’:
ptrace/_UPT_ptrauth_insn_mask.c:38:40: error: ‘NT_ARM_PAC_MASK’ undeclared (first use in this function); did you mean ‘EF_ARM_EABIMASK’?
   ret = ptrace (PTRACE_GETREGSET, pid, NT_ARM_PAC_MASK, &iovec);
                                        ^~~~~~~~~~~~~~~
                                        EF_ARM_EABIMASK
ptrace/_UPT_ptrauth_insn_mask.c:38:40: note: each undeclared identifier is reported only once for each function it appears in
```

This patch disables this codepath unless this is a Linux host and `NT_ARM_PAC_MASK` is defined in `elf.h`. 

See also #873 which proposes a fix for FreeBSD platforms. This patch is a superset of #870.

### DIFF
--- a/src/ptrace/_UPT_ptrauth_insn_mask.c
+++ b/src/ptrace/_UPT_ptrauth_insn_mask.c
@@ -22,7 +22,7 @@
  */
 #include "_UPT_internal.h"
 
-#if defined(UNW_TARGET_AARCH64) && defined(__linux__) && defined(NT_ARM_PAC_MASK)
+#if defined(UNW_TARGET_AARCH64) && defined(NT_ARM_PAC_MASK)
 
 unw_word_t _UPT_ptrauth_insn_mask (UNUSED unw_addr_space_t as, void *arg)
 {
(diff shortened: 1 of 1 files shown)


bregma:
If a check for `NT_ARM_PAC_MASK` not enough (*ie* is the check for the OS kernel necessary)?

phongn:
> After some investigation I believe it should be `#if defined(UNW_TARGET_AARCH64) && defined(NT_ARM_PAC_MASK)`. Compilation support for that macro is independent of the target operating system.
> 
> Please make that change and I would be happy to commit.

Changed committed. Once merged I'd like to issue a PR to backport this change and some of the `UNUSED` fixes into `v1.8-stable`

PR #893, opened 2025-06-20
Integer Overflow: fixes bit shifting 32bit with subsequent expansion to 64bit

GermanAizek:
@bregma,
I hope that no one will take advantage any integer overflow with malicious intentions, so it is advisable to close possible vulnerabilities.

- 1u - unsigned int 32-bit
- 1ULL - unsigned wide long 64-bit 

References:

-  https://cwe.mitre.org/data/definitions/190.html

Example same type real CVE vuln in Linux Kernel with Integer Overflow:

- https://nvd.nist.gov/vuln/detail/CVE-2022-49748
- https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=08245672cdc6505550d1a5020603b0a8d4a6dcc7

### DIFF
--- a/src/aarch64/Gtrace.c
+++ b/src/aarch64/Gtrace.c
@@ -70,7 +70,7 @@ trace_cache_free (void *arg)
   }
   tls_cache_destroyed = 1;
   tls_cache = NULL;
-  mi_munmap (cache->frames, (1u << cache->log_size) * sizeof(unw_tdep_frame_t));
+  mi_munmap (cache->frames, (1ULL << cache->log_size) * sizeof(unw_tdep_frame_t));
   mempool_free (&trace_cache_pool, cache);
   Debug(5, "freed cache %p\n", cache);
 }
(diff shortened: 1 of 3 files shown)


GermanAizek:
> I see the same problem in `src/aarch64/Gtrace.c`. Could you also apply the same fix please?

indeed, there are also other sections subsequent expansion with unsigned int shift.

GermanAizek:
@bregma, I also found a similar issue in `arm` target and fixed other shifts.

https://github.com/libunwind/libunwind/pull/893/commits/b97766513499383f4190c6950ba17cd27e5a7bef

bregma:
The arm case is a little problematic because it's unsigned 32-bit int, but at least it's going to move from undefined behaviour (left-shifting past 32 bits) to implementation-defined behaviour (converting from a 64-bit value to a 32-bit value). Fact is, if the value goes past about 8 bits we're probably already in trouble in the code in question so it really is moot.

I'll commit this as-is. Thanks.

PR #905, opened 2025-08-11
Use unw_word_t instead of unsigned long for registers

hjl-tools:
Use unw_word_t instead of unsigned long in _Unwind_SetGR, _Unwind_GetGR and _Unwind_GetIP since the full integer register fits unw_word_t, not unsigned long.


mattst88:
I think this is probably an issue on mips/n32 where registers are 64-bit but `unsigned long` is 32-bit.

PR #932, opened 2025-12-06
libunwind-ppc: include sys/reg.h

neheb:
Needed for a definition of __WORDSIZE under musl.


neheb:
seems to not be needed.

neheb:
Turns out it is.

PR #965, opened 2026-03-27
Fix unwinding through vDSO functions on aarch64

lexprfuncall:
This PR adds stack pointer tracking to the existing forward scan that
tracks the frame record in order to determine the frame state.
                                                                                
The .eh_frame and .eh_frame_hdr information for vDSO functions is
discarded by the Linux kernel
                                                                                
https://github.com/torvalds/linux/blob/46b513250491a7bfc97d98791dbe6a10bcc8129d/arch/arm64/kernel/vdso/vdso.lds.S#L77-L80                                 
                                                                                
and vDSO functions are also compiled without frame pointers. That
combination means that unwinding through an interrupted vDSO call can
leave the CFA in the wrong location, leading subsequent unwind steps
to use the wrong offsets for the register state and causing a SIGSEGV.

### DIFF
--- a/src/aarch64/Gstep.c
+++ b/src/aarch64/Gstep.c
@@ -156,22 +156,389 @@ unw_is_plt_entry (unw_cursor_t *uc)
 	return _is_plt_entry (&((struct cursor *)uc)->dwarf);
 }
 
-/* Recognise when a frame record storing FP+LR has been created and whether FP
-   has been updated to point to the frame record. For example:
-  4183d4:       a9bd7bfd        stp     x29, x30, [sp,#-48]!    <= FP+LR stored
-  4183d8:       d2800005        mov     x5, #0x0
-  4183dc:       d2800004        mov     x4, #0x0
-  4183e0:       910003fd        mov     x29, sp                 <= FP updated
-  4183e4:       a90153f3        stp     x19, x20, [sp,#16]
-  ...
-  418444:       a94153f3        ldp     x19, x20, [sp,#16]
-  418448:       f94013f5        ldr     x21, [sp,#32]
-  41844c:       a8c37bfd        ldp     x29, x30, [sp],#48      <= FP+LR retrieved
-  418450:       d65f03c0        ret
-*/
+typedef enum
+  {
+    SP_UNCHANGED,
+    SP_ADJUSTED,
+    SP_UNKNOWN,
+  } sp_state_t;
+
+/* Detect a single straight-line stack decrement that has not been undone by a
+   matching restore or invalidated by ambiguous control flow.  When no frame
+   record is found, the SP adjustment is returned via the offset field so the
+   caller can correct the CFA for LR fallback.
+
+   From the linux aarch64 vDSO:
+
+      <__kernel_gettimeofday>:
+      d10043ff  sub  sp, sp, #0x10         <= SP adjusted
+      b4000680  cbz  x0, ...
+      ...
+      910043ff  add  sp, sp, #0x10         <= SP restored
+      d65f03c0  ret  */
+static void
+track_stack_pointer (uint32_t insn, sp_state_t *state, int32_t *offset)
+{
+  if (*state == SP_UNKNOWN)
+    return;
+
+  if (*state == SP_UNCHANGED
+      && ((insn & 0x7c000000) == 0x14000000       /* B, BL */
+          || (insn & 0xff000000) == 0x54000000    /* B.cond, BC.cond */
+          || (insn & 0x7e000000) == 0x34000000    /* CBZ, CBNZ */
+          || (insn & 0x7e000000) == 0x36000000    /* TBZ, TBNZ */
+          || (insn & 0xfe9fc000) == 0xd61f0000    /* BR, BLR, RET (+PAC) */
+          || (insn & 0xfe9f8000) == 0xd69f0000))  /* ERET, DRPS */
+    {
+      *state = SP_UNKNOWN;
+      *offset = 0;
+    }
+
+  /* SUB SP, SP, #imm
+
+     | 1 | 1 | 0 | 10001 | shift |     imm12     |   Rn  |   Rd  |
+
+     Rn == SP (x31) => 0b11111
+     Rd == SP (x31) => 0b11111.
+
+     Using a bitmask of 0xff0003ff, the masked instruction would be
+     0xd10003ff. */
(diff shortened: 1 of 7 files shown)


lexprfuncall:
Hi @bregma, this has been soaking in production for us for a few weeks and it has resolved our libunwind crashes on our aarch64 servers.  The code to scan the instruction stream for SP changes is subtle so please let me know if there is anything I can do to help out with the review process.

mattst88:
Any idea if it'd be possible to make a unit test that verifies this doesn't regress?

lexprfuncall:
> Any idea if it'd be possible to make a unit test that verifies this doesn't regress?

Definitely so and the PR includes new unit tests which should cover the new functionality.  If it looks like there is something missing, please let me know!  I would love to make sure the test coverage is thorough so this corner of the unwinder stays working.

mattst88:
Oh, sorry! I was reviewing multiple PRs and got confused :)

I'll have a closer look!

lexprfuncall:
> Oh, sorry! I was reviewing multiple PRs and got confused :)
> 
> I'll have a closer look!

Thank you!  For what it's worth, I think this also addresses some of the other aarch64 issues that have open PRs.

src/aarch64/Gstep.c
```diff
@@ -156,193 +156,439 @@ unw_is_plt_entry (unw_cursor_t *uc)
...
+         where X can be 0/1 above to indicate the instruction variant
+         (post-index, pre-index or signed offset).
+
+         Rt2 == LR (x30) => 0b11110
+         Rn  == SP (x31) => 0b11111
+         Rt  == FP (x29) => 0b11101.
+
+         So using a bitmask of 0xfe407fff, the masked instruction would
+         be 0xa8007bfd */
+      if ((insn & 0xfe407fff) == 0xa8007bfd || (insn & 0xffc07fff) == 0xa9807bfd)
```
mattst88: This is from a Claude Code review of the PR, so take it with a grain of salt:

> Potential redundant conditions in track_frame_record: Both the STP and LDP detection blocks now have the form:
>
> ```
>   if ((insn & 0xfe407fff) == 0xa8007bfd || (insn & 0xffc07fff) == 0xa9807bfd)
> ```
>
>   The first mask (0xfe407fff) doesn't check bits 24 and 23, making it match post-index, signed-offset, and pre-index variants already. The second OR condition (0xa9807bfd, the pre-index form) appears to be redundant. It's harmless but should be verified or documented.

tests/aarch64-test-frame-record.c
```diff
@@ -254,7 +254,7 @@ main ()
     /* IP is pointing to instruction that stores FP and LR */
     c->dwarf.ip = (unw_word_t) (instructions+IpStp);
     fs = get_frame_state(&cursor);
-    if (fs.loc != NONE || fs.offset != 0) return -1;
+    if (fs.loc != NONE || fs.offset != 0x20) return -1;
```
mattst88, line 259: There's some unstated assumption here. Some comment explaining where `0x20` comes from would be good.

tests/aarch64-test-frame-record.c
```diff
@@ -309,7 +309,7 @@ main ()
...
```
mattst88, line 316: There's some unstated assumption here. Some comment explaining where `0x20` comes from would be good.

mattst88 (commented):
This looks solid to me (especially the new tests).

tests/aarch64-test-frame-record.c
lexprfuncall: The 0x20 here should be coming from parsing the immediate field of the `SUB SP, SP, #x20` instruction up on line 233.  I'll add a comment about that so the number doesn't seem to come from out of nowhere.

tests/aarch64-test-frame-record.c
lexprfuncall: Same case as the one above.  This time it's based on the `SUB SP, SP, #0x20` on line 288.  Also deserves a comment.

src/aarch64/Gstep.c
lexprfuncall: I think it's right.  The RHS of the `||` should be covered by the LHS making it redundant.  Let me upload a fix that drops the RHS and update the tests to make sure that all 6 cases are tested.  (It looks like we just get 3.)

lexprfuncall (commented):
Thanks for the feedback.  I'll upload a new revision tonight or tomorrow that addresses each of the issues you raised.

lexprfuncall:
@mattst88 Thanks again for the feedback.  I've made the changes based on your feedback and have also rebased on top of the upstream head.

mattst88:
Thanks, looks good to me. Hopefully @bregma will have a look soon. (I've got some PRs waiting to be merged as well :)

bregma:
For the record test results on aarch64-nto-qnx8.0.0
```
# ./test-runner
TAP version 14
0 ok - test-proc-info
1 ok - test-static-link
2 ok - test-strerror
3 ok - test-eh-frame-hdr-sdata8
4 ok - Gtest-bt
5 ok - Ltest-bt
6 ok - Gtest-init
7 ok - Ltest-init
8 ok - Gtest-concurrent
9 ok - Ltest-concurrent
10 ok - Gtest-sig-context
11 ok - Ltest-sig-context
12 ok - Gtest-trace
13 ok - Ltest-trace
14 ok - Gtest-get_proc_name
15 ok - test-async-sig
16 ok - test-flush-cache
17 ok - test-init-remote
18 ok - test-iterate-phdr-reentry
19 ok - test-iterate-phdr-cache-null
20 ok - test-mem
21 ok - test-reg-state
22 ok - Ltest-varargs
23 ok - Ltest-nomalloc
24 ok - Ltest-nocalloc
25 ok - Lrs-race
26 ok - test-getcontext-gp # SKIP
27 ok - Ltest-mem-validate
28 ok - Ltest-init-local-signal
29 ok - test-elf64-gnu-hash
30 not ok - Garm64-test-sve-signal returned 1
31 not ok - Larm64-test-sve-signal returned 1
32 ok - aarch64-test-plt
33 ok - Laarch64-test-frame-record
34 ok - Gaarch64-test-frame-record
35 ok - Laarch64-test-frame-size
36 ok - Gaarch64-test-frame-size
37 ok - aarch64-test-lr-fallback
1..38
```
(the SVE failures are not relevant)

PR #984, opened 2026-04-09
mips: Fix O32/N32 unwinding and implement unw_resume()

mattst88:
Fix several MIPS unwinding bugs across O32, N32, and N64, implement the previously-stubbed `unw_resume()`, and add N32 CI coverage.

### O32 fixes
- Fix undefined behavior in DWARF access path (`read_s32`/`write_s32` shifted 32-bit values by 32, which is UB on native O32 where `unw_word_t` is `uint32_t`)
- Fix `access_reg()` to use full 64-bit `greg_t` slots (on big-endian O32, writing only 32 bits landed in the wrong half of the 64-bit slot, causing `setcontext()` to restore garbage)
- Implement `unw_resume()` (was a stub returning `-UNW_EINVAL`; now copies cursor state back to ucontext and calls `setcontext()`)

### Unwinding improvements (O32 and N32)
- Guard `_step_n64()` with `#if _MIPS_SIM == _ABI64` and move register defines to file scope (prep for fallback function)
- Add `$ra`-based fallback when DWARF unwinding fails (O32/N32 previously returned the raw error instead of gracefully terminating the unwind chain)

### N32 fixes
- Fix N32 signal frame detection (`read_mips_instr()` incorrectly used the N64 code path assuming 64-bit `unw_word_t`, and `unw_is_signal_frame()` had no N32 case with the correct `rt_sigreturn` syscall number)

### CI
- Add MIPS N32 cross-build and test targets (both little-endian and big-endian)

### DIFF
--- a/.github/workflows/CI-linux-gnu.yml
+++ b/.github/workflows/CI-linux-gnu.yml
@@ -106,6 +106,8 @@ jobs:
           - {target: hppa,     host: hppa-linux-gnu,          qemu: hppa,     gccver: 14, container: ubuntu:26.04 }
           - {target: mips64el, host: mips64el-linux-gnuabi64, qemu: mips64el, gccver: 14, container: ubuntu:24.04 }
           - {target: mips64,   host: mips64-linux-gnuabi64,   qemu: mips64,   gccver: 14, container: ubuntu:24.04 }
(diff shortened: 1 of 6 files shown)


mattst88:
Huh, this appears to make `Ltest-exc` timeout on `mips-linux-gnu`. Investigating...

mattst88:
Okay, I went a little farther than initially planned. But now the 4 mips CI jobs (and two new N32 ABI ones) are all passing.

PR #985, opened 2026-04-10
tests: Add minimum frame count checks to detect silent unwind failures

mattst88:
Several tests only checked if unw_step() returned a negative error code, but not whether a reasonable number of frames were found. When a bug caused unw_step() to silently return 0 (end-of-chain) after just 1 frame, these tests passed despite the unwinder being broken.

Add minimum frame count checks (>= 3 frames) to tests that lacked them, matching the existing pattern in Gtest-bt.c. This ensures that bugs which cause premature unwind termination are caught rather than silently producing false passes.


mattst88:
I ran into cases on alpha and mips where unw_step() incorrectly returned 0 and the test passed.

PR #986, opened 2026-04-10
alpha: Fix getcontext saving wrong GP due to PLT lazy binding

mattst88:
On Alpha, GP ($29) is set up in each function's prologue from the procedure value register ($27/t12). When _Ualpha_getcontext is called across a shared library boundary with lazy PLT binding, the dynamic linker's PLT resolver modifies GP before getcontext can save it. The getcontext assembly function has no GP-setup prologue (it immediately saves all registers), so it captures the corrupted GP value.

Later, unw_resume restores this wrong GP via setcontext. The resumed code then performs GP-relative GOT accesses using a shared library's GP instead of the application's GP, causing a SIGSEGV.

Fix this by replacing the unw_tdep_getcontext macro (which directly called the assembly function) with a static inline wrapper that:
1. Captures GP via inline asm before the cross-module call
2. Calls the assembly _Ualpha_getcontext (which may go through PLT)
3. Overwrites the (potentially wrong) saved GP with the correct value

Since the wrapper is static inline in the header, it compiles in the caller's translation unit where GP is still valid.

This was not caught during development on Gentoo because Gentoo links with -Wl,-z,now by default, which forces eager PLT binding and avoids the lazy resolution that trashes GP.

Fixes: Gtest-exc and Ltest-exc SIGSEGV on alpha-linux-gnu CI.

### DIFF
--- a/include/libunwind-alpha.h
+++ b/include/libunwind-alpha.h
@@ -166,11 +166,27 @@ unw_tdep_proc_info_t;
 #include "libunwind-dynamic.h"
 #include "libunwind-common.h"
 
-#define unw_tdep_getcontext             UNW_ARCH_OBJ(getcontext)
 #define unw_tdep_is_fpreg               UNW_ARCH_OBJ(is_fpreg)
 
-extern int unw_tdep_getcontext (unw_tdep_context_t *);
+extern int UNW_ARCH_OBJ(getcontext) (unw_tdep_context_t *);
 extern int unw_tdep_is_fpreg (int);
+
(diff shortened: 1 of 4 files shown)


mattst88:
Same problem occurred on mips, so I copied the fix there and added a test. Confirmed the test fails without these fixes.

mattst88:
For posterity: I thought HPPA might have the same problem, but it doesn't. Apparently it operates somewhat differently and its PLT trampoline never touches the "DP" register.

mattst88:
@bregma: Any concerns about this one?

PR #991, opened 2026-04-16
PowerPC fixes

mattst88:
A large series that cleans up the [superfund](https://en.wikipedia.org/wiki/Superfund) site that is PowerPC support.

- tests: drop _U${plat}_getcontext check for ppc
- tests: disable tail-call optimization for C++ tests
- ppc: fix dwarf_get/put assertions on mismatched FP/V loc types
- ppc: detect signal trampoline without leading addi instruction
- ppc: fix signal frame detection on 32-bit
- ppc32: fix ucontext gregs index definitions
- ppc32: fix DWARF register numbering to match GCC .eh_frame
- ppc32: rewrite signal frame unwinding
- ppc32: make unw_getcontext save the caller's IP as NIP
- ppc32: implement local_resume and establish_machine_state
- ppc32: validate register number in tdep_access_reg
- ppc32: handle non-RT (legacy) signal frames
- ppc64: tag FP and vector register locs in signal frame handling
- ppc64: check for signal frame before DWARF-based unwinding
- ppc64: resume through signal frames via setcontext instead of rt_sigreturn
- ppc64: capture caller's TOC pointer in unw_getcontext
- ppc64: recognise BE ELFv1 PLT call stubs in _is_plt_entry
- tests: enable ppc64-test-plt on big-endian
- tests: fix fall-through in ppc64-test-altivec count==0 branch
- tests: fix stack buffer overflow reading V register in ppc64-test-altivec
- tests: handle UNW_ENOINFO gracefully in ppc64-test-altivec
- tests: always run ppc64-test-altivec, skip at runtime if unsupported

With this, the entire test suite passes on ppc32, ppc64, and ppc64le.

During development, I discovered and fixed a QEMU bug. See [this patch](https://lists.gnu.org/archive/html/qemu-devel/2026-04/msg02410.html). It is needed for a handful of tests to pass on ppc64 under QEMU. I'm still thinking about how to handle XFAILs such that the CI is green. I wouldn't want the test suite to have a bunch of XPASSs on real hardware, so we'll have to do something more clever (as a follow up).


bregma:
So the failures we now see in the CI powerpc64*-linux-gnu jobs are all resolved by your upstream fix in QEMU?

mattst88:
Yes.

I've got a patch locally to selectively control the PASS/FAIL expectations based on the version of QEMU, but the patch hasn't been accepted upstream yet (though it's gotten a Reviewed-by). Once I know that, and whether it will be cherry-picked to a stable branch as I've requested, I'll make a PR for that commit.

Then I can see about getting the fix cherry-picked into Ubuntu... It's a long road for a 10-line fix.

bregma:
Since there's no *new* failures, you vouch for QEMU being the problem for the remaining failures, and I have no real hardware to verify on I have no problem with this going in.

PR #992, opened 2026-04-20
CI: Add jobs with ASan and UBSan enabled

mattst88:
Add build-gnu-native-sanitizer and build-gnu-cross-sanitizer jobs to CI-linux-gnu.yml. The native job covers i686 and x86_64 with both gcc and clang; the cross job covers the same architecture set as build-gnu-cross, minus the ones that don't have sanitizer support (alpha, hppa, mips).


mattst88:
Summary of failures:

- alpha, hppa, and mips* don't have `-lubsan` in their toolchains
- arm: Fixes in https://github.com/libunwind/libunwind/pull/995
- ppc32: Added two commits to fix unaligned accesses to https://github.com/libunwind/libunwind/pull/991
- ppc64, ppc64le: unaligned accesses. Added a commit to https://github.com/libunwind/libunwind/pull/991 that fixes one issue. https://github.com/libunwind/libunwind/pull/998 Fixes another.
- aarch64: unaligned accesses. Fixes in https://github.com/libunwind/libunwind/pull/999
- s390x: Actually a generic problem (here is it occurring on alpha, https://github.com/libunwind/libunwind/actions/runs/24644183722/job/72053591486?pr=995). Fix is in https://github.com/libunwind/libunwind/pull/993

Then there are some problems that appear to occur independent of architecture:

- Gtest-bt, Ltest-bt: store to null pointer
- Ltest-init-local-signal: load from null pointer

These are intentional. Added a patch here to disable UBSan for just the functions that dereference null.

PR #993, opened 2026-04-20
mi: Fix data race in _mem_validate_pipe causing SIGPIPE under concurrent unwinding

### LINKED ISSUE #715: Race condition in the unw_address_is_valid pipe initialization
There is a race condition in how the helper pipe is initialized in the `unw_address_is_valid` that was introduced by https://github.com/libunwind/libunwind/commit/ec03043244082b8f552881ba9fb790aa49c85468. 
Before that commit, the function used `pthread_once` to ensure that only one thread has initialized the pipe. The above mentioned commit changed that to
```c
if (unlikely (!atomic_flag_test_and_set(&_unw_address_validator_initialized)))
    {
      _open_pipe ();
    }
```
which in case of multiple thread racing here can get it into a situation when the first thread would get into the `_open_pipe` and another thread would use the pipe before the first thread completed the pipe opening.
In subsequent changes this pipe opening got moved to the `_write_validate` method, but the race is still the same.

I've observed this as random rare unwind failures in one of the .NET runtime tests where 10 threads are started at the same time and they all do the same exception handling where the libunwind is used to unwind a piece of native code. Reverting this change has fixed the issue.

mattst88:
_mem_validate_pipe was a process-global array shared by all threads. When multiple threads concurrently call unw_address_is_valid() from signal handlers (e.g. 128 threads each running an unw_step() handler for SIGUSR1), the "re-open closed pipe" path in _write_validate() races: thread A calls _open_pipe() closing fds [3,4] while thread B is mid- write() on fd 4. Closing the read end of a pipe while another thread writes to the write end generates SIGPIPE, killing the process.

Fix by making _mem_validate_pipe _Thread_local. Each thread now creates and manages its own pipe; no cross-thread fd access is possible. Remove the now-unnecessary atomic_flag _unw_address_validator_initialized and replace the test-and-set guard with a simple check on pipe[0] == -1.

Verified: 500 consecutive runs of Ltest-concurrent and 200 of Gtest-concurrent under qemu-s390x with zero failures (previously ~1-5% failure rate with SIGPIPE exit 141).

Fixes: ec030432 ("Simplified address validation")
Closes: https://github.com/libunwind/libunwind/issues/715


bregma:
Could we please future-proof this by including `<threads.h>` and using the `thread_local` macro/keyword instead?

mattst88:
Will do. Thanks!

mattst88:
Added a commit to replace `_Thread_local` with `thread_local`. The CMake stuff is copy and paste from elsewhere. Let's see if Windows CI is okay with it.

PR #994, opened 2026-04-20
all: Fix strict-aliasing violations

### LINKED ISSUE #730: -Wstrict-aliasing warning on ARM
We're currently exploring allowing building more of our distribution with LTO. While doing that, we're testing with `-flto -Werror=strict-aliasing -Werror=lto-type-mismatch -Werror=odr` to find likely runtime issues.

With libunwind-1.8.0, on armv7, I get:
```
/var/tmp/portage/sys-libs/libunwind-1.8.0/work/libunwind-1.8.0/src/arm/Gresume.c: In function ‘_Uarm_local_resume’:
/var/tmp/portage/sys-libs/libunwind-1.8.0/work/libunwind-1.8.0/src/arm/Gresume.c:64:19: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
   64 |             "m" (*(struct regs_overlay *)regs)
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: some warnings being treated as errors
```

This code has a kind of interesting history - see 14fc15928b31c18c06f6c075ba0394c0000d3981.

mattst88:
libunwind's public API requires users to allocate unw_cursor_t (an opaque
buffer), which the library internally accesses as struct cursor. Casting
between these two unrelated struct types is a strict-aliasing violation.

Mark both struct cursor (in all tdep headers) and struct unw_cursor (in
libunwind-common.h.in) with `__attribute__((may_alias))` so that compilers
know pointers to these types may alias one another.

Also mark struct regs_overlay in src/arm/Gresume.c with may_alias, as it
is used to alias an unsigned long array for an inline asm memory input
constraint.

Closes: https://github.com/libunwind/libunwind/issues/730

### DIFF
--- a/.github/workflows/CI-linux-gnu.yml
+++ b/.github/workflows/CI-linux-gnu.yml
@@ -54,8 +54,8 @@ jobs:
         env:
           CC: ${{ matrix.toolchain.CC }}
           CXX: ${{ matrix.toolchain.CXX }}
-          CFLAGS: "${{ matrix.target.CFLAGS }} ${{ matrix.optimization.CFLAGS }} -g -Wall -Wextra"
-          CXXFLAGS: "${{ matrix.target.CFLAGS }} ${{ matrix.optimization.CFLAGS }} -g -Wall -Wextra"
+          CFLAGS: "${{ matrix.target.CFLAGS }} ${{ matrix.optimization.CFLAGS }} -g -Wall -Wextra -fstrict-aliasing -Wstrict-aliasing -Werror=strict-aliasing"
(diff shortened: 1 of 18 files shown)


bregma:
Please fix the failures in the Windows builds.

mattst88:
Oh, sorry. Will do. I missed that the CI caught an actual regression in the sea of red :)

I'll make a PR to mark now-passing CI jobs as required, as to prevent them from regressing in the future.

mattst88:
Made a `MAY_ALIAS` macro that is empty on MSVC and is `__attribute__((may_alias))` under `__GNUC__`. Windows jobs are now passing.

PR #999, opened 2026-04-21
aarch64: Sanitizer fixes

mattst88:
Fixes for issues found with UBSan and ASan.

- aarch64: Use memcpy in dwarf_getfp/dwarf_putfp to avoid misaligned-load
- aarch64: Use memcpy in access_fpreg to avoid misaligned-load
- aarch64: Fix out-of-bounds read in access_fpreg debug print
- tests/aarch64: Use memcpy in mock access_mem to avoid misaligned-load
- tests/aarch64: Fix partial-overflow in mock access_mem bounds check

### DIFF
--- a/include/tdep-aarch64/libunwind_i.h
+++ b/include/tdep-aarch64/libunwind_i.h
@@ -135,7 +135,7 @@ dwarf_getfp (struct dwarf_cursor *c UNUSED, dwarf_loc_t loc, unw_fpreg_t *val)
 {
   if (!DWARF_GET_LOC (loc))
     return -1;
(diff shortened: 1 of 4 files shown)


bregma:
Relational operators on two `void *` is undefined behaviour (ISO/IEC 9899:2011 6.5.8/5). The compiler is within its rights to eliminate those expressions. Maybe cast them to an `intptr_t` to make it defined behaviour?

mattst88:
That's a fair point. Will fix.

mattst88:
Fixed.

PR #1000, opened 2026-04-24
tests/aarch64: Fix SVE test SIGILL on FreeBSD and other non-SVE systems

mattst88:
At -O0 (no optimization), clang allocates all local variables in the function prologue, including SVE vector types. Allocating stack space for an svint64_t requires addvl, which is itself an SVE instruction. In main(), this addvl fires before the sve_is_enabled() guard can run, causing SIGILL on systems where SVE is unavailable (e.g. the QEMU cortex-a57 machine used for FreeBSD CI).

Fix by extracting the SVE code into run_test(), called only after the guard. main() now has no SVE locals and its prologue is SVE-free.

Also add FreeBSD SVE detection via elf_aux_info(AT_HWCAP) so the test properly skips when FreeBSD/QEMU lacks SVE, and runs when it is present.

Update the expected call stack to include run_test between square and main.

### DIFF
--- a/tests/Garm64-test-sve-signal.c
+++ b/tests/Garm64-test-sve-signal.c
@@ -105,7 +105,7 @@ signal_handler (int signum)
    * This is the sequencce of calls expected in the call stack.
    */
   static const char *  expected[] = {
-    UNW_TEST_SIGNAL_FRAME, UNW_TEST_KILL_SYSCALL, "sum", "square", "main",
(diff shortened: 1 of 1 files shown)


bregma:
Funny thing is I had this indirection in intermediate fixes for other reasons then removed it before committing as unnecessary.

Still, the FreeBSD detection bit is great.

PR #1006, opened 2026-04-28
 Fix `get_frame_state` on Aarch64 for an unusual assembly code layout

gleocadie:
## Summary

 Improve `get_frame_state` when function assembly code layout is unusual. This was observe on aarch64/musl-libc function.

  ## The problem

  On aarch64, when DWARF unwind info is unavailable for a function (e.g. hand- written assembly with no `.cfi_*` directives, or stripped binaries), `unw_step`  falls back to `get_frame_state` (`src/aarch64/Gstep.c`). That routine performs  a static instruction walk from the function's entry to the current IP and classifies the state of the FP/LR frame record:

  - `STP x29, x30, [sp, …]` → `fs.loc = AT_SP_OFFSET`
  - `mov x29, sp` → `fs.loc = AT_FP`
  - `LDP x29, x30, [sp …]` → `fs.loc = NONE` (frame record popped)

  The walk is purely textual — it does not model control flow. So if the IP is  in a region of the function that lies *past the bytes of the epilogue* but  the epilogue has not actually been executed, the walk wrongly concludes the  frame record is gone.

  A common compiler layout that triggers this is a **mid-function epilogue**:
```
  target:
      stp  x29, x30, [sp, #-16]!     ; prologue
      mov  x29, sp
      cbnz w0, .Lpart2               ; conditional jump forward
      mov  w0, #1
  .Lepilogue:
      ldp  x29, x30, [sp], #16       ; epilogue (mid-function)
      ret
  .Lpart2:                            ; body part 2 — reached via the cbnz above
      mov  w0, #2
      b    .Lepilogue                ; loop back to the (single) epilogue
```
  When IP is anywhere in body part 2, the LDP+RET *bytes* sit between  `start_ip` and IP, but the LDP+RET have not run. The frame record at `x29`  is still valid. Without the fix, `get_frame_state` returns  `fs.loc = NONE`, and `unw_step` takes the LR fallback path which does **not**  update CFA/FP — leading to a wrong/stale unwind state on subsequent steps.

  ## The fix

  `src/aarch64/Gstep.c`, `get_frame_state`:

  1. Introduce a `saved_fs` snapshot.
  2. Before clearing `fs` on LDP detection, copy it into `saved_fs`.
  3. Add a new check at the end of the per-instruction loop body: if we now see
     a `RET` (`0xd65f03c0`) opcode that the IP is strictly past, and we have a
     non-empty `saved_fs`, restore `fs = saved_fs`. This says "the LDP+RET
     bytes are textually behind the IP but execution did not flow through
     them — the frame record is still intact at FP."

  ## Test

  New deterministic test in `tests/`:

  - `Laarch64-test-mid-func-epilogue.S` — assembly target laid out exactly as  shown above. No `.cfi_*` directives, so `dwarf_step` returns `-UNW_ENOINFO`  and the `get_frame_state` fallback is exercised. The exported label
    `target_mid_func_epilogue_in_body2` lets the test pin the IP inside body  part 2.

  - `Laarch64-test-mid-func-epilogue-test.c` — builds a fake `ucontext`  positioned at the body-2 label, populates a fake stack with a frame record  `[caller_x29, return_addr]`, and crucially sets `regs[30]` (the LR) to a
    sentinel value distinct from the saved return address. This makes the two  unwind paths produce **different** outputs:

    | Path                    | Unwound IP                   |
    | ----------------------- | ---------------------------- |
    | `AT_FP` (with fix)      | saved return addr at FP+8    |
    | LR fallback (w/o fix)   | `regs[30]` sentinel (garbage)|

    After `unw_init_local2(UNW_INIT_SIGNAL_FRAME)` + `unw_step`, the test  asserts the unwound IP equals the saved return addr.

  ## Verification

  - **With fix applied**: test prints `PASS` and exits 0 — `unwound IP = 0x2222…2222`  (read from FP+8).
  - **With fix temporarily removed**: test prints  `FAIL: unw_step took the LR fallback path` and exits 1 — `unwound IP =
    0xdeadbeefcafebabe` (read from `regs[30]`).

### DIFF
--- a/src/aarch64/Gstep.c
+++ b/src/aarch64/Gstep.c
@@ -559,6 +559,10 @@ get_frame_state (unw_cursor_t *cursor)
 
   start_ip = c->dwarf.ip - offp;
 
+  frame_state_t saved_fs;
+  saved_fs.loc = NONE;
+  saved_fs.offset = 0;
+
   /* Check for frame record instructions since the start of the procedure (start_ip).
    * access_mem reads WSIZE bytes, so two instructions are checked in each iteration
    */
@@ -581,7 +585,32 @@ get_frame_state (unw_cursor_t *cursor)
           /* Frame-record tracking.  Detects the STP/MOV/LDP sequence that
              creates and destroys a frame record, tracking fs.loc through
              NONE -> AT_SP_OFFSET -> AT_FP -> (back to NONE).  */
+          frame_record_location_t prev_loc = fs.loc;
+          int32_t prev_offset = fs.offset;
           track_frame_record (insn, ip + j * 4, &fs.loc, &fs.offset);
+
+          /* If track_frame_record detected an LDP (AT_FP -> NONE), save the
+             pre-LDP state so we can restore it if we later see a RET that
+             the IP is past (mid-function epilogue pattern). */
+          if (prev_loc == AT_FP && fs.loc == NONE)
+            {
+              saved_fs.loc = prev_loc;
+              saved_fs.offset = prev_offset;
+            }
+
+          /* Check for RET instruction (0xd65f03c0).  If IP is past a RET,
+             execution reached this point via a branch that skipped the
+             epilogue, so the frame record is still intact.  Restore the
+             state from before the epilogue. */
+          if (fs.loc == NONE && saved_fs.loc != NONE && insn == 0xd65f03c0)
+            {
+              fs = saved_fs;
+              saved_fs.loc = NONE;
+              saved_fs.offset = 0;
+
+              Debug (4, "ip=0x%lx => past RET, restoring frame state to loc=%d\n",
+                     (long)(ip + j * 4), fs.loc);
+            }
         }
     }
 
--- a/tests/Laarch64-test-mid-func-epilogue-test.c
+++ b/tests/Laarch64-test-mid-func-epilogue-test.c
@@ -1,0 +1,96 @@
+/* Deterministic test for the get_frame_state RET-detection fix.
+ *
+ * Construct a fake ucontext positioned in body part 2 of
+ * target_mid_func_epilogue (defined in Laarch64-test-mid-func-epilogue.S).
+ * That function has its epilogue (LDP+RET) placed mid-function, with a
+ * body-part-2 region that follows it and whose last instruction branches
+ * back to the epilogue. While the IP is in body part 2 the frame record
+ * is still intact on the stack — the LDP/RET haven't executed.
+ *
+ * unw_step must read the return address from the frame record at FP+8
+ * (AT_FP path), not from x30 (LR fallback). To make the two paths
+ * distinguishable we set ucontext.regs[30] to a value that DIFFERS from
+ * the saved return address at FP+8:
+ *
+ *   AT_FP path  (with fix):    unwound IP = FAKE_RET_ADDR
+ *   LR fallback (without fix): unwound IP = LR_GARBAGE
+ */
+
+#include "libunwind.h"
+#include <ucontext.h>
+#include <string.h>
+#include <stdio.h>
+#include <stdint.h>
+
+extern int  target_mid_func_epilogue(int);
+extern char target_mid_func_epilogue_in_body2[];
+
+#define LR_GARBAGE      ((uint64_t)0xDEADBEEFCAFEBABEULL)
+#define FAKE_CALLER_X29 ((uint64_t)0x1111111111111111ULL)
+#define FAKE_RET_ADDR   ((uint64_t)0x2222222222222222ULL)
+
+int main(void)
+{
+    /* Build a fake stack hosting the frame record that
+     * target_mid_func_epilogue's prologue would have created.
+     * After 'stp x29,x30,[sp,#-16]!' and 'mov x29,sp':
+     *   x29 = SP = original_sp - 16
+     *   memory[x29 + 0] = saved caller x29
+     *   memory[x29 + 8] = saved return address (caller's x30)               */
+    static uint64_t fake_stack[64];
+    memset(fake_stack, 0, sizeof(fake_stack));
+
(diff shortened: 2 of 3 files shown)


mattst88:
> Improve get_frame_state when function assembly code layout is unusual. This was observe on aarch64/musl-libc function.

Just to confirm, this isn't expected to fix any existing unit tests, right? It fixes the new unit tests added in this PR?

> The remaining hunks of upstream commit 016ef7b0 (unw_step SP=X29 line and src/dwarf/Gparser.c return-value tweaks) target code that does not exist on this branch and were intentionally not ported.

There is no upstream commit 016ef7b0.

Other comments:
- We should have a commit message in the git commit
- I think the new tests should be added in a separate commit from the fix itself

gleocadie:
> Just to confirm, this isn't expected to fix any existing unit tests, right? It fixes the new unit tests added in this PR?

@mattst88 
exactly

>There is no upstream commit https://github.com/libunwind/libunwind/commit/016ef7b03cb417d33ce43bf98348641f7f370e95.

oops, I rewrote this branch (was included non-upstream-able commits)

bregma:
Would you be able to reevaluate this in light of #965 being committed? You will need to rebase anyway.

gleocadie:
> Would you be able to reevaluate this in light of #965 being committed? You will need to rebase anyway.

@bregma I did rebase. 
I tried #965 only and I still observed crashes (alpine/aarch64).

lexprfuncall:
I think I see why my change won't necessarily cover this case.  My change will mark the state as unknown when it sees control flow that could invalidate its heuristic.  This change proposes using the last state if it encounters control flow.  That will work in cases where there is an interrupt mid-epilogue and the last state being tracked was correct.  However, there are a few cases that come to mind that might not be handled by this PR:
1) (easy) there is a tail call like, `LDP` followed by a `BR`, instead of a `RET`
2) (medium) there are instructions between the `LDP` and the `RET`
3) (hard) there are multiple epilogues and the last computed state is not valid for the current one.  Since we are trying to handle hand-written assembly without CFI information this cannot be ruled out

I wonder if the right answer here is to simply make the unwind step fail at this frame with a descriptive error rather than try to unwind through it with potentially bad information?

gleocadie:
> I think I see why my change won't necessarily cover this case. My change will mark the state as unknown when it sees control flow that could invalidate its heuristic. This change proposes using the last state if it encounters control flow. That will work in cases where there is an interrupt mid-epilogue and the last state being tracked was correct. However, there are a few cases that come to mind that might not be handled by this PR:
> 
> 1. (easy) there is a tail call like, `LDP` followed by a `BR`, instead of a `RET`
> 2. (medium) there are instructions between the `LDP` and the `RET`
> 3. (hard) there are multiple epilogues and the last computed state is not valid for the current one.  Since we are trying to handle hand-written assembly without CFI information this cannot be ruled out
> 
> I wonder if the right answer here is to simply make the unwind step fail at this frame with a descriptive error rather than try to unwind through it with potentially bad information?

@lexprfuncall 
> where there is an interrupt mid-epilogue and the last state being tracked was correct. 

Not only, my case was the one describe in the description: epilogue | code 1 | prologue | code 2
This PR actually addresses a production crash I got and not from an hand-written assembly code (musl-libc).

> I wonder if the right answer here is to simply make the unwind step fail at this frame with a descriptive error rather than try to unwind through it with potentially bad information?

As of today, when `get_frame_state` returns, it tries something and there is no way (yet?) to stop the unwinding with a descriptive error (more true if you use `uwn_backtrace/unw_backtrace2`).
As I see, this PR is the same vein as `get_frame_state`: best-effort. I agree that there have holes, but since we are statically analyzing assembly code, it will always be a best-effort.

If you have idea to strengthen a bit more this PR, I'd be happy to do it.

> 1. (easy) there is a tail call like, `LDP` followed by a `BR`, instead of a `RET`

I could do this one
```
if (fs.loc == NONE && saved_fs.loc != NONE
    && (insn == 0xd65f03c0                          /* RET */
        || (insn & 0xfffffc1f) == 0xd61f0000))      /* BR Xn */
```
(`BR` and its  PAC variants)

> 2. (medium) there are instructions between the `LDP` and the `RET`

In that case, `fs.loc` == `NONE`, so should be ok not ? What did you have in mind specifically ?

> 3. (hard) there are multiple epilogues and the last computed state is not valid for the current one.  Since we are trying to handle hand-written assembly without CFI information this cannot be ruled out

And yes this one is hard, in that case we should analyze the assembly code and if we detect multiple epilogues, no bother to analyze.

lexprfuncall:
> As of today, when `get_frame_state` returns, it tries something and there is no way (yet?) to stop the unwinding with a descriptive error (more true if you use `uwn_backtrace/unw_backtrace2`). As I see, this PR is the same vein as `get_frame_state`: best-effort. I agree that there have holes, but since we are statically analyzing assembly code, it will always be a best-effort.

This is a significant problem with the way things are structured today.  On aarch64, when `dwarf_step` fails, `get_frame_state` is called which has no way to return an error to its caller.  What x86_64 does is better: when `dwarf_step` fails, `_unw_step_fallback` is called and any error it returns is propagated to the caller of `unw_step`.  I will put together a simple PR that tries to fix this.

> If you have idea to strengthen a bit more this PR, I'd be happy to do it.
> 
> > 1. (easy) there is a tail call like, `LDP` followed by a `BR`, instead of a `RET`
> 
> I could do this one

Great!

> > 2. (medium) there are instructions between the `LDP` and the `RET`
> 
> In that case, `fs.loc` == `NONE`, so should be ok not ? What did you have in mind specifically ?

I now see that you are right and how the case I referred to is correctly handled.  Please disregard my erroneous suggestion.

> > 3. (hard) there are multiple epilogues and the last computed state is not valid for the current one.  Since we are trying to handle hand-written assembly without CFI information this cannot be ruled out
> 
> And yes this one is hard, in that case we should analyze the assembly code and if we detect multiple epilogues, no bother to analyze.

Give it a try?  This might be hard to do in general but a solution for your specific case might be easier.

PR #1007, opened 2026-04-29
alpha, mips: Always inline unw_tdep_getcontext

mattst88:
```
alpha: Always inline unw_tdep_getcontext

When inlining is suppressed (e.g. at -O0), the wrapper acquires its own
stack frame, causing _Ualpha_getcontext to capture two wrong values:

  sc_pc = $26 pointing back into the wrapper (not the caller's call
          site), so the unwinder starts from the wrong IP.

  sc_regs[30] = the wrapper's SP, which is 48 bytes below the caller's
                SP.  dwarf_step uses this as r30 when computing the
                CFA (CFA = r30 + frame_size), producing a CFA that is
                48 bytes too low.  The saved return address is then read
                from below the current stack top, which has been
                overwritten by the unw_step call chain itself.

Add __attribute__((always_inline)) to eliminate the wrapper frame.
When inlined, the JSR to _Ualpha_getcontext is emitted at the true
call site, so both $26 and $30 are captured correctly.
```

```
mips: Always inline unw_tdep_getcontext

The MIPS wrapper has the identical structure: save $28 (GP), call
_Umips_getcontext via PLT, restore GP.  _Umips_getcontext saves $31
(ra) as PC and $sp as the stack pointer.  If the wrapper is not inlined
(e.g. at -O0), both values reflect the wrapper's frame rather than the
caller's, breaking dwarf_step CFA computation in the same way as the
Alpha bug fixed in the previous commit.

Add __attribute__((always_inline)) to eliminate the wrapper frame.
```

### DIFF
--- a/include/libunwind-alpha.h
+++ b/include/libunwind-alpha.h
@@ -176,8 +176,11 @@ extern int unw_tdep_is_fpreg (int);
    across a shared library boundary with lazy PLT binding, the dynamic
    linker's PLT resolver trashes GP before getcontext can save it.
    This inline wrapper captures GP before the cross-module call and
-   fixes up the saved value afterward.  */
-static inline int
+   fixes up the saved value afterward.
+
+   The wrapper must be always_inline so that _Ualpha_getcontext captures
+   the caller's $26 (pc) and $30 (sp), not the wrapper's.  */
+static __attribute__((always_inline)) inline int
 unw_tdep_getcontext (unw_tdep_context_t *uc)
 {
     unsigned long saved_gp;
--- a/include/libunwind-mips.h
+++ b/include/libunwind-mips.h
@@ -165,8 +165,11 @@ extern int unw_tdep_is_fpreg (int);
    When getcontext is called across a shared library boundary with lazy PLT
    binding, the dynamic linker's PLT resolver trashes GP before getcontext
    can save it.  This inline wrapper captures GP before the cross-module
-   call and fixes up the saved value afterward.  */
-static inline int
+   call and fixes up the saved value afterward.
+
+   The wrapper must be always_inline so that _Umips_getcontext captures
+   the caller's $31 (ra/pc) and $sp, not the wrapper's.  */
+static __attribute__((always_inline)) inline int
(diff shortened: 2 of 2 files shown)


include/libunwind-alpha.h
```diff
@@ -176,8 +176,11 @@ extern int unw_tdep_is_fpreg (int);
    across a shared library boundary with lazy PLT binding, the dynamic
    linker's PLT resolver trashes GP before getcontext can save it.
    This inline wrapper captures GP before the cross-module call and
-   fixes up the saved value afterward.  */
-static inline int
+   fixes up the saved value afterward.
+
+   The wrapper must be always_inline so that _Ualpha_getcontext captures
+   the caller's $26 (pc) and $30 (sp), not the wrapper's.  */
+static __attribute__((always_inline)) inline int
```
bregma, line 183: Should use ALWAYS_INLINE from compiler.h for portability.

bregma (changes requested):

include/libunwind-alpha.h
mattst88: I thought that too, but this file is installed to the system while compiler.h is not. I didn't want to make the change to install compiler.h for just this purpose.

mattst88: Should I factor out these bits of compiler.h into another header (preexisting or otherwise) that is installed?

mattst88: @bregma?

mattst88: In the absence of a response, I've tried to consider alternative approaches:

- (this PR) just use `__attribute__((always_inline))`. Directly solves problem. Is supported by gcc and clang, the only compilers we care about on alpha and mips.
  - Could inline the `ALWAYS_INLINE` macro from `compiler.h` into these headers
- Use a GNU statement expression (`({ ... })`) as ppc, ppc64 (conditionally), arm, and aarch64 (unconditionally) do.

I don't have a strong preference, other than landing a fix.

bregma: After some consideration I don't think this is a blocker.

bregma (approved):

PR #1032, opened 2026-05-28
Do not enable C++ exception support on RISC-V by default

andreas-schwab:


mattst88:
Why?

It [doesn't appear to be broken](https://github.com/libunwind/libunwind/actions/runs/26576710474/job/78297903521#step:9:112)

andreas-schwab:
It's incompatible.

mattst88:
You're going to have to use more words.

andreas-schwab:
https://build.opensuse.org/package/live_build_log/home:Andreas_Schwab:riscv:libunwind/python-spyder-kernels/standard/riscv64

bregma:
I don't think the C++ language runtime should be built by default on any target architecture. Linux distro packgers who do not explicitly disable it are going to have unhappy campers knocking at their doors.

If someone is building an embedded system without a toolchain-vendor-supplied C++ language runtime, they can enable it in linunwind themselves since they're already building the libunwind runtime for their target. If you're just looking for call backtracing like most people relying on their distro for libraries, you definitely don't want to replace the vendor-supplied language runtime.

PR #1035, opened 2026-06-16
configure: disable C++ exceptions support by default

mattst88:
Distro packagers enabling C++ exceptions unintentionally would replace the vendor-supplied C++ runtime, causing breakage for users. Those who need it can pass --enable-cxx-exceptions explicitly.

Closes: #1032


bertwesarg:
I think you pushed to many commits into this PR

mattst88:
> I think you pushed to many commits into this PR

Indeed you're right. Thanks for catching that. Fixed.

PR #1043, opened 2026-08-20
ppc32: build and unwind correctly on musl

BKPepe:
libunwind does not build on ppc32 with musl. `__WORDSIZE` is a glibc extension, so `unw_word_t` silently becomes 64 bits wide on a 32-bit target and elfxx.c fails to compile with GCC 14. Once it builds, `unw_getcontext()` on musl leaves frame 0 inside getcontext() itself. The Meson build already pulls in libucontext when the C library has no getcontext(), and its ppc getcontext() is a bare wrapper around the kernel's swapcontext syscall, which saves the NIP as the address after the `sc`. glibc has the same quirk and already gets the fix-up.

Two commits:

1. `ppc: don't rely on __WORDSIZE to pick the word size`. Selects the types from `__powerpc64__` and fixes the same test in Gis_signal_frame.c.
2. `ppc32: fix up the NIP after getcontext() on every Linux C library`. Extends the existing glibc fix-up to every Linux libc, with the musl layout where `uc_regs` is a direct member of `ucontext_t`.

Earlier revisions of this PR shipped a getcontext()/setcontext() implementation in assembly for C libraries without one. It is dropped: the Meson build already requires libucontext in that case, and the header fix-up is all that was missing.

Tested on a Turris 1.x (e500v2, musl soft-float) with libucontext 1.5.2, autotools build: Gtest-bt, Ltest-bt, Gtest-resume-sig, Ltest-resume-sig, Gtest-resume-sig-rt, Ltest-resume-sig-rt, Gtest-exc, Ltest-exc, Gtest-trace, Ltest-trace, Ltest-varargs, Gtest-concurrent, Ltest-concurrent, Gtest-sig-context, Ltest-sig-context, Ltest-init-local-signal, test-async-sig and Gtest-get_proc_name pass. test-setjmp fails as it does on every musl target in master. The Meson cross build for ppc32 musl configures and builds libunwind.a and libunwind-ppc32.a without changes to the build files.


src/ppc32/getcontext.S
```diff
@@ -0,0 +1,113 @@
...
+    is interchangeable with the ucontext_t a signal handler receives.
+
+    Only built where the C library does not provide getcontext().  glibc
+    does, and libunwind-ppc32.h keeps using it there.
+
+    The offsets are verified against the real ucontext_t by the
+    _Static_assert()s in Ginit.c; a layout change breaks the build
+    instead of silently corrupting memory.  */
+
+#if !defined(__GLIBC__)
```
mattst88:
```c
#if !defined(__GLIBC__)
```
This guard never fires. `__GLIBC__` is defined by `<features.h>`, which nothing in this file includes, and the compiler does not predefine it. In a `.S` translation unit it is always undefined, so this body is always assembled, including on glibc.

Confirmed on ppc32 glibc by building this branch:

```
$ nm -D src/.libs/libunwind.so | grep Uppc32_.etcontext
00003fe4 T _Uppc32_getcontext
000040bc T _Uppc32_setcontext
```

And it is assembled with musl's offsets:

```
00003fe4 <_Uppc32_getcontext>:
    3fe4:	stw     r0,192(r3)
    3fe8:	stw     r1,196(r3)
```

which is wrong for glibc:

```
sizeof(ucontext_t)=1184
offsetof uc_mcontext         = 48    (UC_MCONTEXT_OFF assumes 192)
offsetof uc_mcontext.uc_regs = 48
offsetof uc_sigmask          = 52
offsetof uc_reg_space        = 180
```

So on glibc the GPR stores land in the middle of `uc_reg_space`, and `stw 0, UC_REGS_PTR_OFF(3)` overwrites the real `uc_regs` pointer. Harmless today only because `libunwind-ppc32.h` does not wire `unw_tdep_getcontext` to it on glibc, but two wrong-offset symbols ship in every glibc ppc32 build.

Suggest gating at the build system level instead: an `AC_CHECK_FUNC`-driven automake conditional (`if !HAVE_LIBC_GETCONTEXT`) around the `_SOURCES` entries, so both the `.S` and the header guards key off the same test the C code sees. If you'd rather keep it in the preprocessor, `#include <features.h>` has to come first, and it needs an `__has_include` guard since it is glibc-only.

src/ppc32/setcontext.S
```diff
@@ -0,0 +1,103 @@
...
+
+    NIP is loaded into CTR and jumped through, so the context's own CTR
+    value is not restored.  Reaching the target costs either LR or CTR,
+    and LR is the one that matters here: ppc32_local_resume() resumes a
+    frame that may return through it.
+
+    This provides what unw_resume() needs; it is not meant as a general
+    purpose replacement for the libc setcontext().  */
+
+#if !defined(__GLIBC__)
```
mattst88: Same guard problem as `getcontext.S:36`: `__GLIBC__` is never defined in a `.S` translation unit, so this is always assembled. Same fix.

src/ppc32/getcontext.S
```diff
@@ -0,0 +1,113 @@
...
+/*  int _Uppc32_getcontext (ucontext_t *ucp)
+
+    Fill UCP the way the kernel fills a signal frame, so that the result
+    is interchangeable with the ucontext_t a signal handler receives.
+
+    Only built where the C library does not provide getcontext().  glibc
+    does, and libunwind-ppc32.h keeps using it there.
+
+    The offsets are verified against the real ucontext_t by the
+    _Static_assert()s in Ginit.c; a layout change breaks the build
```
mattst88:
```
    Only built where the C library does not provide getcontext().  glibc
    does, and libunwind-ppc32.h keeps using it there.

    The offsets are verified against the real ucontext_t by the
    _Static_assert()s in Ginit.c; a layout change breaks the build
    instead of silently corrupting memory.
```
Neither claim currently holds. It is built on glibc (see the comment on the guard below), and the checks in `ucontext_offsets.h` are inside `#if defined(__linux__) && !defined(__GLIBC__)` on the C side, where `__GLIBC__` *is* visible, so they compile out on exactly the configuration where the offsets are wrong. The verification and the code it verifies are guarded by two different answers to the same question.

src/Makefile.am
```diff
@@ -857,7 +857,9 @@ noinst_HEADERS +=                              \
 libunwind_la_SOURCES_ppc32_common =            \
 	$(libunwind_la_SOURCES_common)         \
 	ppc32/get_func_addr.c                  \
+	ppc32/getcontext.S                     \
```
mattst88:
```make
	ppc32/getcontext.S                     \
```
`libunwind_la_SOURCES_ppc32_common` feeds both `libunwind_local_la_SOURCES` and `libunwind_ppc32_la_SOURCES`, so these land in the remote-only library too:

```
.libs/libunwind-ppc32.so:00003f64 T _Uppc32_getcontext
.libs/libunwind-ppc32.so:0000403c T _Uppc32_setcontext
```

Every other arch puts `getcontext.S` in `libunwind_local_la_SOURCES` under `if ARCH_*`: see aarch64 (line 451), riscv (972), alpha (1032), s390x (1091). Please move both there.

Separately, `src/ppc32/ucontext_offsets.h` needs adding to the `noinst_HEADERS +=` block just above, next to `ppc32/ucontext_i.h`. Without it `make dist` omits the header and the tarball does not build:

```
$ tar tzf libunwind-1.10-pre.tar.gz | grep 'ppc32/.*context'
libunwind-1.10-pre/src/ppc32/ucontext_i.h
libunwind-1.10-pre/src/ppc32/getcontext.S
libunwind-1.10-pre/src/ppc32/setcontext.S

$ ./configure && make
ppc32/Ginit.c:33:10: fatal error: ucontext_offsets.h: No such file or directory
make[2]: *** [Makefile:5147: ppc32/Ginit.lo] Error 1
```

That is a hard break for release tarballs on ppc32, glibc included. `make distcheck` would have caught it.

src/ppc32/getcontext.S
```diff
@@ -0,0 +1,113 @@
...
+	stw	0, UC_GREG(UC_XER_IDX)(3)
+	mfctr	0
+	stw	0, UC_GREG(UC_CTR_IDX)(3)
+
+	/* The kernel points uc_regs at the embedded mcontext; uc_addr()
+	   dereferences it, so it has to be set here too.  */
+	addi	0, 3, UC_MCONTEXT_OFF
+	stw	0, UC_REGS_PTR_OFF(3)
+
+	li	3, 0
```
mattst88, line 182:
```
	li	3, 0
```
Nothing has written `uc_sigmask`, but `ppc32_local_resume()` reads it on the non-signal path:

```c
sigprocmask (SIG_SETMASK, &uc->uc_sigmask, NULL);
setcontext (uc);
```

glibc's `getcontext` fills it with an `rt_sigprocmask(SIG_BLOCK, NULL, &uc_sigmask)` syscall; this one leaves whatever was on the stack, so `unw_resume()` installs a garbage signal mask. Needs the same syscall here, or at minimum zeroing `uc_sigmask`.

FPRs have the same shape of problem: the `PPC_SCF_LINUX_RT_SIGFRAME` path in `Gresume.c` does

```c
memcpy (&_UCONTEXT_UC_REGS (sig_uc)->fpregs, &_UCONTEXT_UC_REGS (uc)->fpregs, ...);
```

which copies uninitialized stack contents into the kernel signal frame, where `sigreturn` then loads them as FP state. `uc_flags`, `uc_link`, `uc_stack` and `gregs[MSR_IDX]` are also left uninitialized.

src/ppc32/unwind_i.h
```diff
@@ -36,6 +36,12 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.  */
 #include <ucontext.h>
 
 #define ppc32_lock                      UNW_OBJ(lock)
+#if !defined(__GLIBC__)
+/* glibc provides setcontext(); use our implementation otherwise.  */
+#define setcontext                      UNW_ARCH_OBJ (setcontext)
```
mattst88:
```c
#define setcontext                      UNW_ARCH_OBJ (setcontext)
```
Macro-defining a libc function name is risky: any translation unit that includes `unwind_i.h` and then reaches `<ucontext.h>` gets the libc prototype rewritten under it. It is also asymmetric with the getcontext side, which correctly routes through `unw_tdep_getcontext`.

Suggest `#define ppc32_setcontext UNW_OBJ(setcontext)` alongside the neighboring `ppc32_local_resume` etc., and updating the four call sites in `Gresume.c` to pick between `ppc32_setcontext` and `setcontext`.

include/libunwind-ppc64.h
```diff
@@ -69,7 +69,12 @@ extern "C" {
 
 #define UNW_TDEP_CURSOR_LEN 280
 
-#if __WORDSIZE==32
+/* Use the compiler's ABI macro rather than the glibc-specific
+   __WORDSIZE, which musl only defines in <sys/reg.h> and <sys/user.h>:
+   an undefined macro would silently select the 64-bit types on a 32-bit
+   target.  mips and riscv key off _MIPS_SIM and __riscv_xlen the same
+   way.  */
+#ifndef __powerpc64__
```
mattst88, line 77:
```c
#ifndef __powerpc64__
```
In the ppc32 header this is right. Here it inverts a branch that was previously dead: `__WORDSIZE==32` was only ever true for a genuinely 32-bit build, whereas `#ifndef __powerpc64__` is true whenever this header is compiled anywhere that is not ppc64, which selects `uint32_t` for `unw_word_t` on the ppc64 target.

That branch should be unreachable in this header, not merely inverted. Either drop the 32-bit half entirely, or make the intent explicit:

```c
#ifndef __powerpc64__
# error "libunwind-ppc64.h requires a 64-bit PowerPC target"
#endif
```

src/ppc32/setcontext.S
```diff
@@ -0,0 +1,103 @@
...
+
+	lwz	0, UC_GREG(UC_CCR_IDX)(3)
+	mtcr	0
+	lwz	0, UC_GREG(UC_XER_IDX)(3)
+	mtxer	0
+	lwz	0, UC_GREG(UC_LINK_IDX)(3)
+	mtlr	0
+
+	/* Jump target.  */
+	lwz	0, UC_GREG(UC_NIP_IDX)(3)
```
mattst88:
```
	lwz	0, UC_GREG(UC_NIP_IDX)(3)
```
This addresses the mcontext through the fixed `UC_MCONTEXT_OFF`, while every other consumer goes through the `uc_regs` pointer: `ucontext_i.h`'s non-glibc `_UC_MCONTEXT_GPR` uses `dmy_ctxt.uc_regs->gregs[x]`, and `Gresume.c` uses `_UCONTEXT_UC_REGS(uc)`. It happens to agree for kernel rt_sigframes, where the kernel points `uc_regs` at the embedded `uc_mcontext`, but the two access paths should not be able to disagree. One `lwz 3, UC_REGS_PTR_OFF(3)` at entry, then `UC_GREG(n)` relative to that, makes this robust and matches `getcontext.S`'s own comment about `uc_addr()` dereferencing the pointer.

src/ppc32/setcontext.S
```diff
@@ -0,0 +1,103 @@
...
+	mtctr	0
+
+	lwz	1, UC_GREG(1)(3)
```
mattst88:
```
	lwz	1, UC_GREG(1)(3)
```
`r1` is restored 30 loads before the last read off `r3`. If the ucontext lives below the restored SP, a signal arriving in that window builds its frame over the memory still being read. ppc32 SysV has no red zone, so this is not merely theoretical. glibc sidesteps it by doing the whole thing in a `swapcontext` syscall. Cheapest fix here is to move the `r1` restore down next to `r0` and `r3`.

src/ppc32/ucontext_offsets.h
```diff
@@ -0,0 +1,54 @@
...
+
+#ifndef __ASSEMBLER__
+
+#include <stddef.h>
+#include <ucontext.h>
+
+#define UC_CHECK(name, expr, want) \
+  typedef char name[((expr) == (want)) ? 1 : -1]
+
+UC_CHECK (uc_check_regs_ptr, offsetof (ucontext_t, uc_regs),     UC_REGS_PTR_OFF);
```
mattst88, line 60:
```c
UC_CHECK (uc_check_greg0,    offsetof (ucontext_t, uc_mcontext.gregs[0]),  UC_GREG (0));
```
The `*_IDX` names come from `ucontext_i.h`, which this header does not include. It works only because `Ginit.c` happens to include `ucontext_i.h` first. Please add the include so the header is self-contained.

Also, `_Static_assert` is available and gives a readable diagnostic instead of `size of array is negative`.

src/ppc/Gis_signal_frame.c
```diff
@@ -58,7 +58,7 @@ unw_is_signal_frame (unw_cursor_t * cursor)
 
   a = unw_get_accessors_int (as);
 
-#if __WORDSIZE == 64
+#ifdef __powerpc64__
```
mattst88, line 61:
```c
#ifdef __powerpc64__
```
Equivalent to what it replaces, so no objection. Worth noting that this file is shared between `libunwind-ppc32.la` and `libunwind-ppc64.la`, and both the old and new tests key off the compiler rather than the unwind target. `UNW_TARGET_PPC64` would be the predicate that actually matches the intent. Fine to leave for a follow-up.

mattst88 (commented):
Nice diagnosis on `__WORDSIZE` -- that's a real bug, and keying off `__powerpc64__` matches the `_MIPS_SIM` / `__riscv_xlen` precedent.

The getcontext/setcontext half has a few problems that I confirmed by building this branch natively on ppc32 glibc (gcc 15.2.1, glibc 2.43). Summary of what I hit:

1. The `#if !defined(__GLIBC__)` guards in the two `.S` files never fire, because `__GLIBC__` comes from `<features.h>` and neither file includes it. Both functions get assembled and exported on glibc, with musl's offsets.
2. `src/ppc32/ucontext_offsets.h` is missing from `noinst_HEADERS`, so `make dist` drops it and the release tarball fails to build on ppc32.
3. The two `.S` files went into `libunwind_la_SOURCES_ppc32_common`, so they also land in the remote-only `libunwind-ppc32.la`.

Details inline.

One more thing that doesn't fit an inline anchor: on a non-Linux ppc32 host (FreeBSD), `ucontext_offsets.h`'s body is skipped by its `defined(__linux__)` guard while the `.S` bodies still compile, so the offsets vanish out from under the code. Simulated with `-U__linux__`:

```
src/ppc32/getcontext.S:51: Error: junk at end of line: `(3)'
```

Not tested on real FreeBSD, but the guards clearly don't line up.

src/ppc32/unwind_i.h
BKPepe: Worth flagging that riscv and s390x do exactly what this used to do, and
without a guard:

-    src/riscv/unwind_i.h:38:#define setcontext  UNW_ARCH_OBJ (setcontext)
-    src/s390x/unwind_i.h:42:#define setcontext  UNW_ARCH_OBJ(setcontext)

That is where I took the pattern from, so ppc32 is now the odd one out.

src/ppc32/ucontext_offsets.h
```diff
@@ -0,0 +1,81 @@
...
+UC_CHECK (uc_check_idx_nip,  NIP_IDX,  UC_NIP_IDX);
+UC_CHECK (uc_check_idx_ctr,  CTR_IDX,  UC_CTR_IDX);
+UC_CHECK (uc_check_idx_link, LINK_IDX, UC_LINK_IDX);
+UC_CHECK (uc_check_idx_xer,  XER_IDX,  UC_XER_IDX);
+UC_CHECK (uc_check_idx_ccr,  CCR_IDX,  UC_CCR_IDX);
+UC_CHECK (uc_check_idx_msr,  MSR_IDX,  UC_MSR_IDX);
+
+UC_CHECK (uc_check_flags,    offsetof (ucontext_t, uc_flags),   UC_FLAGS_OFF);
+UC_CHECK (uc_check_link,     offsetof (ucontext_t, uc_link),    UC_LINK_OFF);
+UC_CHECK (uc_check_stack,    offsetof (ucontext_t, uc_stack),   UC_STACK_OFF);
```
mattst88, line 77: Every other field whose size the assembly depends on is pinned here as well as its offset: `sizeof(greg_t)`, `sizeof(sigset_t)`, `sizeof(double)`. `uc_stack` is the exception. `getcontext.S` zeroes exactly three words at `UC_STACK_OFF`:

```
	stw	0, UC_STACK_OFF(3)
	stw	0, UC_STACK_OFF + 4(3)
	stw	0, UC_STACK_OFF + 8(3)
```

which hardcodes `sizeof(stack_t) == 12` with nothing asserting it. That holds for musl's `{void *ss_sp; int ss_flags; size_t ss_size;}` today, but if a libc ever pads or extends it, the tail silently stays uninitialized: the same class of bug the rest of this header exists to prevent, and it will not show up in the offset checks because `uc_sigmask` at 52 is far enough away that nothing overlaps.

Suggest giving it the same treatment as the other sizes, and driving the zeroing loop from the macro rather than from three open-coded stores:

```c
#define UC_STACK_SIZE     12    /* sizeof(stack_t) */

UC_CHECK (uc_check_stack_sz, sizeof (stack_t), UC_STACK_SIZE);
```

While here, `UC_FPREGS_OFF` has an offset check (`uc_check_fpreg0`) but no check that the 32-entry array the save loop writes actually fits, i.e. something like `sizeof (((mcontext_t *) 0)->fpregs.fpregs) == 32 * UC_FPREG_SIZE`. Same argument, lower stakes.

mattst88:
All the previous review comments appear to be fixed, but I don't have permissions to mark them as resolved.

In lieu of @bregma running the CI, does the test suite pass on ppc32/ppc64 with musl with these patches?

bregma:
> In lieu of @bregma running the CI, does the test suite pass on ppc32/ppc64 with musl with these patches?

There's only the one ppc64le-linux-musl target in CI but it doesn't look terribly successful.

BKPepe:
Test suite on ppc32 musl (Turris 1.x, P2020 e500v2): 34 pass, 1 skip, 4 fail.

Three failures are test harness issues: `Ltest-init` and `Gtest-init` need
`libstdc++.so.6`, while `test-coredump-unwind` expects a core dump argument.

`test-setjmp` aborts due to a likely ppc32/musl bug. `jmp_buf` comes from
the C library, but `longjmp.c` interprets it using `JB_SP` from
`tdep-ppc32/jmpbuf.h`, whose comment says it was copied from x86_64.
On glibc this code is effectively disabled by the canary check, so it
looks like it has never been exercised on non-glibc ppc32. The actual
unwinding works correctly.

For the getcontext guard, I'd prefer doing it through configure:
`AC_SEARCH_LIBS([getcontext], [ucontext])` already detects both musl and
libucontext, and an automake conditional can remove the `__GLIBC__`
guards from both `.S` files.

The only complication is `libunwind-ppc32.h`, which is installed and
cannot see `config.h`. We could expose the result through
`libunwind-common.h.in`, e.g. `UNW_HAS_LIBC_GETCONTEXT`, but that adds a
macro to the installed API.

Would you prefer that, or should we keep the `__GLIBC__` check in the
installed header?

BKPepe:
@mattst88 Can you please take a look? :) I rebased my PR.

mattst88:
I've reviewed it again. It looks good to me. Unfortunately I don't have the powers to do anything more than that.

BKPepe:
@bregma Can this be merged, please? 🙏

BKPepe:
Rebased on top of the master branch.

mattst88:
@bregma: can you approve the GitHub workflows?

mattst88:
@bregma: looks to me like this is ready. no regressions in CI.

PR #1045, opened 2026-08-29
Enable C++ exception support in CI

mattst88:
## Summary
- C++ exception support is disabled by default but we should still test it in CI
- Enable `--enable-cxx-exceptions` / `-Dcxx_exceptions=enabled` in the Linux GNU, Linux musl, and FreeBSD jobs
- Add `g++` to the Alpine package lists for the musl jobs, since `build-base` does not pull it in

## Test plan
- [x] CI passes on all affected jobs


mattst88:
@bregma: this is ready

PR #1047, opened 2026-08-30
README: document Meson build and note autotools deprecation

mattst88:
Lead with Meson setup/compile/install/test steps since that is now the recommended build system. Autotools is still supported but noted as deprecated and planned for removal in a future release.


mattst88:
@bregma: this is ready

PR #1048, opened 2026-08-31
ci: upgrade base packages before installing foreign-arch dev packages

mattst88:
The meson cross jobs that enable a foreign dpkg architecture (arm64, armhf, ppc64el, riscv64, s390x) fail while installing the toolchain:
```
  zlib1g-dev:riscv64 : Depends: zlib1g:riscv64 (= 1:1.3.dfsg+really1.3.1-1ubuntu3) but it is not going to be installed
  E: Broken packages
```
zlib1g is Multi-Arch: same, so zlib1g:$DPKG_ARCH must match the version of zlib1g:amd64 already installed in the container. The ubuntu:26.04 image ships 1:1.3.dfsg+really1.3.1-1ubuntu3 while the archive has moved on to -1ubuntu3.1 in resolute-updates, and apt refuses to reconcile the two, so the install aborts.

Run apt-get upgrade after adding the ports sources so the amd64 side is current before any :$DPKG_ARCH development package is requested. Do it in the autotools cross job too: it only escapes today because liblzma5 happens to have no pending update.

### DIFF
--- a/.github/workflows/CI-linux-gnu.yml
+++ b/.github/workflows/CI-linux-gnu.yml
@@ -144,6 +144,11 @@ jobs:
             echo "deb [arch=$DPKG_ARCH] http://ports.ubuntu.com/ubuntu-ports $CODENAME-updates main restricted universe" \
               >> /etc/apt/sources.list.d/ports.list
             apt-get update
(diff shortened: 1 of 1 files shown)


mattst88:
@bregma: this is ready

PR #1049, opened 2026-08-31
tests: fix Ltest-mem-validate gating on x86_64 and s390x

mattst88:
Ltest-mem-validate mprotects a stack page PROT_NONE and expects the unwinder to stop gracefully instead of faulting. That only works if unw_step validates addresses before dereferencing them.

On x86_64 and s390x that validation is compiled out unless conservative checks are enabled: unw_step only calls dwarf_set_validate() inside #if CONSERVATIVE_CHECKS (src/x86_64/Gstep.c, src/s390x/Gstep.c), and access_mem() dereferences unvalidated addresses otherwise. Every other arch sets c->validate unconditionally in unw_step, so the test works there regardless.

The gating added in 998b2cf3 was therefore inverted: it built the test exactly when the support it needs was absent. On x86_64 with --disable-conservative-checks (-Dconservative_checks=false) the test segfaults in access_mem() and the suite fails.

Build it when conservative checks are enabled, or on the other arches, which validate regardless. Verified on x86_64 with both build systems: with conservative checks the test is built and passes; without, it is not built and the suite has no failures.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -117,8 +117,10 @@ endif
  noinst_PROGRAMS_cdep += forker Gperf-simple Lperf-simple \
 			Gperf-trace Lperf-trace
 
-# only enable Ltest-mem-validate on archs without conservative checks
-if !CONSERVATIVE_CHECKS
+# x86_64 and s390x only validate memory addresses while unwinding when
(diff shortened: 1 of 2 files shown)


mattst88:
@bregma: this is ready

PR #1050, opened 2026-09-01
Bump SONAMEs to .so.11 and add ABI baselines for every cross-compiled architecture

### LINKED ISSUE #1042: Bump SONAMES for 1.10 release
Most of the ABIs have broken since 1.8. The SONAMES need to bump accordingly and the baselines should be refreshed.

_Originally posted by @bregma in https://github.com/libunwind/libunwind/issues/1028#issuecomment-5316561070_

mattst88:
Fixes #1042.

## The SONAME bump

Building v1.8.3 and master for every architecture we can cross-compile and
comparing them with `abidiff` shows an incompatible change on **all** of them.
Interfaces were removed as well as changed, so libtool rule 6 applies and `age`
goes back to zero: `SOVERSION` becomes `11:0:0`, giving `libunwind.so.11` and
`libunwind-<arch>.so.11`.

Common to every architecture:

- `unw_dyn_info_t` lost its `load_offset` member when 62dfb323 reverted
  a4014f33, shrinking a public struct from 704 to 640 bits
- `struct unw_debug_frame_list` lost `load_offset` likewise, 512 to 448 bits
- `_U<arch>_get_elf_image()` gained and reordered parameters

and per architecture:

- **ppc32**: `unw_tdep_fpreg_t` went from `long double` to `double` in 80b1165f
- **hppa**: the exported `_UPT_reg_offset` grew from 65 to 90 entries
- **hppa, ppc, riscv, sh, sparc, alpha**: no longer export the `_Unwind_*` entry
  points or `_ReadSLEB`/`_ReadULEB` by default, because `--enable-cxx-exceptions`
  stopped defaulting to yes on them

`libunwind-setjmp` is byte-identical to 1.8.3 everywhere and keeps `0:0:0`.
`libunwind-coredump` only gained `_UCD_add_backing_file_at_vaddr`, which the
existing `1:0:1` already accounts for. `libunwind-ptrace` did break, on hppa and
ppc32, but the SONAME stays 0; it gets an explicit `PTRACE_SO_VERSION` since it
had no `-version-info` at all and was relying on libtool's default.

Note the SONAME goes 8 -> **11**, not 8 -> 9. `current` is an interface counter,
not the SONAME, and 1.8.0 and 1.8.3 bumped it to 9 and 10 while raising `age` to
keep the SONAME at 8. The shipped `libunwind.so.8` therefore already advertises
interfaces 8, 9 and 10, so those numbers are spent.

## The baselines

Only aarch64, riscv64, s390x, x86 and x86_64 had baselines, and CI only ran
`abi-check` for those, so an ABI break anywhere else was invisible. This adds
alpha, arm, hppa, loongarch64, mips o32/n32/n64, ppc32, ppc64 be/le, riscv32, sh
and sparc64, for 21 verified target configurations. alpha and sparc64 are new
since 1.8.3 and have no earlier ABI to compare against.

The baseline path gains a variant level where one libunwind architecture name
covers several incompatible data models:

    src/abi/mips/linux-gnu/{o32,n32,n64}
    src/abi/riscv/linux-gnu/{32,64}
    src/abi/ppc64/linux-gnu/{be,le}

ppc64 needs the endian split only because glibc lays out `struct _libc_vscr`,
reachable from `ucontext_t`, differently on the two; the mips dumps were checked
and are byte-identical across endianness. The `<target_os>` component is
normalized so that every `linux-gnu*` target shares the plain `linux-gnu`
baselines, since an abi-corpus records neither endianness nor the
floating-point ABI.

`scripts/regen-abi-baselines` is new and drives the whole thing
(`--check`, `--target=`, `--list`). The comparison now runs from `make check`
and `meson test` rather than only on request, and a library with no baseline is
an error rather than a warning when its architecture has a baseline directory.

## Incidental fixes needed to get there

- `src/elfxx.h` had no include guard, so `elf64.c` emitted every definition
  twice on an elfxx 32-bit target
- `libunwind-riscv` and `libunwind-loongarch64` linked `libunwind-elf64.la`
  although configure builds `libunwind-elfxx.la` for them; riscv32 did not
  compile at all
- riscv32 had no `jmp_buf` offsets, just `#error "Add offsets here"`
- `AC_SYS_LARGEFILE`: Meson defines `_FILE_OFFSET_BITS=64` project-wide and
  autotools did not, so `struct coredump_phdr` differed between the two build
  systems on 32-bit targets and they could not share a baseline

## Testing

`scripts/regen-abi-baselines --check` is clean on all 21 configurations, and
`meson test --suite abi` passes for each of them, including with
`-Dcxx_exceptions=enabled`. Native `make check` passes at `-Og`, `-O0` and
`-O3`. Both build systems produce identical library filenames.

## Still uncovered

`ia64` has no baseline, and the two `nto-qnx7.1.0` sets could not be
regenerated here and are now stale, since they still record the pre-1.10
SONAMEs. **Someone with a QNX toolchain should refresh those before 1.10
ships**, otherwise the QNX build will fail its own ABI check.


mattst88:
@bregma: this is ready

BKPepe:
@bregma Gentle ping

PR #1052, opened 2026-09-01
Verify the CRC of a separate debug file found via .gnu_debuglink

### LINKED ISSUE #422: Separate symbols file found via the "debug link" mechanism fails if the link is the same as the name of the executable
For the sake of an example if the executable is "/usr/bin/ls" (and it's been stripped of symbols) and the debug-link is also "ls" (there is no requirement to add a ".debug" suffix and old Debian releases (prior to the move to buildid) don't seem to add such a suffix) then libunwind will correctly consider /usr/bin/ls to be the first (after buildid) candidate location for the symbols. However, we know this file has been stripped but libunwind uses it anyway and thus the symbolic decodes are poor at best (just what's available in the dynamic symbol table that can't be stripped). The actual separate symbols file is under the /usr/lib/debug/... path.

Now in the "debug-link" in addition to the file name there is a CRC and the CRC is supposed to be checked to confirm that the correct file has been found. If it hasn't been found then you should skip onto the next potential location for the separate symbols file. Libunwind doesn't appear to be checking the CRC and thus incorrectly thinks that it's found the symbols when it hasn't. To confirm how this is supposed to work I took a look at the GDB source code and it does check the CRC and only declares that it's found the file if the CRC matches.

```
static bool
separate_debug_file_exists (const char *name, void *crc32_p)
{
  unsigned char buffer[8 * 1024];
  unsigned long file_crc = 0;
  FILE *f;
  bfd_size_type count;
  unsigned long crc;

  BFD_ASSERT (name);
  BFD_ASSERT (crc32_p);

  crc = *(unsigned long *) crc32_p;

  f = _bfd_real_fopen (name, FOPEN_RB);
  if (f == NULL)
    return false;

  while ((count = fread (buffer, 1, sizeof (buffer), f)) > 0)
    file_crc = bfd_calc_gnu_debuglink_crc32 (file_crc, buffer, count);

  fclose (f);

  return crc == file_crc;
}
```

mattst88:
A .gnu_debuglink section records the CRC-32 of the separate debug file along with its name, and the checksum is what tells a consumer whether the file it found is the right one.  libunwind ignored it and accepted the first file whose name matched.

The debug link need not carry a .debug suffix, so a program stripped of its symbols can carry a debug link naming a file with its own basename. The first location libunwind searches is the directory of the program itself, so it found the stripped program, took it for the debug file, and produced poor symbolic decodes using only the dynamic symbol table. A mismatched debug file installed under the expected name was accepted the same way, giving bogus decodes.

Compute the CRC-32 of each candidate and reject it unless it matches the one recorded in the .gnu_debuglink section, so the search continues with the next candidate location.

Fixes: https://github.com/libunwind/libunwind/issues/422


mattst88:
@bregma: this is ready

mattst88:
FreeBSD fixed. Ready to go.

PR #1054, opened 2026-09-01
ci: remove the Linux/musl jobs

mattst88:
The CI - Linux - MUSL workflow has never had a green run.  All 291 completed runs since the workflow was added in March 2026 failed, on every one of its eight jobs (autotools and Meson, aarch64, ppc64le, riscv64 and x86_64).

The failures are not CI flakiness, they are real gaps in libunwind's musl support:

  - test-setjmp and x64-unwind-badjmp-signal-frame fail on every architecture.  libunwind's setjmp()/longjmp() replacement is wired to glibc: include/tdep-*/jmpbuf.h hardcode glibc's jmp_buf slot indices, and src/setjmp/{longjmp,siglongjmp}.c handle glibc's pointer mangling by deferring to glibc's own longjmp.  musl's jmp_buf has a different layout and no such entry point, so the replacement cannot work as written.

  - The signal-related tests (Ltest/Gtest-resume-sig, -resume-sig-rt, Ltest-sig-context, Ltest-init-local-signal, Ltest/Gtest-trace) fail because musl's signal trampoline and mcontext_t layout differ from glibc's, which is what the signal-frame detection and the register-location tables assume.

  - ppc64le fails 25 of 46 tests, essentially the whole unwinding suite.

Fixing this is a substantial porting effort spread across the setjmp replacement, per-architecture signal-frame handling and mcontext layouts, not a CI configuration problem.  The same failures have been reported since 1.3.1:

  https://github.com/libunwind/libunwind/issues/164

Keeping a workflow that is red on every push and every pull request trains people to ignore CI results and hides genuine regressions in the jobs that do pass.  Remove it.  It can be restored once musl support is in a state where it can pass.


mattst88:
@bregma: this is ready

bregma:
Mmm, OK, but we should open an Issue for a port to linux-musl because there are a lot of Alpine-based containers out there and I'm always concerned about "a Linux port" really being a "GNU libc port".

mattst88:
Good idea. I posted https://github.com/libunwind/libunwind/issues/1061

PR #1057, opened 2026-09-13
Add port for Linux on OpenRISC (or1k)

lrfe:
This PR adds support for Linux on OpenRISC (or1k), a 32-bit big-endian
architecture. All tests in the test suite pass.

DWARF register numbers match the hardware register numbers. Ordinary frames
return through r9, while signal frames use column 32 for the PC, matching
libgcc's or1k_fallback_frame_state.

No distribution packages an or1k cross toolchain, so there is no CI runner
for this target. The results below are from a Bootlin
or1k-unknown-linux-gnu toolchain. run-ptrace-mapper and run-ptrace-misc
are XFAIL because OpenRISC does not implement arch_has_single_step(), so
PTRACE_SINGLESTEP returns -EIO.

If CI coverage is useful, I can add a job that downloads a Bootlin toolchain.

`test-ptrace` and `run-coredump-unwind` cannot run under user-mode QEMU, so
they were checked separately on a Linux 6.18 guest under qemu-system-or1k:

```
test-ptrace                              rc=0
test-coredump-unwind -testcase           rc=0
run-ptrace-mapper, run-ptrace-misc       ptrace(PTRACE_SINGLESTEP) failed (errno=5)
```

`make check` under qemu-or1k:

```
PASS: check-namespace.sh
PASS: Gtest-bt
PASS: Gtest-concurrent
PASS: Gtest-exc
PASS: Gtest-get_proc_name
PASS: Gtest-init
PASS: Gtest-resume-sig
PASS: Gtest-resume-sig-rt
PASS: Gtest-sig-context
PASS: Gtest-trace
PASS: Lrs-race
PASS: Ltest-bt
PASS: Ltest-concurrent
PASS: Ltest-cxx-exceptions
PASS: Ltest-exc
PASS: Ltest-init
PASS: Ltest-init-local-signal
PASS: Ltest-mem-validate
PASS: Ltest-nocalloc
PASS: Ltest-nomalloc
PASS: Ltest-resume-sig
PASS: Ltest-resume-sig-rt
PASS: Ltest-sig-context
PASS: Ltest-trace
PASS: Ltest-varargs
PASS: test-async-sig
PASS: test-eh-frame-hdr-sdata8
PASS: test-elf32-gnu-hash
PASS: test-flush-cache
PASS: test-init-remote
PASS: test-iterate-phdr-cache-null
PASS: test-iterate-phdr-reentry
PASS: test-mem
PASS: test-proc-info
PASS: test-reg-state
PASS: test-setjmp
PASS: test-static-link
PASS: test-strerror
SKIP: run-coredump-unwind
SKIP: test-getcontext-gp
SKIP: test-ptrace
```

`meson test` with the same toolchain: 37 Ok, 0 Fail, 4 Skipped.


src/or1k/Gstep.c
```diff
@@ -0,0 +1,90 @@
...
+{
+  struct cursor *c = (struct cursor *) cursor;
+  int ret;
+
+  Debug (1, "(cursor=%p)\n", c);
+
+  if (unw_is_signal_frame (cursor) > 0)
+    return or1k_handle_signal_frame (cursor);
+
+  ret = dwarf_step (&c->dwarf);
```
mattst88, line 91: `sigcontext_format` is set in `or1k_handle_signal_frame()` but never cleared, so once the cursor steps through a signal frame it stays `OR1K_SCF_LINUX_RT_SIGFRAME` for every frame above it. If `unw_resume()` is then called on one of those frames, `or1k_local_resume()` takes the sigreturn path with the old `sigcontext_addr`, `sigcontext_sp` and `sigcontext_pc`, rather than restoring registers and jumping to the target frame. aarch64, s390x and alpha reset it before calling `dwarf_step()`.

```suggestion
  c->sigcontext_format = OR1K_SCF_NONE;
  ret = dwarf_step (&c->dwarf);
```

src/or1k/Gstep.c
```diff
@@ -0,0 +1,90 @@
...
+    c->dwarf.loc[i] = DWARF_LOC (sc_addr + LINUX_SC_GPR_OFF
+                                 + (i - UNW_OR1K_R0) * sizeof (unw_word_t), 0);
+
+  c->dwarf.loc[UNW_OR1K_PC] = DWARF_LOC (sc_addr + LINUX_SC_PC_OFF, 0);
+
+  /* Set SP/CFA and PC/IP.  */
+  dwarf_get (&c->dwarf, c->dwarf.loc[UNW_OR1K_R1], &c->dwarf.cfa);
+  dwarf_get (&c->dwarf, c->dwarf.loc[UNW_OR1K_PC], &c->dwarf.ip);
+
+  c->dwarf.pi_valid = 0;
```
mattst88: The IP recovered from the sigcontext is the interrupted PC, not a return address, so `use_prev_instr` needs to be cleared. Otherwise the next FDE lookup uses `ip - 1`, which misses when the signal arrived on the first instruction of a function. The other ports with signal frame handlers (aarch64, alpha, loongarch64, mips, s390x) do this. The `dwarf_get()` results should probably be checked too.

```suggestion
  ret = dwarf_get (&c->dwarf, c->dwarf.loc[UNW_OR1K_R1], &c->dwarf.cfa);
  if (ret < 0)
    return ret;
  ret = dwarf_get (&c->dwarf, c->dwarf.loc[UNW_OR1K_PC], &c->dwarf.ip);
  if (ret < 0)
    return ret;

  c->dwarf.pi_valid = 0;
  c->dwarf.use_prev_instr = 0;
```

(`ret` would need declaring.)

meson.build
```diff
@@ -422,6 +422,7 @@ archnames = {
     'loongarch64': 'loongarch64',
     'mips': 'mips',
     'mips64': 'mips',
+    'or1k': 'or1k',
```
mattst88, line 426: `or1k` also needs adding to the `cpu_family()` list for `build_coredump` further up in this file. configure.ac enables coredump for `or1k*`, but under meson it's rejected as unsupported, so the two builds disagree.

README.md
```diff
@@ -47,6 +47,7 @@ This library supports several architecture/operating-system combinations:
 | Linux   | MIPS         | ✓      |
 | Linux   | RISC-V       | ✓      |
 | Linux   | LoongArch    | 64-bit only |
+| Linux   | OpenRISC     | ✓      |
```
mattst88, line 50: Could you also add an or1k row to `doc/libc-requirements.md`? This port has its own `getcontext` and doesn't need `setcontext`.

include/api/libunwind-or1k.h
```diff
@@ -0,0 +1,147 @@
...
+
+    UNW_OR1K_PC,
+
+    UNW_TDEP_LAST_REG = UNW_OR1K_PC,
+
+    UNW_TDEP_IP = UNW_OR1K_PC,
+    UNW_TDEP_SP = UNW_OR1K_R1,
+    /* The exception-data registers are r25 and r27, so they are not
+       contiguous.  */
+    UNW_TDEP_EH = UNW_OR1K_R25
```
mattst88, line 112: Since r25 and r27 aren't next to each other, `UNW_REG_EH + 1` is r26. `src/setjmp/longjmp.c` and `siglongjmp.c` set `UNW_REG_EH + 0` and `UNW_REG_EH + 1`, so the second write would go to the wrong register. It doesn't matter yet, since `siglongjmp.S` is a stub, but it will once someone fills it in. Could you either note it here, or have `tdep_access_reg()` accept `UNW_REG_EH + 1` as an alias for r27?

src/or1k/Gget_save_loc.c
```diff
@@ -0,0 +1,61 @@
...
+#include "unwind_i.h"
+
+int
+unw_get_save_loc (unw_cursor_t *cursor, int reg, unw_save_loc_t *sloc)
+{
+  struct cursor *c = (struct cursor *) cursor;
+  dwarf_loc_t loc;
+
+  if (reg >= UNW_OR1K_R0 && reg <= UNW_OR1K_PC)
+    loc = c->dwarf.loc[reg];
```
mattst88, line 42: For `UNW_OR1K_PC`, `loc[32]` is only set by `common_init()` and the signal frame handler. After an ordinary `dwarf_step()` it still points at the initial context (`Gregs.c` has a comment saying the same). As written, this reports a save location for the PC that's wrong for every frame but the first. Returning `UNW_SLT_NONE` for the PC in that case, or the location of r9, would be more accurate.

src/or1k/Gapply_reg_state.c
```diff
@@ -0,0 +1,37 @@
+/* libunwind - a platform-independent unwind library
+   Copyright (c) 2002-2003 Hewlett-Packard Development Company, L.P.
+        Contributed by David Mosberger-Tang <davidm@hpl.hp.com>
+
+   Modified for x86_64 by Max Asbock <masbock@us.ibm.com>
```
mattst88: This header still says "Modified for x86_64 by Max Asbock". Same in `Greg_states_iterate.c`.

src/or1k/Greg_states_iterate.c
```diff
@@ -0,0 +1,37 @@
+/* libunwind - a platform-independent unwind library
+   Copyright (c) 2002-2003 Hewlett-Packard Development Company, L.P.
+        Contributed by David Mosberger-Tang <davidm@hpl.hp.com>
+
+   Modified for x86_64 by Max Asbock <masbock@us.ibm.com>
```
mattst88: Same as above.

lrfe:
Thanks, all fixed.

By the way, sh and riscv seem to have the same two issues. Neither clears `sigcontext_format` before `dwarf_step()` (`src/sh/Gstep.c:111`, `src/riscv/Gstep.c:101`), and neither clears `use_prev_instr` in its signal frame handler.

mattst88:
> By the way, sh and riscv seem to have the same two issues. Neither clears sigcontext_format before dwarf_step() (src/sh/Gstep.c:111, src/riscv/Gstep.c:101), and neither clears use_prev_instr in its signal frame handler.

Oh, thanks for the heads up. I'll investigate.

mattst88:
I opened https://github.com/libunwind/libunwind/pull/1059 for those fixes.

mattst88:
I landed https://github.com/libunwind/libunwind/pull/1059 which added a new test to catch the issue you noted. Could you rebase and check that it passes on or1k?

lrfe:
Sorry for the late reply, I’m away for a few days, so I won’t be able to rebase and test this properly until I’m back. I should be back in about 4-5 days and will check it.

PR #1059, opened 2026-09-17
Fix unwinding and resuming through signal frames

mattst88:
After unw_step() loaded the interrupted PC from a signal frame, several architectures still looked up its FDE at IP - 1. When the signal arrived on a function's first instruction, that found the preceding function or nothing at all. The rs cache made the same mistake on cache hits for any architecture. Separately, sigcontext_format was never reset, so resuming any frame above a signal frame went back through sigreturn with the signal frame's registers.

This series fixes those, a few leaks of the validate flag and unchecked dwarf_get() calls in the signal frame handlers. It also gives x86 the sigreturn resume path x86_64 has.

The new Ltest-sig-first-insn covers the first-instruction case and the resume path with a cold cache, a warm cache and caching disabled. On master it fails on alpha, arm, hppa, loongarch64, ppc32, ppc64le, riscv64, sh4, sparc64, x86, and x86_64 with caching disabled.

The ABI baselines change because dwarf_reg_cache_entry gains a bitfield. The struct stays the same size.


mattst88:
@bregma: this is ready

